PR #4667, opened 2025-03-14
Use display charset to match patterns in mc config files

### LINKED ISSUE #4665: Patterns in mc's configuration files are assumed to be in ASCII character set

### Midnight Commander version

```shell
GNU Midnight Commander 4.8.33
Built with GLib 2.82.4
Built with libssh2 1.11.1
With builtin editor
With mouse support on xterm
With internationalisation support
Virtual File Systems:
Data types:
```

### Operating system

```shell
Darwin Mac 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:24 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6030 arm64
```

### Current behavior

To demonstrate incorrect pattern matching in the user menu (the same applies to `mc.ext', syntax detection and the `ssh_config' parser):

```shell
touch 'file_without_spaces'
touch 'file with spaces'
touch 'file_without_spaces_utf8_local_chars_ä_ü_ö_ß_'
touch 'file_without_spaces_utf8_nonlocal_chars_à_á_'
```

`cat >.mc.menu`
```
shell_patterns=0

+? f [[:space:]] & t r
R Rename - Replace " " in Name
        rename 's/ /\./g' %f -v
```
`CTRL+D`

Press `F2` and observe how the last file matches, even though it shouldn't, because `0xa0` is the code for "Non-Breaking Space" in ASCII.

```pycon
>>> [hex(c) for c in '_à_á_'.encode('utf-8')]
['0x5f', '0xc3', '0xa0', '0x5f', '0xc3', '0xa1', '0x5f']
```

### Expected behavior

The problem is that the search is initialized with `DEFAULT_CHARSET`, which has a value of `ASCII`:

https://source.midnight-commander.org/R/DEFAULT_CHARSET.html

This is correct if `cp_display` is set to ASCII, but if it's set to `UTF-8`, then UTF-8 target strings are searched byte by byte as if they were ASCII strings.

### Additional context

I'm not sure what the right way to solve this problem is.

We can declare that the patterns must be 7-bit ASCII. In this case we can pass `NULL` to the search call and `cp_display` will be used. This will result in correct behavior for all display charsets that are superset or ASCII. Are there any deal-breaking exceptions we should consider?

Or we could declare that we allow UTF-8 patterns and then convert the pattern to `cp_display` (if necessary and possible) before searching. Is this a better option, or does it just make things more complicated and problematic?

zyv:
## Proposed changes

Resolves: #4665.

## Checklist


tests/src/usermenu.c
```diff
@@ -0,0 +1,163 @@
...
+#ifdef HAVE_CHARSET
+    // In fully UTF-8 environment, only file with real spaces should match
+    { "UTF-8", TRUE, 0, FALSE },
+    { "UTF-8", TRUE, 1, TRUE },
+    { "UTF-8", TRUE, 2, FALSE },
+    { "UTF-8", TRUE, 3, FALSE },
+#endif
+
+    // Last filename contains 0xa0 byte, which would be interpreted as a non-breaking space in
+    // single-bit encodings
```
ossilator: byte

aborodin:
> @zyv requested your review on: MidnightCommander/mc#4667 Use display charset to match patterns in mc config files.

Added tow fixups to the 4665_search_display_charset branch.

Commit "tests: share fixtures between tests in `src`" is not specific to this branch. I propose to move it to 4633_cleanup.

zyv:
> Added tow fixups to the 4665_search_display_charset branch. 

Thank you! I didn't find `dir_list_free_list` :(

> Commit "tests: share fixtures between tests in `src`" is not specific to this branch. I propose to move it to 4633_cleanup.

Let's keep it here, please. If I move this commit somewhere else, I have two options:

1. Either I have to share `mc.charsets` with the editor test by hardcoding the path to a different folder,
3. Or I have to introduce a copy of this file for new test, only to remove it later in cleanup... after this branch is merged and cleanup is rebased on `master`.

On top of that, if that commit ends up in cleanup, it will hang there until the cleanup branch is merged back into `master`, and will be inaccessible from the new branches based on `master` until that happens.

It's a lot of work, and I don't see any benefit to myself or other developers:

This change is isolated in a commit and not mixed in with the others. What's the problem with committing it to this branch? Also, logically, it's a refactoring that affects those very tests, and is minor in nature. So it belongs here.

PR #4671, opened 2025-03-20
Ticket #4670: customize error message when `known_hosts` file cannot …

### LINKED ISSUE #4670: SFTP Link gives error "Failed to open file (-16)" when using different $HOME folder..

### Midnight Commander version

```shell
> ./bin/mc -V
GNU Midnight Commander 4.8.33
Built with GLib 2.74.6
Built with libssh2 1.10.0
With builtin editor
With mouse support on xterm
With support for X11 events
Virtual File Systems:
Data types:
```

### Operating system

```shell
Linux hostname 6.11.10+bpo-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.11.10-1~bpo12+1 (2024-12-19) x86_64 GNU/Linux
```

### Current behavior

- I compiled MC myself
- used --prefix=/somewhere/else
- I set HOME=/somewhere/else/mc_home
- I run /somewhere/else/bin/mc
- I try to access an SFTP link
- I get this error message

Image

### Expected behavior

- Connect to SFTP as usual, as it is when not changing the HOME directory.
or
- Give more detailed error message about what file cannot be opened (so we can fix ourself?!).

### Additional context

I assume this is related to MC not finding ssh-related files in the new / redirected HOME folder, but I'm not sure and if yes, which one to put there to make it work.

My main goal was to compile MC on my own and create a separate MC in its own folder and its own settings which I can run from anywhere without installing from package manager first. So, I would like to have a portable application or AppImage style experience. Just run and use a separate config, but setting HOME to some other place does not work, if you still need SFTP links to work (for now).

Other aspects and functions seem to work with a different HOME, but I did not test much, after realizing SFTP fails. Also the output of "mc -F" still references system folders and I guess MC will read things from there, if there is a regular MC installed via package manager, which I don't want to happen.

Thank you! o)

zyv:
…be opened

## Proposed changes

I suggest adding a custom error message before trying to access `known_hosts`, because the stuff coming from `libssh2` is too cryptic.

Resolves: #4670

### DIFF
--- a/src/vfs/sftpfs/connection.c
+++ b/src/vfs/sftpfs/connection.c
@@ -273,6 +273,14 @@ sftpfs_read_known_hosts (struct vfs_s_super *super, GError **mcerror)
 
     sftpfs_super->known_hosts_file =
         mc_build_filename (mc_config_get_home_dir (), ".ssh", "known_hosts", (char *) NULL);
+
+    if (!exist_file (sftpfs_super->known_hosts_file))
+    {
+        mc_propagate_error (mcerror, 0, _ ("sftp: cannot open %s:\n%s"),
+                            sftpfs_super->known_hosts_file, unix_error_string (errno));
+        return FALSE;
+    }
+
     rc = libssh2_knownhost_readfile (sftpfs_super->known_hosts, sftpfs_super->known_hosts_file,
                                      LIBSSH2_KNOWNHOST_FILE_OPENSSH);
     if (rc > 0)


mc-worker:
Added fixup.
Please add `sftpfs` to the commit message. Something like

> Ticket #4670: sftpfs: customize error message when `known_hosts` file cannot be opened

zyv:
> Added fixup. 

Thanks!

> Please add `sftpfs` to the commit message. Something like

Done.

PR #4672, opened 2025-03-20
Improve issue template

ossilator:
i tried to actually submit a bug, and noticed these two issues.
the changes are not tested, because i wouldn't know how to do that.

on a side note, i find such forms overly bureaucratic, to the point of being silly, and will often just enter "..." into required but obviously redundant fields.
otoh, i'm quite aware of most people needing lots of guidance to produce an even remotely useful report, so this is fine.
just thinking aloud ...

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug.yml
@@ -8,7 +8,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Please briefly explain what is not working properly.
+        Please summarize the issue in a few words.
 
         :stop_sign: _Make sure you are reporting an issue. For general questions, please use [Discussions](/MidnightCommander/mc/discussions)._
 
@@ -16,7 +16,7 @@ body:
     id: search-confirmation
     attributes:
       label: Is there an existing issue for this?
-      description: 'Please [search :mag: the issues](/MidnightCommander/mc/issues) to check if this bug has already been reported.'
+      description: 'Please [:mag: search the issues](/MidnightCommander/mc/issues) to check if this bug has already been reported.'
       options:
         - label: I have searched the existing issues
           required: true
@@ -24,8 +24,7 @@ body:
   - type: textarea
     id: mc-version
     attributes:
-      label: Midnight Commander version
-      description: What Midnight Commander version are you running?
+      label: Midnight Commander version and build configuration
       placeholder: Output of `mc -V`
       render: shell
     validations:
@@ -35,7 +34,6 @@ body:
     id: mc-os
     attributes:
       label: Operating system
-      description: What operating system do you use?
       placeholder: Output of `uname -a`
       render: shell
     validations:
@@ -51,10 +49,10 @@ body:
         - label: I confirm the issue is still reproducible with the latest version of Midnight Commander
 
   - type: textarea
-    id: behavior-current
+    id: reproduction-steps
     attributes:
-      label: Current behavior
-      description: Describe the problem you are experiencing. Screenshots or screencasts are welcome.
+      label: How to reproduce
+      description: Please give minimal, yet complete instructions how to reproduce the issue.
     validations:
       required: true
 
@@ -62,7 +60,15 @@ body:
     id: behavior-expected
     attributes:
       label: Expected behavior
-      description: Describe what you expect to happen instead.
+      description: Please describe what you think should happen.
+    validations:
+      required: true
+
+  - type: textarea
+    id: behavior-current
+    attributes:
+      label: Actual behavior
+      description: Please describe what is happening instead. Screenshots or screencasts are welcome.
     validations:
       required: true
 
--- a/.github/ISSUE_TEMPLATE/1-bug.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug.yml
@@ -52,7 +52,7 @@ body:
     id: reproduction-steps
     attributes:
       label: How to reproduce
-      description: Please give minimal, yet complete instructions how to reproduce the issue.
+      placeholder: Please give minimal, yet complete instructions how to reproduce the issue.
     validations:
       required: true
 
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/1-bug.yml
```diff
@@ -31,7 +31,7 @@ body:
     validations:
       required: true
 
-  - type: textarea
+  - type: input
```
zyv: I intentionally set this to `textarea`.

Yes, it's just an input line. However, input lines are always rendered by GitHub as plain text, and for text areas you can force it to put fences (```) around them:

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema#input

If I kept it as an input line, it seemed to me that it will be too easy for users to destroy the output formatting.

.github/ISSUE_TEMPLATE/1-bug.yml
```diff
@@ -21,7 +21,7 @@ body:
         - label: I have searched the existing issues
           required: true
 
-  - type: textarea
+  - type: input
```
zyv: Have you actually tried running `mc -V`? I really want to have it all and as a fenced code block.

.github/ISSUE_TEMPLATE/1-bug.yml
```diff
@@ -41,6 +41,14 @@ body:
     validations:
       required: true
 
+  - type: textarea
```
zyv: This sort of duplicates the "Current behavior" block. However, it is true that nothing specifically talks about reproduction, which is not good.

I'd rather see your text combined with the description of `behavior-current` rather than making the annoying form even more annoying.

zyv (changes requested):
> on a side note, i find such forms overly bureaucratic, to the point of being silly, and will often just enter "..." into required but obviously redundant fields.
otoh, i'm quite aware of most people needing lots of guidance to produce an even remotely useful report, so this is fine.
just thinking aloud ...

So do I, but it seems to be working surprisingly well so far. I honestly didn't expect to see any improvement. However, with the exception of one person who decided to sabotage the report out of spite (in his own words) and earned himself a ban, incoming reports can now mostly be processed without asking unnecessary questions to the submitter. So I'll stick with it.

.github/ISSUE_TEMPLATE/1-bug.yml
ossilator: did you try that?
and how are inputs different from textareas? i fail to see the difference.

.github/ISSUE_TEMPLATE/1-bug.yml
ossilator: yes, but i would have naturally used only the first line.
if you want all of it (which i guess is reasonable), then the description needs to mention "and build configuration".

.github/ISSUE_TEMPLATE/1-bug.yml
ossilator: i actually had the same thought, but then one can make an even stronger case for collapsing the "actual" and "expected" blocks.

but if we approach this from the "guidance" perspective, then it actually makes sense to atomize it, to force the submitter to think about what is cause and what is effect, and possibly consider implicit assumptions.

.github/ISSUE_TEMPLATE/1-bug.yml
zyv: Yes, of course I tried that. Did you?

https://github.com/zyv/test-github-issue-templates/issues/1

Notice how the value of the input is inserted literally and the value of the textarea is enclosed as a block of code.

.github/ISSUE_TEMPLATE/1-bug.yml
zyv: Alright, feel free to suggest a better description. As I said, so far everyone has actually included the full output, so I was quite happy about that.

.github/ISSUE_TEMPLATE/1-bug.yml
zyv: > but if we approach this from the "guidance" perspective, then it actually makes sense to atomize it, to force the submitter to think about what is cause and what is effect, and possibly consider implicit assumptions.

Ditto, here is the latest (positive) example: https://github.com/MidnightCommander/mc/issues/4670

.github/ISSUE_TEMPLATE/1-bug.yml
ossilator: no, because i don't know how. just enabling issues in my clone didn't help.

but i tried messing with your test repo, and as you can see, GH clearly escapes/filters my attempts. so i think an input for the os will be ok.

zyv: > but i tried messing with your test repo, and as you can see, GH clearly escapes/filters my attempts. so i think an input for the os will be ok.

No, it renders it as Markdown. Try inserting an image with a link to a rickroll video.

Screenshot 2025-03-20 at 17 41 00

ossilator: well, a rickroll is silly, but not really bad.

anyway, you can see in issue 4 that one can mess up things even with textfields. so this is actually a lost cause. and i mean ... big deal. if somebody tries hard to get themselves banned, let them.

zyv (approved):
I'll stick with my textarea because I like it. Thanks for the rest. I've updated the feature request template and fixed DCO.

I've also changed the textarea descriptions to placeholders. Hopefully this will be easier to read (larger type) and make the form a little less annoying.

ossilator:
> I'll stick with my textarea because I like it.

well, i find it kinda confusing to have an obviously oversized field. and it steals screen real estate. but whatever.

> I've also changed the textarea descriptions to placeholders. Hopefully this will be easier to read (larger type) and make the form a little less annoying.

i had the same idea, but left it alone in the end, because at least in the case of the "context" field, not having the enumeration disappear might be potentially helpful, and i wanted to keep it consistent. otoh, it probably doesn't make too much of a difference unless the reporter has the attention span of a goldfish, and consistency might be overrated in this case.

Issue #2920, opened 2012-11-02
Better skin files


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/2920 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `iwfmp` |
> | **Mentions** | `egmont@….com` (@egmontkob) |

<p>
I forked and improved a theme called "solarized".<br />
I made a few improvements:<br />
</p>
<ol><li>Adding comments on each line.
</li><li>Collecting all variables I could find.
</li></ol><p>
I shouldn't explain why this is necessary - <br />
The documentation at the moment is pure garbage, <br />
and it took me a while to collect all entries,<br />
and a while more to understand what each entry does.<br />
</p>
<p>
It is still incomplete, and I hope some could help with the stuff I, personally, don't use (like the editor or viewer), or refer to the TODO's there to help explain entries I couldn't find or understand.<br />
</p>
<p>
You are free to view it here:<br />
<a class="ext-link" href="https://github.com/iwfmp/mc-solarized-skin">https://github.com/iwfmp/mc-solarized-skin</a><br />
</p>

Issue #1514, opened 2009-08-09
mcedit: optional different background for columns > 80


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/1514 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `sergey-feo` (sergey-feo@….ru) |

<p>
In programming good style is do not exceed 80 columns.<br />
It will be good if mcedit will have option for different background color for columns 81 and righter.<br />
</p>
<p>
You can see examples in attached files.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [80qtc.gif](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/1514/80qtc.gif) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/1514/80qtc.gif)) by `sergey-feo` (sergey-feo@….ru) on <span title="2009-08-09T18:13:48+00:00">`Aug 9, 2009 at 18:13 UTC`</span>
> * [80.gif](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/1514/80.gif) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/1514/80.gif)) by `sergey-feo` (sergey-feo@….ru) on <span title="2009-08-09T18:14:20+00:00">`Aug 9, 2009 at 18:14 UTC`</span>

PR #4683, opened 2025-04-12
Ticket #4683: add menu item to toggle right margin

zyv:
## Proposed changes

In ticket #1514 a function was introduced to show the right margin in the text editor. The option is only accessible through the INI-file. No menu entry or editor option was added to access it from the TUI.

In ticket #2920, it was commented that this hidden option should be taken into account when designing skins.

Many years have passed. I suggest adding a menu entry, just like for other toggle options. I understand that the concern at the time was to avoid an explosion of menus and options, but a hidden feature is not helpful.

We can take this into account when we rebuild the menu/options dialog one day, until then at least this option will be directly usable.


zyv:
@aborodin the correct command is `/approve`. I fixed the action to run on edits too, in case you had to fix the wrong command.

Issue #2731, opened 2012-02-10
long filenames scroll


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/2731 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `angel_il` (@ilia-maslakov) |

<p>
The horizontal scroll of filenames in panel is very useful to overview long filenames and see the trancated path of filename.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [far_scroll.avi](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/2731/far_scroll.avi) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/2731/far_scroll.avi)) by `angel_il` (@ilia-maslakov) on <span title="2012-02-13T19:30:47+00:00">`Feb 13, 2012 at 19:30 UTC`</span>
> * [mc_scroll.avi](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/2731/mc_scroll.avi) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/2731/mc_scroll.avi)) by `angel_il` (@ilia-maslakov) on <span title="2012-02-14T05:33:41+00:00">`Feb 14, 2012 at 5:33 UTC`</span>

PR #4685, opened 2025-04-13
Enable filename scrolling by default and fix background color

zyv:
## Proposed changes

The filename scrolling feature has a background coloring bug:

| | Full listing | Long listing | User listing |
|-|-|-|-|
| < | before-full | before-long | before-user |
| > | after-full | after-long | after-user |

Refs: #2731


mc-worker:
Added commit with small fix.

zyv:
> Added commit with small fix.

So I actually thought about adding the exact same "fix" you did, but then I realized that this was probably done on purpose. Apparently, this feature was implemented to be completely hidden:

Initially, `panel->content_shift == -1` and not `0`. So you "enable" it by asking it to shift to the right once, even though it doesn't actually shift anything. That is why both indicators (`{` & `}`) appear. You "disable" it by shifting left to `-1`, so both indicators disappear.

I think that in FAR Manager this behavior is always unconditionally enabled. What do you think about enabling this feature by setting `panel->content_shift == 0` by default and leaving the traditional collapsing in the status bar?

I think right now the discoverability of this feature is zero. I didn't even know it existed. But I actually like it better than collapsing, and if collapsing stays in all other places, hopefully this would not make people unhappy.

zyv:
> /approve

Ha-ha, there is no end to this. Do you know why this failed now? Because @mc-worker is a **private** member of the organization, unlike @aborodin (or me). So GitHub cannot really check if you are authorized to post the comment o_O If you set it to public in @mc-worker  profile, it should finally work :)

mc-worker:
> I think that in FAR Manager this behavior is always unconditionally enabled. What do you think about enabling this feature by setting `panel->content_shift == 0` by default and leaving the traditional collapsing in the status bar?

Done. Please test.

mc-worker:
@zyv please review

mc-worker:
Hm... Scrolling of long filename is not documented at all. There is no any info in the man page.

zyv:
> Hm... Scrolling of long filename is not documented at all. There is no any info in the man page.

I have added the shortcuts to the manual page.

Thanks for your help! Is it okay to merge now?

aborodin:
> > Hm... Scrolling of long filename is not documented at all. There is no any info in the man page.
> 
> I have added the shortcuts to the manual page.

I think it is not enough. The feature of filename scrolling is not described.

> Thanks for your help! Is it okay to merge now?

It would be great to add few words about this feature.

zyv:
> I think it is not enough. The feature of filename scrolling is not described.
> It would be great to add few words about this feature.

So what is there to describe? I checked FAR's help file and this is all they have on the subject:

> Прокрутка длинных имён и описаний               #Alt+Left,Alt+Right#
>                                                   #Alt+Home,Alt+End#

Can you make a suggestion? I can translate it into English. I really have no idea what to write.

I think the real problem was that this feature was not only hidden, but the default shortcuts were not assigned (and of course not described).

mc-worker:
> I checked FAR's help file and this is all they have on the subject:

Indeed.

Ok. Approved.

PR #4690, opened 2025-04-22
Use compiler attribute detection

zyv:
## Proposed changes

This fixes the `master` build.

I think we can get rid of `autoconf` attribute detection, because `__has_attribute`, which I think is cleaner, is supported in all newer compilers, and for old compilers we can just say that you should build without `-Werror`, and the tests are not guaranteed to pass anyway.

What do you think?

Refs: https://github.com/MidnightCommander/mc/discussions/4687


lib/global.h
```diff
@@ -9,11 +9,15 @@
 
 #include <glib.h>
 
-#if defined(HAVE_FUNC_ATTRIBUTE_WEAK) && defined(HAVE_TESTS)
+#if defined __has_attribute
```
ossilator: you're leaving the various constants undefined when this is undefined.

you should probably define an MC_HAS_ATTRIBUTE macro for that purpose.

ossilator:
seems sensible.

the commit messages should be a bit more verbose, saying why you're doing these things.

lib/global.h
zyv: Thanks for noticing, I really hate that macro stuff. I hope it's fixed now.

lib/global.h
```diff
@@ -9,7 +9,13 @@
 
 #include <glib.h>
 
-#if defined(HAVE_FUNC_ATTRIBUTE_WEAK) && defined(HAVE_TESTS)
+#if defined(__has_attribute)
+#define MC_HAS_ATTRIBUTE __has_attribute
```
ossilator: for legibility, i'd prefer this to be a function-like macro. the branches would be consistent then, too.
if that doesn't work for some reason, a comment should say so.

zyv: I don't know what you mean. Can you just suggest a fix? As I said, my relationship with macros and C in general is a difficult one to remain politically correct.

zyv: I think I finally figured it out and made it work. Not sure what I was doing wrong before as I was getting compiler errors...

ossilator: yep, looking good now.

PR #4695, opened 2025-04-28
Add SortByVersion keymap

vpatkov:
MC already has `SortByName`, `SortByExt`, etc. but a keymap for sorting by version is missing. This patch adds it.


zyv (approved):
I added a forgotten `# SortByVersion =`. Other than that I think it looks good.

zyv:
Thank you for your contribution!

PR #4697, opened 2025-04-30
Update and rename README to README.md to fix rendering on Github

notnout:
Update link to doc/README.md and rename to README.md and reformat using Github markdown. The goal is to make the README render as expected on Github.

### DIFF
--- a/README
+++ b/README
@@ -1,1 +1,0 @@
-doc/README
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+doc/README.md


zyv:
Thanks, but you also need to update the build system if you change the name of the file. And it would be nice to have it reformatted using proper Markdown if renamed.

notnout:
Unfortunately I'm not familiar with the build system. I actually did update the format to proper markdown, but I failed to include it in the Github web editing UI (I just don't know where to do it in the UI). 
So let me just drop this PR, sorry for taking some of your time.

PR #4700, opened 2025-05-02
Show size in 'bsize' format for dirs if's computed

vpatkov:
This patch makes `bsize` listing format show real directory size if it's computed by "Show directory sizes" command (ctrl-space) instead of `SUB-DIR`.

### DIFF
--- a/doc/man/mc.1.in
+++ b/doc/man/mc.1.in
@@ -810,7 +810,8 @@ displays the file size.
 is an alternative form of the
 .B size
 format. It displays the size of the files and for directories it just
-shows SUB\-DIR or UP\-\-DIR.
+shows SUB\-DIR or UP\-\-DIR unless the directory size is computed by the
+"Show directory sizes" command.
 .TP
 .B type
 displays a one character wide type field.  This character is similar to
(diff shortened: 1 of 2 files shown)


zyv:
The `bsize` format was introduced precisely to NOT show directory sizes, because the "directory" size is actually the size of the file entry, not the size in the sense of containing files.

I think this change makes sense though, because the computed sizes are the "real" sizes, and otherwise this menu is useless when using this format. It's also consistent with FAR.

@aborodin, what do you think?

But if this change is made, the documentation should be updated as well:

> bsize is an alternative form of the size format. It displays the size of the files and for directories it just shows SUB-DIR or UP--DIR

vpatkov:
> It's also consistent with FAR.

And Total Commander.

JonnyTech:
> > It's also consistent with FAR.

> And Total Commander.

And Double Commander ;)

zyv:
Sorry for the spam and thanks for your contribution!

I took the opportunity to fix the remaining bugs (hopefully!) in our infrastructure, which doesn't get used very often :(

Issue #3169, opened 2014-02-11
Additional appearance options for skins


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3169 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `egmont` (@egmontkob) |

<p>
Forking from #2165:<br />
</p>
<p>
Currently skins define multiple things: colors, single vs double lines, certain Unicode symbols etc. One might argue that it's the intended behavior: a skin should define all details of the look (but not the layout or behavior). On the other hand, it might be reasonable to decouple these from each other, since that gives more flexibility for those users who don't dig into editing skin files. For all the three ideas listed here there might be scenarios (e.g. Linux console) where one of the two states is not supported, hence currently the available skins are quite limited and moving these properties to separate options significantly increases the number of usable skins. It's also useful because it allows the users to easily create something that visually pleases them more, e.g. I personally prefer single lines but there are some great color schemes using double lines.<br />
</p>
<p>
I have three ideas in mind. At this moment the first two are implemented.<br />
</p>
<ol><li>Single lines vs. double lines.
</li></ol><p>
All the skins were altered to properly define the double line drawing characters. It's a runtime decision whether those are actually used or not.<br />
</p>
<p>
It's not only a personal preference, it's also a workaround for some limited terminals. E.g. my Linux console cannot display double lines, whereas the modarcon* skins specifically designed for console uses double lines.<br />
</p>
<p>
The "darkfar" skin is removed, since this was the double-line version of "dark". The "double-lines" skin was also removed, it's quite like the default with double lines, except for the (IMO very ugly) dark blue menu/button lines. If you prefer to keep this skin, it should be renamed to something else. The "featured" skin could probably also be removed, it's almost like "default", but uses UTF-8 instead of ASCII compatibility stuff (see the 3rd point). Maybe the whole [Lines] section could be removed from the skins and be hardcoded in mc, I see no reason why anyone would want to redefine them - but in the initial patch it's not removed so I didn't remove any feature.<br />
</p>
<ol start="2"><li>Underlined hotkeys
</li></ol><p>
[...]

PR #4702, opened 2025-05-08
Remove "Display bits" dialog

### LINKED ISSUE #1612: Please remove "Display bits"
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/1612 |

<p>
Please remove the "Display bits" configuration option on systems with locale support (configure could be able to detect this).<br />
</p>
<p>
This is just another stupid source of breaking things. It's not really clear what the possible values are, it's a source of problems if it's set to an incorrect value (even let's say incorrect default or incorrect value shipped by a distributor). It's also a problem if the user's switching back and forth between two different systems - e.g. during transition from iso-x to utf-8.<br />
</p>
<p>
On systems with locale support mc can detect the character set to be used. Let it do so and not ask for the user to set it correctly or break it. Obviously to have a working operating system with applications, you have to set the global LANG/LC_stuff variables correctly anyway; requiring to set them in each and every application is the wrong way to go. Let's not offer a config option: "Do you want mc to work correctly? Yes No".<br />
</p>
<p>
It should Just Work.<br />
</p>

zyv:
## Proposed changes

I went ahead and removed the "Display bits" dialog. This entails the following changes:

* The display encoding is now automatically detected and cannot be changed manually, so it is no longer stored in the configuration file.
  * If the `mc.charset` file doesn't have an appropriate entry, it defaults to a "safe" 7-bit ASCII encoding.
  * If something is not displayed correctly, the user has to adjust the `LC_*` variables (this could be overridden by manual selection before) and/or possibly `mc.charset` (this hasn't changed).
  * The `use_8th_bit_as_meta` functionality (last used in some super old weird QNX terminals, apparently) has been removed.
* Source (file I/O) encoding can be selected as before

I added the currently detected display and the currently selected source encoding to the "Appearance" dialog. This may be controversial, but I think it's important and useful to show what the values are somehow, and I couldn't find a better place. Also, it is indeed related to appearance...

Screenshot 2025-05-08 at 18 40 50

Resolves: #1612


mc-worker:
Approved with some fixups of coding style. Feel free to reject any of them if you dislike it,

ossilator:
> This may be controversial, but I think it's important and useful to show what the values are somehow, and I couldn't find a better place. Also, it is indeed related to appearance...

i think it would be better to put it in a separate dialog under a menu item "about terminal" (such items are usually in the "help" menu, but i guess mc could have it in a separate section at the top or bottom of the options menu).

zyv:
> > This may be controversial, but I think it's important and useful to show what the values are somehow, and I couldn't find a better place. Also, it is indeed related to appearance...
> 
> i think it would be better to put it in a separate dialog under a menu item "about terminal" (such items are usually in the "help" menu, but i guess mc could have it in a separate section at the top or bottom of the options menu).

Yes, I would generally agree, which is why I said it might be controversial... but I really don't like the idea of introducing yet another box for just two pieces of information, and the appearance box seems topically related and quite empty to me.

Maybe it would be a good idea to merge "Appearance..." with "Layout..." and move the codepage information into a separate box as you describe, especially if there is something else that can be shown there. Do you have any other detected parameters in mind that would be good to show in such a dialog?

I would suggest merging this as is and opening another ticket to make a new box and merge the other two if there is more to show.

ossilator:
> I really don't like the idea of introducing yet another box for just two pieces of information,

i know that feeling, but objectively it's a bad reason.

> and the appearance box seems topically related [...].

not really. it looks out of place, and for someone looking for the info, it would have terrible discoverability.

> Maybe it would be a good idea to merge "Appearance..." with "Layout..."

probably not. firstly, these option groups operate on distinctly different aspects of the UI, so the separation actually makes sense. secondly, when #3169 gets addressed, the appearance dialog will become somewhat crowded.

> Do you have any other detected parameters in mind that would be good to show in such a dialog?

it probably wouldn't be a bad idea to generalize it into "about ...", and include the `mc -V` contents, in addition to moving (parts of) the top-level help screen there.

> I would suggest merging this as is and opening another ticket to make a new box and merge the other two if there is more to show.

i don't like that approach, as it adds momentum to a bad solution.

zyv:
I have created #4704.

Issue #4205, opened 2021-02-17
Re-organize the mc file manager and editor pull-down menus


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4205 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `ossi` (@ossilator) |

<p>
the editor's menus have grown over time into an incoherent mess and need a complete reorganization imo.<br />
</p>
<p>
this should be understood as a continuation of #320, #334, and #408.<br />
</p>
<p>
i'll come up with concrete proposals later unless someone beats me to it.<br />
</p>

ossilator:
let's generalize this into "mc's menus" (as the linked issues mostly were).

ossilator:
let's start with the main mc menu:

left/right menus:
- it's not really clear that the first section is a radio group. conventionally, one would add a checkmark on the left of the current mode's item, but that would require a whole new concept and feature set in the menus.
- the "encoding ..." item seems just wrong. is it actually side-specific? if so, then this is extremely confusing/error prone, and should be changed. the (consolidated) item belongs into the options menu.

file menu:
- it would be really a lot clearer to make the main f-keys (3-8) adjacent. these are a core aspect of the `nc` legacy, and should be treated as such.
- the "view file ..." entry is weird, and mostly useless. it should be available under shift-f3, kinda consistently with shift-f4. how to make it discoverable without it being in the menu is another question, but that's a pre-existing problem for other functions, and thus requires a solution regardless.
- the "quick cd" item doesn't fit, because it doesn't apply to the current entry. maybe it fits into left/right (because these have the navigation-related sections), otherwise command.
- the selection-related section doesn't really fit here. more conventionally, it would be in an "edit" menu. maybe "view". or "panel".
  - it's missing a plain "select current" item, which is kinda confusing.

command menu:
- the swap and switch panels items don't really belong in there. again something for the new "panel" (or whatever) menu.
- the "show directory sizes" item is kinda misplaced as well. firstly, it being a one-shot action, it should be named "calculate dir sizes" ("show" is conventionally used for toggles; see also the checkmark idea above). secondly, it doesn't pop up a window like the other actions, so it feels wrong. it might fit next to the "rescan" item in the left/right menus.
- intuitively, the "edit [...]" items all belong into the options menu

options menu:
[...]

Issue #320, opened 2009-03-30
Reorganization of menus and configuration dialogs


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/320 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `andrew_b` (@aborodin) |

<p>
Currently, there are following configuration dialogs in MC (see <tt>Menu/Options</tt>):<br />
</p>
<p>
Configuration<br />
Layout<br />
Confirmation<br />
Display bits<br />
Learn keys<br />
Virtual FS<br />
</p>
<p>
I think, there is a reason to reorganize this structure to make configuration dialogs less congested.<br />
So, panel options can be moved to new 'Panel settings' dialog.<br />
New dialogs can be created for dialog options and viewer options.<br />
Main editor options dialog can be called via main MC menu.<br />
Anything else.<br />
</p>
<p>
Any opinions are welcome.<br />
</p>

PR #4709, opened 2025-05-18
Ticket #4704: add about box for `mc` and move display info there

### LINKED ISSUE #4704: Create "About..." dialog showing mc info as in mcedit

As a follow up to #4702, we would like to have a place to put various run and build time information about mc.

An "About..." box is added as the last item in the "Options" pull-down menu. It should contain the codepage information from "Appearance..." and maybe something from `mc -V` (or a reference to `mc -V`).

Extending `mc -V` is not necessarily helpful.

### Additional context

_No response_

zyv:
## Proposed changes

As discussed in the related tickets, this adds an "About..." box to `mc`.

Screenshot 2025-05-18 at 10 56 07

The menu entry is positioned so that it does not interfere with established patterns, such as selecting the last item to save the setup.

Screenshot 2025-05-18 at 10 59 42

Relates to: #4702
Resolves: #4704

### DIFF
--- a/src/filemanager/boxes.c
+++ b/src/filemanager/boxes.c
@@ -493,7 +493,7 @@ about_box (void)
     char *label_cp_display;
     char *label_cp_source;
 
-    char *version = g_strdup_printf ("Midnight Commander %s", mc_global.mc_version);
+    char *version = g_strdup_printf ("%s %s", PACKAGE_NAME, mc_global.mc_version);
 
     const char *name_cp_display =
         ((codepage_desc *) g_ptr_array_index (codepages, mc_global.display_codepage))->name;
@@ -512,7 +512,7 @@ about_box (void)
         QUICK_SEPARATOR (TRUE),
         QUICK_LABEL (N_ ("Classic terminal file manager inspired by Norton Commander."), NULL),
         QUICK_SEPARATOR (FALSE),
-        QUICK_LABEL (N_ ("Copyright (C) 1996-2025 the Free Software Foundation"), NULL),
+        QUICK_LABEL (PACKAGE_COPYRIGHT, NULL),
         QUICK_SEPARATOR (TRUE),
         QUICK_LABEL (label_cp_display, NULL),
         QUICK_LABEL (label_cp_source, NULL),
--- a/src/filemanager/boxes.c
+++ b/src/filemanager/boxes.c
@@ -488,6 +488,64 @@ task_cb (WButton *button, int action)
 /* --------------------------------------------------------------------------------------------- */
 
 void
+about_box (void)
+{
+    char *label_cp_display;
+    char *label_cp_source;
+
+    char *version = g_strdup_printf ("Midnight Commander %s", mc_global.mc_version);
+
+    const char *name_cp_display =
+        ((codepage_desc *) g_ptr_array_index (codepages, mc_global.display_codepage))->name;
+
+    const char *name_cp_source = mc_global.source_codepage >= 0
+        ? ((codepage_desc *) g_ptr_array_index (codepages, mc_global.source_codepage))->name
+        : N_ ("No translation");
+
+    label_cp_display =
+        g_strdup_printf ("%s: %s", N_ ("Detected display codepage"), name_cp_display);
+    label_cp_source =
+        g_strdup_printf ("%s: %s", N_ ("Selected source (file I/O) codepage"), name_cp_source);
+
+    quick_widget_t quick_widgets[] = {
+        QUICK_LABEL (version, NULL),
+        QUICK_SEPARATOR (TRUE),
+        QUICK_LABEL (N_ ("Classic terminal file manager inspired by Norton Commander."), NULL),
+        QUICK_SEPARATOR (FALSE),
+        QUICK_LABEL (N_ ("Copyright (C) 1996-2025 the Free Software Foundation"), NULL),
+        QUICK_SEPARATOR (TRUE),
+        QUICK_LABEL (label_cp_display, NULL),
+        QUICK_LABEL (label_cp_source, NULL),
+        QUICK_START_BUTTONS (TRUE, TRUE),
+        QUICK_BUTTON (N_ ("&OK"), B_ENTER, NULL, NULL),
+        QUICK_END,
+    };
+
+    WRect r = { -1, -1, 0, 40 };
+
+    quick_dialog_t qdlg = {
+        .rect = r,
+        .title = N_ ("About"),
+        .help = "[Overview]",
+        .widgets = quick_widgets,
+        .callback = NULL,
+        .mouse_callback = NULL,
+    };
+
+    quick_widgets[0].pos_flags = WPOS_KEEP_TOP | WPOS_CENTER_HORZ;
+    quick_widgets[2].pos_flags = WPOS_KEEP_TOP | WPOS_CENTER_HORZ;
+    quick_widgets[4].pos_flags = WPOS_KEEP_TOP | WPOS_CENTER_HORZ;
+
+    (void) quick_dialog (&qdlg);
+
+    g_free (version);
+    g_free (label_cp_display);
+    g_free (label_cp_source);
+}
+
+/* --------------------------------------------------------------------------------------------- */
+
+void
 configure_box (void)
 {
     const char *pause_options[] = {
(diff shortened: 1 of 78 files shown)


src/filemanager/boxes.c
```diff
@@ -487,6 +487,64 @@ task_cb (WButton *button, int action)
...
+        g_strdup_printf ("%s: %s", N_ ("Detected display codepage"), name_cp_display);
+    label_cp_source =
+        g_strdup_printf ("%s: %s", N_ ("Selected source (file I/O) codepage"), name_cp_source);
+
+    quick_widget_t quick_widgets[] = {
+        QUICK_LABEL (version, NULL),
+        QUICK_SEPARATOR (TRUE),
+        QUICK_LABEL (N_ ("Classic terminal file manager inspired by Norton Commander."), NULL),
+        QUICK_SEPARATOR (FALSE),
+        QUICK_LABEL (N_ ("Copyright (C) 1996-2025 the Free Software Foundation"), NULL),
```
mc-worker: Copyright years are updated using maint/update-years.sh. Please add this file there.

src/filemanager/filemanager.c
```diff
@@ -327,6 +327,8 @@ create_options_menu (void)
 #ifdef ENABLE_VFS
     entries = g_list_prepend (entries, menu_entry_new (_ ("&Virtual FS..."), CK_OptionsVfs));
 #endif
+    entries = g_list_prepend (entries, menu_separator_new ());
+    entries = g_list_prepend (entries, menu_entry_new (_ ("A&bout..."), CK_About));
```
mc-worker: What about to make this item last in menu?

ossilator (commented):
the placement of the menu item makes no sense.
well, the more i look at it, the more i think that the entire menu structure makes no sense at all. yeah, i know, attempts at improving it have been made, but ...
i didn't find an issue for that, so maybe create one, unless i'm just blind?

given the current structure, the bottom of the command menu is probably the least confusing place. dunno.

src/filemanager/filemanager.c
zyv: I explained why I didn't make it the last item in the PR description. I'm pretty sure someone would complain that it breaks selecting the "Save Setup" item. Also, the editor has it second to last.

Screenshot 2025-05-18 at 12 30 23

I don't have a strong opinion on that, though. If you think it's better, I can move it.

zyv:
> the placement of the menu item makes no sense. well, the more i look at it, the more i think that the entire menu structure makes no sense at all. yeah, i know, attempts at improving it have been made, but ... i didn't find an issue for that, so maybe create one, unless i'm just blind?

I think there was a ticket for the file manager menus, but it seems that Andrew resolved what he thought was problematic: #320. There is an open ticket (by you) for the editor menus: #4205.

mc-worker:
> [Use macros for copyright and package name to avoid duplication](https://github.com/MidnightCommander/mc/pull/4709/commits/af38fc5bf87a7b9c65a5427a4a73074340b4ed27)

Will gettext find this macro?

src/filemanager/filemanager.c
ossilator: > I'm pretty sure someone would complain that it breaks selecting the "Save Setup" item.

possibly, but i wouldn't worry about that too much.

i'm more concerned about the submenu being just wrong. this is probably the last place i'd look in.

zyv: > i'm more concerned about the submenu being just wrong. this is probably the last place i'd look in.

Wait, you did suggest that, didn't you?

> i think it would be better to put it in a separate dialog under a menu item "about terminal" (such items are usually in the "help" menu, but i guess mc could have it in a separate section at the top or bottom of the options menu).

zyv:
> > [Use macros for copyright and package name to avoid duplication](https://github.com/MidnightCommander/mc/pull/4709/commits/af38fc5bf87a7b9c65a5427a4a73074340b4ed27)
> 
> Will gettext find this macro?

No, for some reason, it can't find it. But it can if it's made into a string constant. I'm not sure if we want it translated, but that's how it was before.

src/filemanager/boxes.c
zyv: Thank you. I have changed that.

Actually, I think it would be a good idea to throw away our license blurbs and replace them with SPDX tags. However, I'm not sure if I have the strength to take on that project.

https://github.com/spdx/spdx-examples/blob/master/software/example1/content/src/hello.c

```c
/* SPDX-License-Identifier: GPL-3.0-or-later */
/* Copyright Free Software Foundation, Inc. and contributors to the GNU Midnight Commander project. */
```

https://www.linuxfoundation.org/blog/blog/copyright-notices-in-open-source-software-projects

Years are no longer required by US copyright law, and most files have incorrect lists of contributors. If necessary, it can be reconstructed from version control in the most accurate and reliable way.

If we change this, then we won't have to maintain the years and incorrect lists of contributors, which I think has become completely absurd. Thanks to the SPDX tags, copyright can be correctly deduced automatically.

What do you think?

src/filemanager/filemanager.c
ossilator: > Wait, you did suggest that, didn't you?

well, the suggested location was obviously rather noncommittal. :innocent: 

i suppose we should fix #4205 before finalizing this.

zyv: Oh no! There are so many open tickets and so few resources to work on them. I have nothing against #4205, but I don't know when I'll find the time to work on it. In any case, it looks like a major project to me that should be well thought out; otherwise, it will require another follow-up and only cause more frustration due to continuous breakage.

I prefer merging something that's kind of finished, a small improvement, or at least something that isn't actively breaking things or making them much worse, and doing the next step later. Having pull requests pending because of dependencies on other things that ideally should or could be done in the same area is demotivating and discourages me from attempting to do anything at all.

ossilator: dunno, it doesn't seem that monumental to me.

but fair enough. putting it right above exit into the file menu like in the editor seems like a sensible quick fix.

zyv: In its current form, this makes "File" menu overrun 80x24.

src/filemanager/boxes.c
mc-worker: I have nothing against SPDX tags.

> Years are no longer required by US copyright law

And GNU?

https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

zyv: > And GNU?

TL;DR: As long as we want it and know what we're doing and why, I don't think it matters what they think. I think it's a waste of time and doesn't offer any benefits, so I would change it.

I believe that GNU bases this guidance on two main points:

1. Is the copyright statement legally valid?
  * Obviously, they want the statements to be foolproof. If the statement is invalid, the software will be considered public domain, and GPL license restrictions (e.g., not selling without providing the source code) won't apply.
  * GNU stems from the US, which didn't join the Berne Convention until 1989. Therefore, if you omitted the initial year in 1987, your claim would have been invalid.
  * This is why they want an initial year in the statement.
2. Can copyright durations be easily derived from the statement?
  * The copyright is limited in time (currently, you can assume ~100 years), meaning that 100 years after the last significant modification, the software enters the public domain, and license restrictions no longer apply.
  * This is why they want you to update the years after every major change, so the GPL's coverage lasts as long as possible.

The first point only concerned the US, and it hasn't mattered in the US for the last 35 years. The same applies to all signatory countries of the Berne Convention.

The second point means that if we stop updating the years, then, in 2125, someone could write a derivative file manager and sell it without providing the source code under the GPL.

However, as I said, the authors and the last significant change can be deduced from the version control. This information would be much more accurate than what is in the file.

Honestly, I'm skeptical that the human race will still exist if things continue as they are, but if that happens, I'm pretty certain that the ability to sell Midnight Commander without providing the source code won't be at the top of the world's agenda.

zyv: @aborodin so what is your opinion on that?

mc-worker: Ok but that is not quite relevant to the this issue. Let's do that in the cleanup or even in the separate issue.

zyv: That's fine. I just wanted to either reject this discussion or create a new ticket. I'll create a new ticket and merge this one when I get the chance.

zyv: #4713

PR #4722, opened 2025-06-09
Ticket #4720: delete resource forks on macOS last

### LINKED ISSUE #4720: Can't delete folder recursively due to MacOS resource forks (`._*` files)

```shell
GNU Midnight Commander 4.8.33
Gebaut mit GLib 2.82.4
Gebaut mit S-Lang 2.3.3 mit terminfo-Datenbank
Gebaut mit libssh2 1.11.1
Mit eingebautem Editor
Mit Subshell-Unterstützung als Standard
Mit Unterstützung für Hintergrundoperationen
Mit Mausunterstützung für xterm
Mit Internationalisierungsunterstützung
Mit Unterstützung mehrerer Codepages
Virtuelles Dateisystem:
Datentypen:
```

### Operating system

```shell
Darwin Mac.fritz.box 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041 arm64
```

### How to reproduce

Deleting folder with some files and their pesty ._* counterparts created by MacOS
(recursive delete = yes)

### Expected behavior

Folder get's deleted

### Actual behavior

n files reside in this folder
having to delete this folder n+1 times until it's really gone
(recursive delete = yes)

### Additional context

I think it has to something to do with this stupid ._* files, which MacOS creates for every regular file.
When I get rid of these ._* files before and try to delete the folder, it works as expected.

zyv:
## Proposed changes

Resolves: #4720

### DIFF
--- a/lib/fs.h
+++ b/lib/fs.h
@@ -108,6 +108,11 @@
 #define DIR_IS_DOT(x)     ((x)[0] == '.' && (x)[1] == '\0')
 #define DIR_IS_DOTDOT(x)  ((x)[0] == '.' && (x)[1] == '.' && (x)[2] == '\0')
 
+/* On macOS, if a file system doesn't support metadata, it is automatically stored in shadow files
+ * called resource forks. The names of these files begin with `._`. They are managed automatically
+ * by the OS. */
+#define FILE_IS_RESOURCE_FORK(x) ((x)[0] == '.' && (x)[1] == '_' && (x)[2] != '\0')
+
 /*** enums ***************************************************************************************/
 
 /*** structures declarations (and typedefs of structures)*****************************************/
--- a/lib/fs.h
+++ b/lib/fs.h
@@ -108,9 +108,9 @@
 #define DIR_IS_DOT(x)     ((x)[0] == '.' && (x)[1] == '\0')
 #define DIR_IS_DOTDOT(x)  ((x)[0] == '.' && (x)[1] == '.' && (x)[2] == '\0')
 
-/* On macOS, if a file system doesn't support metadata, it is automatically stored in shadow files
- * called resource forks. The names of these files begin with `._`. They are managed automatically
- * by the OS. */
+/* On macOS, if a file system doesn't support resource forks, metadata is automatically stored in
+ * shadow files beginning with `._`. They are automatically managed by the OS and will reappear if
+ * deleted. */
 #define FILE_IS_RESOURCE_FORK(x) ((x)[0] == '.' && (x)[1] == '_' && (x)[2] != '\0')
 
 /*** enums ***************************************************************************************/
--- a/lib/fs.h
+++ b/lib/fs.h
@@ -108,6 +108,11 @@
 #define DIR_IS_DOT(x)     ((x)[0] == '.' && (x)[1] == '\0')
 #define DIR_IS_DOTDOT(x)  ((x)[0] == '.' && (x)[1] == '.' && (x)[2] == '\0')
 
+/* On macOS, if a file system doesn't support resource forks, metadata is automatically stored in
+ * shadow files beginning with `._`. They are automatically managed by the OS and will reappear if
+ * deleted. */
+#define FILE_IS_RESOURCE_FORK(x) ((x)[0] == '.' && (x)[1] == '_' && (x)[2] != '\0')
+
 /*** enums ***************************************************************************************/
 
 /*** structures declarations (and typedefs of structures)*****************************************/
--- a/src/filemanager/file.c
+++ b/src/filemanager/file.c
@@ -1479,7 +1479,7 @@ recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath,
         }
         else
         {
-            if (!FILE_IS_RESOURCE_FORK (next->d_name) || delete_resource_forks)
+            if (delete_resource_forks || !FILE_IS_RESOURCE_FORK (next->d_name))
                 return_status = erase_file (ctx, tmp_vpath);
             else
                 return_status = FILE_SKIP;
--- a/src/filemanager/file.c
+++ b/src/filemanager/file.c
@@ -1439,9 +1439,14 @@ try_erase_dir (file_op_context_t *ctx, const vfs_path_t *vpath)
   abort -> cancel stack
   ignore -> warn every level, gets default
   ignore_all -> remove as much as possible
+
+  This function should either be called with delete_resource_forks = TRUE always, or called twice,
+  first with FALSE and then with TRUE to make sure that no errors pop up due to resource forks being
+  deleted on macOS and then being re-created on the fly by the OS.
 */
 static FileProgressStatus
-recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath)
+recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath,
+                 const gboolean delete_resource_forks)
 {
     struct vfs_dirent *next;
     DIR *reading;
@@ -1467,9 +1472,18 @@ recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath)
             return FILE_RETRY;
         }
         if (S_ISDIR (buf.st_mode))
-            return_status = recursive_erase (ctx, tmp_vpath);
+        {
+            return_status = recursive_erase (ctx, tmp_vpath, FALSE);
+            if (return_status != FILE_ABORT)
+                return_status = recursive_erase (ctx, tmp_vpath, TRUE);
+        }
         else
-            return_status = erase_file (ctx, tmp_vpath);
+        {
+            if (!FILE_IS_RESOURCE_FORK (next->d_name) || delete_resource_forks)
+                return_status = erase_file (ctx, tmp_vpath);
+            else
+                return_status = FILE_SKIP;
+        }
         vfs_path_free (tmp_vpath, TRUE);
     }
     mc_closedir (reading);
@@ -3351,7 +3365,11 @@ erase_dir (file_op_context_t *ctx, const vfs_path_t *vpath)
         // not empty
         error = query_recursive (ctx, vfs_path_as_str (vpath));
         if (error == FILE_CONT)
-            error = recursive_erase (ctx, vpath);
+        {
+            error = recursive_erase (ctx, vpath, FALSE);
+            if (error != FILE_ABORT)
+                error = recursive_erase (ctx, vpath, TRUE);
+        }
         return error;
     }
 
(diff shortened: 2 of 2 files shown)


zyv:
Could you please test this, @ollixx77?

@aborodin I'm still not sure if I'm handling return codes correctly.

lib/fs.h
```diff
@@ -108,6 +108,11 @@
 #define DIR_IS_DOT(x)     ((x)[0] == '.' && (x)[1] == '\0')
 #define DIR_IS_DOTDOT(x)  ((x)[0] == '.' && (x)[1] == '.' && (x)[2] == '\0')
 
+/* On macOS, if a file system doesn't support metadata, it is automatically stored in shadow files
+ * called resource forks. The names of these files begin with `._`. They are managed automatically
```
ossilator: that's kinda misleading. see https://en.wikipedia.org/wiki/Resource_fork .

src/filemanager/file.c
```diff
@@ -1467,9 +1472,18 @@ recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath)
...
-            return_status = recursive_erase (ctx, tmp_vpath);
+        {
+            return_status = recursive_erase (ctx, tmp_vpath, FALSE);
+            if (return_status != FILE_ABORT)
+                return_status = recursive_erase (ctx, tmp_vpath, TRUE);
+        }
         else
-            return_status = erase_file (ctx, tmp_vpath);
+        {
+            if (!FILE_IS_RESOURCE_FORK (next->d_name) || delete_resource_forks)
```
ossilator: the conditions should be swapped for efficiency in the negative (== normal) case.

src/filemanager/file.c
```diff
@@ -1439,9 +1439,14 @@ try_erase_dir (file_op_context_t *ctx, const vfs_path_t *vpath)
   abort -> cancel stack
   ignore -> warn every level, gets default
   ignore_all -> remove as much as possible
+
+  This function should either be called with delete_resource_forks = TRUE always, or called twice,
```
ossilator: this feels complex and dirty.
my first thought would be to just put files that look like forks into a list, and delete them in a second loop. it's kinda like your sorting idea, just more targeted.

ossilator (changes requested):

ollixx77:
> Could you please test this, @ollixx77?
> 
> @aborodin I'm still not sure if I'm handling return codes correctly.

The original change worked so far.
But took me while to get this all setup to install.

lib/fs.h
zyv: So, what exactly is misleading?

src/filemanager/file.c
zyv: I thought about it, but this looks much more complicated to me. I'm not ready to implement anything like this in the current code base.

I thought the functions would retrieve a list of items to delete from the panels. However, it's a recursive call tree connected with callbacks to the progress dialog. The easiest and safest method seems to be simply calling the recursion twice.

lib/fs.h
ossilator: it's not the files that are called resource forks. resource (and data) forks are stored in these magic shadow files if the fs doesn't support forks natively.
technically, it's also not metadata ... paradata would fit better, if that was a thing.

src/filemanager/file.c
ossilator: easiest, i guess. but safest ...? this looks like an invitation for double accounting, dialogs asking twice for the same error, etc. it's much harder to guarantee that it works in every case, and that it keeps working.

i'm not sure why you think the separate loop would be much more complicated. it would be a little bit more code (part of the loop body would have to be duplicated or factored out), but seems rather straight-forward.

src/filemanager/file.c
zyv: One would think that AI-powered compilers running on laptops suspended from quadcopters would optimize for that. But you are right; they do not.

Screenshot 2025-06-09 at 09 59 33

lib/fs.h
zyv: Okay, I've revised my comment.

src/filemanager/file.c
zyv: I think it's complicated because it's difficult **for me** to do. Maybe someone can help.

src/filemanager/file.c
```diff
@@ -1467,9 +1472,18 @@ recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath)
             return FILE_RETRY;
         }
         if (S_ISDIR (buf.st_mode))
-            return_status = recursive_erase (ctx, tmp_vpath);
+        {
+            return_status = recursive_erase (ctx, tmp_vpath, FALSE);
+            if (return_status != FILE_ABORT)
```
mc-worker: Please make this compiled conditionally. Users of OSes other than macOS must not feel bad.

src/filemanager/file.c
```diff
@@ -1467,9 +1472,18 @@ recursive_erase (file_op_context_t *ctx, const vfs_path_t *vpath)
...
+                return_status = recursive_erase (ctx, tmp_vpath, TRUE);
+        }
         else
-            return_status = erase_file (ctx, tmp_vpath);
+        {
+            if (delete_resource_forks || !FILE_IS_RESOURCE_FORK (next->d_name))
```
mc-worker: Please make this compiled conditionally.

src/filemanager/file.c
```diff
@@ -3351,7 +3365,11 @@ erase_dir (file_op_context_t *ctx, const vfs_path_t *vpath)
         // not empty
         error = query_recursive (ctx, vfs_path_as_str (vpath));
         if (error == FILE_CONT)
-            error = recursive_erase (ctx, vpath);
+        {
+            error = recursive_erase (ctx, vpath, FALSE);
+            if (error != FILE_ABORT)
```
mc-worker, line 3405: Please make this compiled conditionally.

src/filemanager/file.c
zyv: I don't think it will put users of other operating systems at a disadvantage, but if you think this is the right approach, I can compile it conditionally. I was concerned that it would make the code more difficult to maintain.

ossilator: i also had the thought to make this os-specific. but then, what if it's an nfs mount from a mac?

zyv: > i also had the thought to make this os-specific. but then, what if it's an nfs mount from a mac?

Yes, this is a good argument.

But what about non-macOS users? What is the disadvantage for them? Only if there are files beginning with `._` in the tree will anything be different; in all other cases, there will be no change.

I think this is extremely rare on other systems, and when it does happen, the impact will be negligible, so I would rather drop my fixups; they don't look good.

ossilator:
now i wonder ...

given that these are "shadow" files that always go along with "real" files, and we're talking about a situation of actually running on macos (otherwise the case of the files magically disappearing would be irrelevant), why aren't the files hidden by the os? this looks like a shortcoming that would cause all kinds of problems.

which brings me to the next point: shouldn't for example move/rename also have related code? that would suggest that this should be handled at the vfs level.

but then there is copy. how would copying a file with forks even work on an fs that supports them natively, when they are invisible to regular unix apis? or does the posix layer actually inject the shadow files?

i guess this should be properly researched before going any further with this.

src/filemanager/file.c
zyv: I have rebased the changes on top of the latest master, and the conditional compilation as requested is in 5469c92c0152e54616a605f7f8566580cc5949dc. However, I think that it is wrong to do this. When I access the FAT32 volumes on Mac via mc from a Linux machine, I run into the same problem because mc compiled on Linux doesn't delete resource forks last.

Also, see the argument why the recursive deletion actually doesn't add a performance or other penalty of significance on other systems. So I would rather drop this part of the patch.

zyv:
> now i wonder ...
> 
> given that these are "shadow" files that always go along with "real" files, and we're talking about a situation of actually running on macos (otherwise the case of the files magically disappearing would be irrelevant), why aren't the files hidden by the os? this looks like a shortcoming that would cause all kinds of problems.
> 
> which brings me to the next point: shouldn't for example move/rename also have related code? that would suggest that this should be handled at the vfs level.
> 
> but then there is copy. how would copying a file with forks even work on an fs that supports them natively, when they are invisible to regular unix apis? or does the posix layer actually inject the shadow files?
> 
> i guess this should be properly researched before going any further with this.

I did some basic research on that, and I can't tell why the designers of macOS made the decision to expose the resource forks like this via the POSIX API.

I have discovered that if the files are deleted, then their resource forks are removed by "magic". If the files are copied, the resource forks are apparently taken along (or created anew, it's hard to find out).

If the files are moved, then their resource forks are also magically moved with them. So probably ideally one could implement a similar mechanism for moves, as in, first move everything that doesn't look like a fork, and then there is probably nothing else to move.

However, in cases of copies and moves the situation is much less annoying than in the case of deletion. If the fork is moved first, apparently it gets re-created. And then the newly created one is taken along when the file is moved. If the file is moved, then its fork is moved, and so there is nothing else to move. For copies, it also strangely seems to work somehow.

So, I wouldn't do anything here because it doesn't present a practical problem (apparently), and makes things more complicated. Also, I find it quite annoying that apparently there is no good way to detect forks other than by name, and even then you don't know if it's a valid file or a fork.

Anyways, it has been like that for decades. If this problem gets reported, then we could do something about that. A big issue that I have with our whole system is that we are doing things recursively, so it's not possible to build a list first and then just sort it before executing the command. Changing this is way beyond my time and capabilities at the moment.

zyv:
@mc-worker I'm now working a lot with FAT32 volumes on macOS, and this has become a real problem for me. I would like to merge the current solution because it works and doesn't seem to make any issues.

mc-worker:
> @mc-worker I'm now working a lot with FAT32 volumes on macOS, and this has become a real problem for me. I would like to merge the current solution because it works and doesn't seem to make any issues.

On Linux, if file `._*` is in a directory that is being deleted, the extra error window is raised.
w1
w2
I think, users will be confused.

zyv:
> > @mc-worker I'm now working a lot with FAT32 volumes on macOS, and this has become a real problem for me. I would like to merge the current solution because it works and doesn't seem to make any issues.
> 
> On Linux, if file `._*` is in a directory that is being deleted, the extra error window is raised. I think, users will be confused.

Hmmm, this is very strange... Thank you for testing! It seems that I am doing something wrong with the return codes in our crazy recursive system :( Do you happen to know why or how to schedule files to be deleted last properly?

zyv:
Maybe I should return `FILE_CONT` instead? Could you please give me your test directory?

mc-worker:
> Could you please give me your test directory?

```
mkdir tmp/a tmp/b
touch tmp/a/aaa tmp/a/._aaa tmp/b/bbb tmp/b/._bbb
```
Press `F8` on `tmp`.

JFYI: https://gist.github.com/lpar/1515061

mc-worker:
I think I've fixed this issue.

PR #4724, opened 2025-06-19
subshell: reorder pty/pipe file descriptor acquisition

### LINKED ISSUE #4634: Clean up subshell precmd and PS1 handling
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4634 |
> | **Mentions** | `ossi` (@ossilator), `d3m3t3r` (d3m3t3r@….com) |

<p>
Follow-up for #3748:<br />
</p>
<p>
We want to remove our PS1 / prompt fallbacks for all shells and cleanup fallback precmd handling:<br />
</p>

```
        "_mc_precmd() { "
        "  pwd>&%d; "
        "  kill -STOP $$; "
        "}; "
        "_MC_PRECMD=_mc_precmd; "
        "PS1='$($_MC_PRECMD)'\"${PS1:-'$PWD $ '}\"\n";
```
<blockquote class="citation">
<p>
to actually answer the question whether PS1 can be unset in the first place: probably not. i mean, it can if you manually unset it, but dash actually shows no prompt at all in this case. so arguably, providing even a minimal PS1 fallback is unnecessary.<br />
</p>
</blockquote>
<blockquote class="citation">
<p>
It's definitely quite unlikely an interactive shell not setting PS1. The user would need to mess up something. IMO, the default PS1 could be removed from every 'precmd' code if we check all shells cope with empty PS1 reasonably. For bash, you can test it by adding "unset PS1" into .bashrc. It seems to work fine.<br />
</p>
</blockquote>

---

> add `/bin/lksh` as well, it’s present on many systems (all Debian derivatives, plus others) and is an mksh binary built with POSIX semantics where they cannot be made a run-time option (such as width of the arithmetic data type) 

 _Originally posted by @mirabilos in [9f5d80d](https://github.com/MidnightCommander/mc/commit/9f5d80def20b642cdb3a4a1d1dd34fdbfb8edd22#r160170895)_

mirabilos:
Swap the code allocating the pty and the block getting the pipe for cwd communication, to hopefully get an fd lower than 10 for the latter, to ameliorate a few from the problems mentioned in #4634 as long as there isn’t more global code allocating file descriptors added.

Specifically, this moves the shell’s side of the fd in my local tests on Debian trixie from 10 to 8.

I’ve merely swapped two blocks without changing any of the actual code. Swapping WRITE with READ in line 158ﬀ would gain us another one in this specific instance as it would then use the lower number of the pair.

https://github.com/MidnightCommander/mc/issues/4634#issuecomment-2989003485 mentions a few ideas I have for really actually fixing this in the long term (and ensuring the issue will not reappear) which need more invasive refactoring though.

## Proposed changes

See above. Note it does **not** address #4634, it merely gives us more time to resolve it, as this is currently broken in Debian soon-to-be-stable. (Incidentally, the bug does not always occur; e.g. I can only reproduce it in an X11 session on the device itself but not when ssh’ing in. This likely has to do with other factors that influence which file descriptors `mc` acquires and when.)

## Checklist

(as I merely swapped the order of two code blocks, I’ve not done more extensive testing, other than `strace`ing to compare which fd numbers are assigned in practice; this may still not be enough in some configurations, but we’ll know when it appears)

### DIFF
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -1556,37 +1556,6 @@ init_subshell (void)
         if (mc_global.shell->type == SHELL_NONE)
             return;
 
-        // Open a pty for talking to the subshell
-
-        // FIXME: We may need to open a fresh pty each time on SVR4
-
-#ifdef HAVE_OPENPTY
-        if (openpty (&mc_global.tty.subshell_pty, &subshell_pty_slave, NULL, NULL, NULL))
-        {
-            fprintf (stderr, "Cannot open master and slave sides of pty: %s\n",
-                     unix_error_string (errno));
-            mc_global.tty.use_subshell = FALSE;
-            return;
-        }
-#else
-        mc_global.tty.subshell_pty = pty_open_master (pty_name);
-        if (mc_global.tty.subshell_pty == -1)
-        {
-            fprintf (stderr, "Cannot open master side of pty: %s\r\n", unix_error_string (errno));
-            mc_global.tty.use_subshell = FALSE;
-            return;
-        }
-
-        subshell_pty_slave = pty_open_slave (pty_name);
-        if (subshell_pty_slave == -1)
-        {
-            fprintf (stderr, "Cannot open slave side of pty %s: %s\r\n", pty_name,
-                     unix_error_string (errno));
-            mc_global.tty.use_subshell = FALSE;
-            return;
-        }
-#endif
-
         // Create a pipe for receiving the subshell's CWD
 
         if (mc_global.shell->type == SHELL_TCSH)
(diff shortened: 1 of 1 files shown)


src/subshell/common.c
```diff
@@ -1629,6 +1598,37 @@ init_subshell (void)
             mc_global.tty.use_subshell = FALSE;
             return;
         }
+
+        // Open a pty for talking to the subshell
```
zyv: It would be great if you could add a comment here like

> Do this **after** opening the communication pipes in the forked child to give those a lower file descriptor number (hopefully, below 10). Some subshells have problems redirecting to file descriptors higher than 9.

Your commit message is great, but it would be fantastic to have a short, practical summary here if this code stays like this for longer...

zyv:
> See above. Note it does **not** address #4634, it merely gives us more time to resolve it, as this is currently broken in Debian soon-to-be-stable. (Incidentally, the bug does not always occur; e.g. I can only reproduce it in an X11 session on the device itself but not when ssh’ing in. This likely has to do with other factors that influence which file descriptors `mc` acquires and when.)

Yes, it must be compiled with the `--with-x` option and establishes communication with X11 when run on the device, but not via an SSH session.

src/subshell/common.c
mirabilos: OK, will do. I merely moved the two code blocks, but can do.

mirabilos:
Like this?

zyv (approved):
Yes, thank you! 👍

src/subshell/common.c
```diff
@@ -1629,6 +1598,41 @@ init_subshell (void)
             mc_global.tty.use_subshell = FALSE;
             return;
         }
+
+        // Open a pty for talking to the subshell; do this after
+	// acquiring the pipes, so the latter have a higher chance
```
mc-worker: Please fix an indentation in lines 1603...1606.
And please don't use non-ASCII symbols in comments (line 1604) unless you really need them.

zyv: @mirabilos I wanted to force-push for you to save you having to do it, but you don't allow maintainers to push to your PR. Here is the patch:

[0001-Ticket-4724-subshell-reorder-pty-pipe-file-descripto.patch](https://github.com/user-attachments/files/20845831/0001-Ticket-4724-subshell-reorder-pty-pipe-file-descripto.patch)

zyv: @mirabilos could you please force-push my patch above or allow maintainer edits of this PR?

mirabilos: sorry, I was too busy with the new job, will do that now. What did you do to the indentation for tabs to not work… and to use only ASCII, oh well.

zyv (approved):
@aborodin ok now?

mc-worker:
> @aborodin ok now?

Yes.

zyv:
Thanks to both of you! I would appreciate it if maintainer edits were enabled in the next PRs. It makes small changes much easier and faster instead of having to ask.

mirabilos:
maintainer edits destroy the PGP signature though

PR #4728, opened 2025-07-05
Add some patterns to gitignore

krobelus:
There are some build output files that are not gitignore; most are generated
by "make", except for *.trs which is from "make -C tests/src/editor/ check",
and "conftest.c" and "confdefs.h" which is from a configure step.

The missing ignores makes it very easy to accidentally commit files
when working with the Jujutsu VCS which tracks files automatically.
Fix this ignoring all patterns I've seen.

Obviously it would be better to have a dedicated build directory.  Also,
files such as "conftest.c" that don't outlive the build command could be
moved to /tmp.


zyv:
I don't mind updating the `.gitignore` patterns, but are you actually using a VPATH build? I don't understand your comment about a separate build directory.

krobelus:
Ah, I didn't know about VPATH. That helps, yeah.
I hadn't used autotools much, and never with out-of-tree builds.

zyv (approved):

aborodin:
What is `/confaCeovi/{defines,subs,subs1}.awk`? Can't find it in autotools.

krobelus:
On Sat, Jul 05, 2025 at 06:08:37AM -0700, Andrew Borodin wrote:
> aborodin left a comment (MidnightCommander/mc#4728)
> 
> What is `/confaCeovi/{defines,subs,subs1}.awk`? Can't find it in autotools.

Ah, forgot about that one.
I'm not sure, maybe a "mktemp confXXXXXX"... Not yet sure where's the source for this.

zyv:
> What is `/confaCeovi/{defines,subs,subs1}.awk`? Can't find it in autotools.

Good point! It's in `$ac_tmp`, which is generated with `tmp=./conf$$-$RANDOM`, so the pattern should actually be `conf*/...`. Could you please fix this, @krobelus?

zyv:
> Ah, forgot about that one. I'm not sure, maybe a "mktemp confXXXXXX"... Not yet sure where's the source for this.

```shell
# Have a temporary directory for convenience.  Make it in the build tree
# simply because there is no reason against having it here, and in addition,
# creating and moving files from /tmp can sometimes cause problems.
# Hook for its removal unless debugging.
# Note that there is a small window in which the directory will not be cleaned:
# after its creation but before its name has been assigned to '$tmp'.
$debug ||
{
  tmp= ac_tmp=
  trap 'exit_status=$?
  : "${ac_tmp:=$tmp}"
  { test ! -d "$ac_tmp" || rm -fr "$ac_tmp"; } && exit $exit_status
' 0
  trap 'as_fn_exit 1' 1 2 13 15
}
# Create a (secure) tmp directory for tmp files.

{
  tmp=`(umask 077 && mktemp -d "./confXXXXXX") 2>/dev/null` &&
  test -d "$tmp"
}  ||
{
  tmp=./conf$$-$RANDOM
  (umask 077 && mkdir "$tmp")
} || as_fn_error $? "cannot create a temporary directory in ." "$LINENO" 5
ac_tmp=$tmp
```

mc-worker (approved):

Issue #4035, opened 2019-11-23
Compile failure on OS X 10.9


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4035 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `kroleg` (krolega@….ru) |
> | **Mentions** | `egmont` (@egmontkob) |

<p>
I was unable to compile MC on OS X 10.9. The error message is the following:<br />
</p>

```
5 warnings generated.
warning: unknown warning option '-Wfloat-conversion' [-Wunknown-warning-option]
warning: unknown warning option '-Wformat-signedness'; did you mean '-Wformat-non-iso'?
      [-Wunknown-warning-option]
warning: unknown warning option '-Wmaybe-uninitialized'; did you mean '-Wuninitialized'?
      [-Wunknown-warning-option]
warning: unknown warning option '-Wmissing-parameter-type' [-Wunknown-warning-option]
warning: unknown warning option '-Wunused-but-set-variable'; did you mean '-Wunused-const-variable'?
      [-Wunknown-warning-option]
5 warnings generated.
duplicate symbol _menu_map in:
    ./.libs/libinternal.a(keybind-defaults.o)
    ./.libs/libinternal.a(menu.o)
ld: 1 duplicate symbol for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[3]: *** [mc] Error 1
make[2]: *** [all-recursive] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
bash-4.4$ 
```
<p>
The full log is attached.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [mclog.txt](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/4035/mclog.txt) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/4035/mclog.txt)) by `kroleg` (krolega@….ru) on <span title="2019-11-23T13:18:11+00:00">`Nov 23, 2019 at 13:18 UTC`</span>
[...]

PR #4729, opened 2025-07-05
Fix %pkglibexecdir% leaking into man pages in some cases

krobelus:
The generated "doc/man/es/Makefile" contains

	pkglibexecdir = $(libexecdir)/mc
	libexecdir = ${exec_prefix}/libexec
	SED_PARAMETERS = \
		... \
		-e "s{%libexecdir%{${exec_prefix}/libexec{g" \
		-e "s{%pkglibexecdir%{@pkglibexecdir@{g" \

I think "@pkglibexecdir@" is an Automake macro that should have been expanded
when generating this Makefile.

When I install from source, this leads to

	$ man mc | grep bash
	shell to the last directory Midnight Commander was in. Source the file @pkglibexecdir@/mc.sh (bash and zsh

Fix that by expanding earlier. We should also be able to use the Make variable
("$(pkglibexecdir)", which is defined in the Makefile) but I couldn't get
that to work.

Ref: https://github.com/MidnightCommander/mc/pull/4726/commits/7376bc6699c7372fce6186ad6691a9a777777292#r2186982049

### DIFF
--- a/doc/man/date-of-man-include.am
+++ b/doc/man/date-of-man-include.am
@@ -1,10 +1,10 @@
 SED_PARAMETERS = \
 	-e "s/%DATE_OF_MAN_PAGE%/$${MAN_DATE}/g" \
-	-e "s/%MAN_VERSION%/@MAN_VERSION@/g" \
-	-e "s{%sysconfdir%{@sysconfdir@{g" \
-	-e "s{%libexecdir%{@libexecdir@{g" \
-	-e "s{%pkglibexecdir%{@pkglibexecdir@{g" \
-	-e "s{%pkgdatadir%{$(datadir)/@PACKAGE@{g"
+	-e "s/%MAN_VERSION%/$(MAN_VERSION)/g" \
+	-e "s{%sysconfdir%{$(sysconfdir){g" \
+	-e "s{%libexecdir%{$(libexecdir){g" \
+	-e "s{%pkglibexecdir%{$(pkglibexecdir){g" \
+	-e "s{%pkgdatadir%{$(pkgdatadir){g"
 
 MAN_DATE_CMD = \
 	LC_ALL=$(DATE_LANG) @PERL_FOR_BUILD@ -CS -MPOSIX -e '\


zyv:
I think this broke during the refactoring in 4583127644d155ee5c741879aae660fe93249cb0.

Therefore, I would very much like to try to understand why the variable is not being expanded in the same way as the others and fix it "properly", or if this is not possible, add a comment explaining why. If I commit it as it is, the next person will certainly encounter this problem and "fix" it again :(

zyv:
I'm not sure what the problem is with the make variables — they seem to work just fine for me!

https://github.com/MidnightCommander/mc/actions/runs/16088173675/job/45401844503?pr=4729#step:6:907

zyv (approved):

doc/man/date-of-man-include.am
```diff
@@ -1,10 +1,10 @@
...
-	-e "s/%MAN_VERSION%/@MAN_VERSION@/g" \
-	-e "s{%sysconfdir%{@sysconfdir@{g" \
-	-e "s{%libexecdir%{@libexecdir@{g" \
-	-e "s{%pkglibexecdir%{@pkglibexecdir@{g" \
-	-e "s{%pkgdatadir%{$(datadir)/@PACKAGE@{g"
+	-e "s/%MAN_VERSION%/$(MAN_VERSION)/g" \
+	-e "s{%sysconfdir%{$(sysconfdir){g" \
+	-e "s{%libexecdir%{$(libexecdir){g" \
+	-e "s{%pkglibexecdir%{$(pkglibexecdir){g" \
+	-e "s{%pkgdatadir%{$(pkgdatadir){g"
```
krobelus, line 7: LGTM, that's better.
I think the reason why it didn't work for me was that I used curly braces (`${FOO}`) which is only seen by the shell..
And also curly braces probably broke the sed command because weirdly, `{` is used as separator.
I was confused by the `$${MAN_DATE}` above.. I don't understand why we're delaying expansion of `MAN_DATE`, maybe because it makes a small difference

mc-worker (approved):

doc/man/date-of-man-include.am
zyv: This is because it is computed when the rule is executed, rather than when the build system is bootstrapped and/or the project is configured.

krobelus:
> This is because it is computed when the rule is executed, rather than when the build system is bootstrapped and/or the project is configured.

but MAN_DATE is hardcoded in the generated Makefile. Maybe I misread (could be overridden)

zyv:
> This is because it is computed when the rule is executed, rather than when the build system is bootstrapped and/or the project is configured.
> but MAN_DATE is hardcoded in the generated Makefile. Maybe I misread (could be overridden)

Okay, that is confusing. Maybe we should just remove the fallback? I'm not sure if it actually works. In fact, I don't think it does.

```diff
master % git diff
diff --git a/configure.ac b/configure.ac
index 28a595c2b..e908fd031 100644
--- a/configure.ac
+++ b/configure.ac
@@ -495,9 +495,6 @@ dnl ############################################################################
 dnl Documentation
 dnl ############################################################################
 
-MAN_DATE="$(LC_ALL=C date "+%B %Y")"
-AC_SUBST(MAN_DATE)
-
 dnl Determine which help translations we want to install.
 ALL_DOC_LINGUAS="es hu it pl ru sr"
```

If one needs to force the date, such as for reproducible builds, one could use the following:

```
make MAN_DATE_CMD="echo w00t"
```

zyv:
See also #4035, specifically https://github.com/MidnightCommander/mc/issues/4035#issuecomment-2690090658.

While we're at it, maybe we could also add support for `SOURCE_DATE_EPOCH`?

PR #4753, opened 2025-08-24
panel: add git branch show on panel for current directory

KuzinAndrey:
## Proposed changes

Show current git branch on panel for current working directory if it is a git project.

## Checklist

### DIFF
--- a/src/filemanager/layout.c
+++ b/src/filemanager/layout.c
@@ -101,6 +101,9 @@ gboolean xterm_title = TRUE;
 /* Set to show free space on device assigned to current directory */
 gboolean free_space = TRUE;
 
+/* Set to show current branch for directory with git */
+gboolean git_branch = TRUE;
+
 /* The starting line for the output of the subprogram */
 int output_start_y = 0;
 
@@ -134,6 +137,7 @@ typedef struct
     gboolean message_visible;
     gboolean xterm_title;
     gboolean free_space;
+    gboolean git_branch;
     int output_lines;
 } layout_t;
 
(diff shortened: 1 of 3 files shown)


zyv:
Unfortunately, I don't think this is suitable for inclusion in the master branch. To me, this is precisely the type of situation that calls for a scripting engine like mc^2. Whether or not we'll get one anytime soon is a different story.

This functionality is specific to one VCS only. If we're going to do anything like that, why not support jj, hg, svn, etc.? I would argue that it would be better to implement an output of a custom command on the panel. But then again, we come back to the scripting engine argument. Besides, this can easily be achieved with a custom prompt.

Sorry to disappoint you, but thank you for your submission.

KuzinAndrey:
Yes, you are wright. This feature is very useful for my job, as package maintaner in custom Linux distro. As in zsh prompt i want to view active git branch name for current working directory (but i don't want make this in prompt to save free space on terminal). 

This is really good idea to show something on panel by running user defined command, may be sometime you make this for all of us.

zyv:
> This is really good idea to show something on panel by running user defined command, may be sometime you make this for all of us.

Thank you for understanding that we can't accept this PR as it is. Unfortunately, I can't say whether or when mc² is going to happen.

I would be interested to know what @aborodin thinks about the idea of creating a feature that outputs a custom user command next to the free space block.

If the OP is interested in substantially extending the PR to make it more generic, we could take that approach instead of making everyone wait for something else to come along, if it comes at all.

What do you think of this idea? Do we already have any similar requests?

PR #4755, opened 2025-08-25
Add escaping and ignore invalid entries in the filepos file

### LINKED ISSUE #4752: mc crashes upon `M-E` (View/Edit history) if visited filenames contain special chars

  - this statement refers to my searching for
    `is:issue -reason:completed crash history (view OR edit OR file OR "M-E")`

```shell
GNU Midnight Commander 4.8.33-334-g9ec954aaf
Built with GLib 2.80.0 (using GLib 2.80.0)
With builtin editor
With mouse support on xterm
With support for X11 events
Virtual File Systems:
Data types:
```

### Operating system

```shell
Linux codespaces-08a020 6.8.0-1030-azure #35~22.04.1-Ubuntu SMP Mon May 26 18:08:30 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

  - this statement refers to `mc 4.8.33-334-g9ec954aaf`

### How to reproduce

1. In `mc`, navigate to any writable directory in a file system that supports the full range of 254 ASCII characters allowed in linux filenames (i. e. all but `\0` and `/`), e. g. `ext2`, `ext4`, `btrfs`, `zfs`.
2. There, run the shell command `touch $'foo\n\nbar'` to create a file with two newlines in its name.
3. In one of the two panels in `mc`, find and select the file created in 2., then press `F4` to open it in `mcedit`. (I haven't tested whether this bug is triggered with external editors, but it is triggered if you open the file via `mcedit ./$'foo\n\nbar'` directly, and start `mc` afterwards.)
5. In `mcedit`, add some content, press `F2` to save, then `F10` to quit.
6. In `mc`, press `M-E`.

### Expected behavior

You should be shown the list of viewed or edited files, with all special chars either
- properly escaped as in `foo\n\nbar` (preferable), or
- replaced by `.` (as currently done in `mc`'s directory listings).

### Actual behavior

`mc` crashes, producing this final printout:
```
../../../glib/gmem.c:106: failed to allocate 18446603351525610577 bytes
Trace/breakpoint trap (core dumped)
```

### Additional context

The crash persists in subsequent `mc` sessions.

It can only be remedied by editing `~/.local/share/mc/filepos` which, in the crashing state, may look like:
```
/home/codespace/tmp/foo

bar 2;0;5
/home/codespace/tmp/test.txt 1;0;0
/workspaces/mc/configure 8;0;224
```
Apparently, its contents needs to be changed so every filename is restricted to one line, e. g.:
```
/home/codespace/tmp/foo\n\nbar 2;0;5
/home/codespace/tmp/test.txt 1;0;0
/workspaces/mc/configure 8;0;224
```
I haven't found a valid escaping variant that actually allows `mc` to match the existing file, but selecting the corrected entry in the `M-E` menu opens the containing directory at least.

zyv:
## Proposed changes

I've extracted `file_history_parse_entry` as a `void` function, instead of a function returning a `file_history_data_t` to avoid duplicating a NULL pointer check in the caller. Sigh... Unit/regression test is attached.

Resolves: #4752

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -980,7 +980,7 @@ void
 load_file_position (const vfs_path_t *filename_vpath, long *line, long *column, off_t *offset,
                     GArray **bookmarks)
 {
-    char *fn;
+    char *fn, *filename_str;
     FILE *f;
     char buf[MC_MAXPATHLEN + 100];
     const size_t len = vfs_path_len (filename_vpath);
@@ -1001,13 +1001,15 @@ load_file_position (const vfs_path_t *filename_vpath, long *line, long *column, 
     if (bookmarks != NULL)
         *bookmarks = g_array_sized_new (FALSE, FALSE, sizeof (size_t), MAX_SAVED_BOOKMARKS);
 
+    filename_str = str_escape (vfs_path_as_str (filename_vpath), -1, "", TRUE);
+
     while (fgets (buf, sizeof (buf), f) != NULL)
     {
         const char *p;
         gchar **pos_tokens;
 
         // check if the filename matches the beginning of string
-        if (strncmp (buf, vfs_path_as_str (filename_vpath), len) != 0)
+        if (strncmp (buf, filename_str, len) != 0)
             continue;
 
         // followed by single space
@@ -1059,6 +1061,7 @@ load_file_position (const vfs_path_t *filename_vpath, long *line, long *column, 
         g_strfreev (pos_tokens);
     }
 
+    g_free (filename_str);
     fclose (f);
 }
 
@@ -1073,6 +1076,7 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
 {
     static size_t filepos_max_saved_entries = 0;
     char *fn, *tmp_fn;
+    char *filename_str = NULL;
     FILE *f, *tmp_f;
     char buf[MC_MAXPATHLEN + 100];
     size_t i;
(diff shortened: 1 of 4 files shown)


zyv:
@aborodin what do you think?

src/file_history.c
```diff
@@ -58,6 +58,21 @@ typedef struct file_history_data_t
 /*** file scope functions ************************************************************************/
 /* --------------------------------------------------------------------------------------------- */
 
+static void
+file_history_parse_entry (const char *buf, GList **file_list)
+{
+    const char *s = strrchr (buf, ' ');
+
+    file_history_data_t *fhd = g_new (file_history_data_t, 1);
```
mc-worker: Please move the variable declaration to the beginning of block.

src/file_history.c
```diff
@@ -58,6 +58,21 @@ typedef struct file_history_data_t
...
+    fhd->file_name = g_strndup (buf, s - buf);
+    const size_t len = strlen (s + 1);
```
mc-worker, line 76: It would be nice to add empty lines before and after constant.

tests/src/file_history.c
```diff
@@ -0,0 +1,61 @@
...
+#include "src/file_history.c"
+
+/* --------------------------------------------------------------------------------------------- */
+
+START_TEST (test_file_history_parse_entry)
+{
+    GList *file_list = NULL;
+
+    const char *mock_entries[] = { "/home/codespace/tmp/foo\n", "\n", "bar 2;0;5\n", "" };
+    const size_t entries_count = sizeof (mock_entries) / sizeof (mock_entries[0]);
```
mc-worker: G_N_ELEMENTS

tests/src/file_history.c
```diff
@@ -0,0 +1,61 @@
...
+
+    for (size_t i = 0; i < entries_count; i++)
+        file_history_parse_entry (mock_entries[i], &file_list);
+
+    ck_assert_uint_eq (g_list_length (file_list), 1);
+}
```
mc-worker, line 52: file_list isn't free.

src/file_history.c
```diff
@@ -129,10 +131,13 @@ file_history_list_write (const GList *file_list)
         {
             file_history_data_t *fhd = (file_history_data_t *) file_list->data;
 
-            g_string_append (s, fhd->file_name);
+            char *file_name = str_escape (fhd->file_name, -1, "", TRUE);
```
mc-worker: Please move the variable declaration to the beginning of block.

lib/util.c
```diff
@@ -1072,7 +1075,7 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
                     GArray *bookmarks)
 {
     static size_t filepos_max_saved_entries = 0;
-    char *fn, *tmp_fn;
+    char *fn, *tmp_fn, *filename_str = NULL;
```
mc-worker: "Avoid having initialized and uninitialized variables in the same declaration"

mc-worker (commented):
In general:

- the message in the first commit doesn't describe the problem.
- why the escaping is being done twice: in load_file_position() and in save_file_position(), but unescaping never?

mc-worker (approved):
Approved with some fixup.

PR #4767, opened 2025-09-27
Updated Italian translation

ciampix:


zyv:
Hi Marco, could you upload this to Transifex instead, please? Thank you!

ciampix:
oh sorry I forgot that mc uses Tx (and I even have used it some time ago...) & I saw the .po file was a lot behind and I just thought to update it directly. I do not see any way to update a .po file into Transifex, is there a way?

zyv:
> I do not see any way to update a .po file into Transifex, is there a way?

Yes, of course there is a way. Just make sure that you base your translation on the file downloaded from Transifex, and not from this repository. We are only syncing language files with Transifex upon releases, so the state you have here is always outdated.

Screenshot 2025-09-28 at 08 08 52

ciampix:
You are always very kind, thanks!

ciampix:
I didn't see that because it wasn't there... it seems that it's a function unavailable to free accounts... I will do it by hand.

ciampix:
Updating from Transifex...

zyv:
> I didn't see that because it wasn't there... it seems that it's a function unavailable to free accounts... I will do it by hand.

I think that maybe this is a function only available to the project administrators, but it must be definitively available for free accounts, because our account is a free one as a matter of fact.

I'm sorry for the manual work. I could have uploaded the file for you, but then I guess I would have overwritten what was (possibly) already translated there.

The problem is for us that accepting translations via git is unmanageable for the number of languages and different translators, so some platform like Transifex is really necessary...

Issue #4548, opened 2024-06-15
Mc start up is slow on FreeBSD with Bash shell when mc is using a subshell


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4548 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `tonythed1111` (tony@….com) |
> | **Mentions** | `maverick.woo@….com` |

<p>
Mc start up is slow on FreeBSD systems when running a Bash shell and the mc subshell feature is in use. There is an annoying delay of about 1 to 2 seconds during start up of the mc program. This occurs on any FreeBSD version that I have used when running bash as my login shell. The delay does not occur on any Linux system that I have used, it only happens on the FreeBSD systems. On FreeBSD when mc is started as 'mc -u' with no subshell the program starts almost instantly and is the expected behavior. When mc is started as 'mc' which does have a subshell, there is an annoying 1 to 2 second delay before it starts up.<br />
</p>
<p>
What steps will reproduce the problem? <br />
</p>
<ol><li>On any FreeBSD operating system set your login shell to use Bash (any bash version).
</li></ol><ol start="2"><li>On the console or xterm, run 'mc'.
</li></ol><ol start="3"><li>You will observe a 1 to 2 second annoying delay before the program starts up.
</li></ol><ol start="4"><li>Now run 'mc -u'.
</li></ol><ol start="5"><li>You will observe an almost instant start up, which is the expected behavior for mc.
</li></ol><ol start="6"><li>The inclusion of a subshell causes the annoying delay.
</li></ol>

zyv:
I’m going to close this with #4784, as related changes should have fixed this issue. If it still comes back after that, then please open a separate ticket with a reliable reproducer. Thank you!

Issue #3654, opened 2016-06-16
sftp file uploads/downloads hang or break on RPi, SSHDroid, Remarkable, etc.


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3654 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `x905` |

<p>
in 4.8.17 has been broken sftp connection command<br />
server is sshdroid on android device<br />
4.8.17 can connect, browse files and upload small files (up to 20 kb), but 30 kb cant - and hangs<br />
4.8.15 in same situation works fine (except error -31 as always)<br />
</p>
<p>
some add info <a class="ext-link" href="https://www.linux.org.ru/forum/midnight/12592086">https://www.linux.org.ru/forum/midnight/12592086</a><br />
</p>

zyv:
First, a general comment on `libssh2` / `libssh`.

I think that `libssh2` is still actively developed, but it has (or used to have?) issues with modern crypto, and my experience with the interface is not that great, although our usage of it is also quite wrong (see next comment). To my knowledge, for these reasons, Fedora switched to `libssh` for `curl`, and maybe we should too:

https://fedoraproject.org//wiki/Changes/libssh-in-libcurl

My quick appraisal results in the fact that `libssh` API is safer and much easier to use for our purposes:

https://api.libssh.org/master/libssh_tutor_guided_tour.html
https://api.libssh.org/master/libssh_tutor_sftp.html

Specifically, they seem to have a clean synchronous API for SFTP and simple error handling, which is what we need and the lack of which in `libssh2` has been causing most of the issues in the past.

On top of that, I think they have a better `ssh_config` parsing and `known_hosts` support, which would allow us to remove some of our code.

If anyone would like to take on the research, whether it's worth it and solves our (other) longstanding problems (like no proper `known_hosts` interface, etc.), and take up the implementation, especially this time hopefully covered with tests, they're most welcome.

zyv:
Now, to our problems with `libssh2`:

I've looked at it deeper today, and my conclusion is, unfortunately, that our use of it is still a hot mess, irrespective of the issues with the library. Especially, our error handling is sadly still pure madness after many attempts to put lipstick on a pig.

---

Our biggest issue is with the synchronous / asynchronous `libssh2` API usage. According to my understanding, there is a following philosophical difference between the libraries:

[...]

PR #4772, opened 2025-09-29
Fix macos zsh startup delay

### LINKED ISSUE #4625: mc 4.8.32 slow start on MacOS

wallneradam:
## Proposed changes

I've been debugging why the persistent command buffer feature doesn't work on macOS while it works fine on
Linux. Turns out it's a signal handling difference between BSD and Linux systems.

The problem: BSD systems don't send SIGCHLD when you SIGCONT a stopped process (Linux does). MC's persistent
 buffer test relies on getting that SIGCHLD after sending SIGCONT to the subshell.

The fix is pretty straightforward:
- Explicitly clear the SA_NOCLDSTOP flag when setting up the SIGCHLD handler. This tells the kernel we want
SIGCHLD on both stop AND continue.
- Added a small delay on BSD before running the test - gives the shell time to properly initialize
- Made sure MC's precmd hook is first in ZSH's precmd_functions array

The 100ms delay might seem counterintuitive, but it actually makes startup faster. Without it, the test
fails and retries with 1-second timeouts. With it, the test passes first try.

This also fixes that annoying 10-second startup delay some people have been seeing with ZSH on macOS.

Resolves: #4625, #4548

## Checklist

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -171,7 +171,16 @@ init_sigchld (void)
 
 #ifdef SA_RESTART
     sigchld_action.sa_flags = SA_RESTART;
+#else
+    sigchld_action.sa_flags = 0;
 #endif
+    /* BSD systems (macOS, FreeBSD, etc.) differ from Linux in SIGCHLD behavior.
+     * By default, BSD doesn't send SIGCHLD when a stopped child is continued with
+     * SIGCONT, while Linux does. This breaks the subshell's persistent command
+     * buffer test which relies on receiving SIGCHLD after sending SIGCONT.
+     * Explicitly clearing SA_NOCLDSTOP ensures we get SIGCHLD on both stop AND
+     * continue, making behavior consistent across platforms. */
+    sigchld_action.sa_flags &= ~SA_NOCLDSTOP;
 
     if (my_sigaction (SIGCHLD, &sigchld_action, NULL) == -1)
     {
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -1267,7 +1267,7 @@ init_subshell_precmd (void)
             " mc_print_cursor_position () { echo $CURSOR >&%d}\n"
             " zle -N mc_print_cursor_position\n"
             " bindkey '^[" SHELL_CURSOR_KEYBINDING "' mc_print_cursor_position\n"
-            " _mc_precmd(){ pwd>&%d;kill -STOP $$ }; precmd_functions+=(_mc_precmd)\n"
+            " _mc_precmd(){ pwd>&%d;kill -STOP $$ }; precmd_functions=(_mc_precmd $precmd_functions)\n"
             "PS1='%%n@%%m:%%~%%# '\n",
             command_buffer_pipe[WRITE], command_buffer_pipe[WRITE], subshell_pipe[WRITE]);
 
@@ -1676,15 +1676,19 @@ init_subshell (void)
      * assume there must be something wrong with the shell, and we turn persistent buffer off
      * for good. This will save the user the trouble of having to wait for the persistent
      * buffer function to time out every time they try to close the subshell. */
-#ifdef __APPLE__
-    /* On macOS with ZSH, skip the test because the shell is in SIGSTOP state and cannot respond.
-     * The feature works but the test times out. */
-    if (use_persistent_buffer && mc_global.shell->type != SHELL_ZSH)
-#else
     if (use_persistent_buffer)
-#endif
     {
-        if (!read_command_line_buffer (TRUE))
+        gboolean test_passed = FALSE;
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
+        /* On BSD systems, we need a small delay to ensure proper initialization.
+         * This allows the shell's precmd_functions array to be properly set up
+         * with MC's hook in the first position, and gives time for SIGCHLD
+         * handling to stabilize after SIGCONT. Paradoxically, this delay
+         * actually makes startup faster by avoiding failed test retries. */
+        g_usleep (100000);  /* 100ms */
+#endif
+        test_passed = read_command_line_buffer (TRUE);
+        if (!test_passed)
             use_persistent_buffer = FALSE;
     }
 }
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -1676,8 +1676,17 @@ init_subshell (void)
      * assume there must be something wrong with the shell, and we turn persistent buffer off
      * for good. This will save the user the trouble of having to wait for the persistent
      * buffer function to time out every time they try to close the subshell. */
-    if (use_persistent_buffer && !read_command_line_buffer (TRUE))
-        use_persistent_buffer = FALSE;
+#ifdef __APPLE__
+    /* On macOS with ZSH, skip the test because the shell is in SIGSTOP state and cannot respond.
+     * The feature works but the test times out. */
+    if (use_persistent_buffer && mc_global.shell->type != SHELL_ZSH)
+#else
+    if (use_persistent_buffer)
+#endif
+    {
+        if (!read_command_line_buffer (TRUE))
+            use_persistent_buffer = FALSE;
+    }
 }
 
 /* --------------------------------------------------------------------------------------------- */


src/main.c
```diff
@@ -171,7 +171,16 @@ init_sigchld (void)
...
+#else
+    sigchld_action.sa_flags = 0;
 #endif
+    /* BSD systems (macOS, FreeBSD, etc.) differ from Linux in SIGCHLD behavior.
+     * By default, BSD doesn't send SIGCHLD when a stopped child is continued with
+     * SIGCONT, while Linux does. This breaks the subshell's persistent command
+     * buffer test which relies on receiving SIGCHLD after sending SIGCONT.
+     * Explicitly clearing SA_NOCLDSTOP ensures we get SIGCHLD on both stop AND
+     * continue, making behavior consistent across platforms. */
+    sigchld_action.sa_flags &= ~SA_NOCLDSTOP;
```
zyv, line 183: I like this very much indeed! It's good for us to have consistent behaviour across the systems.

src/subshell/common.c
```diff
@@ -1267,7 +1267,7 @@ init_subshell_precmd (void)
             " mc_print_cursor_position () { echo $CURSOR >&%d}\n"
             " zle -N mc_print_cursor_position\n"
             " bindkey '^[" SHELL_CURSOR_KEYBINDING "' mc_print_cursor_position\n"
-            " _mc_precmd(){ pwd>&%d;kill -STOP $$ }; precmd_functions+=(_mc_precmd)\n"
+            " _mc_precmd(){ pwd>&%d;kill -STOP $$ }; precmd_functions=(_mc_precmd $precmd_functions)\n"
```
zyv, line 1270: I'm not quite sure if this is the best or safest way to interact with the other hooks, because they will pause until we have finished, and then run — possibly quite a while later and under different circumstances. I believe this is why our hook was added at the end...

If this isn't an issue, should we also put our hook at the beginning for the other supported shells, like `bash` and `fish`?

src/subshell/common.c
```diff
@@ -1676,8 +1676,21 @@ init_subshell (void)
      * assume there must be something wrong with the shell, and we turn persistent buffer off
      * for good. This will save the user the trouble of having to wait for the persistent
      * buffer function to time out every time they try to close the subshell. */
-    if (use_persistent_buffer && !read_command_line_buffer (TRUE))
-        use_persistent_buffer = FALSE;
+    if (use_persistent_buffer)
+    {
+        gboolean test_passed = FALSE;
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
```
zyv, line 1682: Oh, I still find this quite problematic. :( Everything timing-related is a recipe for disaster.

Also, from a technical viewpoint, listing all the BSD systems is annoying. What about Dragonfly, etc.? Is there a generic BSD macro out there?

zyv (commented):
Just to make sure I understand correctly, does the signal stuff fix the problem and do the other two changes speed up initialization?

If so, what is the timing difference? I'm tempted to only take the signal one for now.

src/subshell/common.c
wallneradam: Without this it was very unstable. With this everything is working for me.

src/subshell/common.c
wallneradam: I agree with you with the timing-related part. However it is working without the timer too. It just make it faster and stable. I think it is acceptable here, because it does not solves the actual problem, but makes it better.
I'm not very familiar with these macros. I think this timer could be good on linux too. Just I've tested only on macOS.

wallneradam:
> Just to make sure I understand correctly, does the signal stuff fix the problem and do the other two changes speed up initialization?

~~For me the signal stuff was not enough. Also changing the hook order was necessary.~~
Without the timeout it was random, sometimes immediate, sometimes more seconds, and sometimes even 10s and buffer was disabled. With the timeout it is reliable, for me it is about 0.5 sec start.

src/subshell/common.c
wallneradam: Oh, sorry! You are right. This is not needed. I remembered wrong. Only the timeout is needed.

src/subshell/common.c
wallneradam: Ok, I've tested again. And timer is necessary. It works even with 100us. But it needs time. I think 1ms is safe here, 10ms is even safer. But maybe there is a better way I don't know.

egmontkob: Do we still need to check for the OS, after the SA_NOCLDSTOP bit?

Could it be replaced by a configure macro checking the behavior?

> [zyv] Everything timing-related is a recipe for disaster.

Absolutely. Anything along the lines of "let's have a tiny delay of a particular amount which works nicely on my system" is super fragile, and is a sign of the overall architecture still being faulty.

mc-worker: > Also, from a technical viewpoint, listing all the BSD systems is annoying. What about Dragonfly, etc.? Is there a generic BSD macro out there?

https://forums.freebsd.org/threads/gcc-general-macro-for-bsd.47769/

The Boost methods: https://github.com/boostorg/config/tree/bf8c8bd7960cd8505210838f86bd28fb4fb71071/include/boost/config/platform

We can define some macro in `lib/unixcompat.h` or, alternatively, in `config.h` via `configure.ac`.

src/subshell/common.c
zyv: Okay, could you please revert that? Thank you!

src/subshell/common.c
zyv: > Do we still need to check for the OS, after the SA_NOCLDSTOP bit?

I don't think so. This would only introduce extra complexity with no apparent benefit. I'd just edit the comment to say that we clear it unconditionally because it's the default on other systems anyway.

zyv: > Ok, I've tested again. And timer is necessary. It works even with 100us. But it needs time. I think 1ms is safe here, 10ms is even safer. But maybe there is a better way I don't know.

In principle, I would prefer a consistent and unconditional 10 ms delay on all systems instead of dealing with macros and multiple code paths, if there is no way around it.

Is there any chance you could conclude this investigation by determining why the delay is necessary? If it's unavoidable without completely refactoring our code, then we could perhaps take it, but document in the comment why it is needed and how to do it better...

wallneradam: I'm going to try to investigate it more, why it is necessary or is there a better, more deterministic way. Usually I don't like timers either.

egmontkob: > In principle, I would prefer a consistent and unconditional 10 ms delay on all systems

What happened to you zyv? :)

How come a 10 ms delay could make an unreliable code run reliably, or speed up things? Across all systems, not just on 1 particular computer that someone is testing right now? I have zero trust towards this.

Other than a giant safety timeout when we no longer wait for the optional response to arrive, give up and continue with degraded functionality, the code MUST NOT have any timers, timeouts, delays.

zyv: > > In principle, I would prefer a consistent and unconditional 10 ms delay on all systems
> 
> What happened to you zyv? :)

I guess I'm getting older and overwhelmed by the lack of time and scope of the project... Check the latest comments in #3654 :(

In view of this, I would be willing to accept a terrible solution as long as we know exactly what and why it is doing, to solve the problems of the users right now and not in 10 years, and if it doesn't introduce more breakage. Our codebase is just... in a very bad shape anyways.

src/main.c
egmontkob, line 183: How can clearing this bit make any difference, given that the field is `memset()` to fully zeros above?

The only way `sa_flags` can have this bit enabled now is if `SA_RESTART` isn't a single bit but multiple bits including `SA_NOCLDSTOP`. https://man.openbsd.org/sigaction.2 talks about "setting the SA_RESTART bit in sa_flags" which suggests to me that this isn't the case.

Also, your comment says "By default, BSD doesn't send SIGCHLD". What does "default" refer to, given that the API requires the _exact_ set of flags to be passed, and both systems define `SA_NOCLDSTOP` but neither defines the opposite `SA_CLDSTOP`, what do you mean by the default being different in the two systems?

What am I missing here?

src/subshell/common.c
egmontkob: Do we know exactly why a 10ms delay improves things? Not just on one computer, for one user, with one config, under one particular system load that he might be having right now? I mean, it could even simply break under a bit higher load, couldn't it?

> Our codebase is just... in a very bad shape anyways.

Unfortunately, I know. But allowing further bad code to enter is not the way to get out of this situation.

src/main.c
ossilator: yeah, this makes no sense.
"by default" refers to the system's actual default behavior, and also what signal() does. sigaction() otoh has defined behavior for this.

src/subshell/common.c
ossilator: it's a question of just _how_ bad it is. i agree that a well-documented and generously timed hack is better than no solution at all, as ugly as i find it. however, i'll insist on understanding the issue so thoroughly that half the work toward the actual solution will be already done, thus making going with the hack somewhat pointless - unless the proper solution would require a really fundamental change, like making the whole thing async (possibly via using the glib main loop).

ossilator (changes requested):

src/main.c
egmontkob: Ossi yes your explanation on how Adam probably meant "default" makes sense. But then, the change still doesn't.

(Also adding the `#else` branch is pointless since the value is memset to 0 above, although it might improve readability. But if readability is a goal, and we potentially modify the flags in multiple different ways, then IMHO the cleaner approach would be to say `sa_flags |= SA_RESTART` and no `#else` branch; this is the syntax that scales to having multiple flag operations in a consistent manner and in any order.)

[This](https://cpip.readthedocs.io/en/stable/_static/dictobject.c/signal.h_bbe000f9714f274340a28e000a369354.html) is a random web find, not sure how up-to-date it is, but according to this SA_NOCLDSTOP is indeed _not_ included in SA_RESTART (and why would it? they are unrelated concepts), so the `sa_flags &= ~SA_NOCLDSTOP` is a no-op.

wallneradam: Yes you are right. I've just found that it is a different in Linux and macOS. And after changing this. It was somehow good. But that was a random thing. Only the delay is needed. But now I implemented a much better solution which is deterministic and works in both macOS and Linux.

egmontkob: > And after changing this. It was somehow good.

So you made a no-op change and somehow the bug went away. Clear proof that there's some randomness in the game. And I'm afraid we still don't understand what is happening.

Do we understand what's going on?

SA_NOCLDSTOP was a red herring. The default behavior _might_ be different on macOS vs. Linux, I haven't verified that claim, but it's irrelevant because mc doesn't rely on the default behavior, it explicitly specifies which of the two possible behaviors it wants.

Do we know what's the first place when things go differently for macOS vs. Linux?

Do we know the sequence of steps that leads to the delay on macOS? (Your previous explanation on its own can't be correct because we now understant that SIGCHLD on SIGCONT _should_ be sent on macOS, too.)

> But now I implemented a much better solution which is deterministic

Is that available somewhere? Or do you mean the current PR with its 100ms delay? Anything with a delay involved might be a heuristic workaround only which works sometimes (maybe quite often) but not always.

If we don't understand what and why is happening, but empirically find that a randomly inserted tiny delay seems to fix it (on one computer, for one user, at one particular system load, maybe about 10 out of 10 times so far -- but how about across all our user base, executed thousands of times?) then it's sorcery, not software development.

zyv:
Superseded by #4934.

PR #4785, opened 2025-10-13
php.syntax: add missed magic constants

KuzinAndrey:
## Proposed changes

Add missed magic constants in PHP syntax flle
https://www.php.net/manual/en/language.constants.magic.php


zyv (approved):
Thank you!

PR #4786, opened 2025-10-13
php.syntax: grep real list of core reserved constants from documentation

KuzinAndrey:
Regexp with prefixes is not convenient for syntax check by coloring.
Get real list from documentation and use it for reserved constants by helper command:
```sh
wget -q -O - https://www.php.net/manual/en/reserved.constants.php | \
grep -Eo "<strong><code><a href=.*>(PHP|ZEND|DEBUG|PEAR|DEFAULT|E)_.*</a></code></strong>" | \
grep -Eo '(PHP_|ZEND_|DEBUG_|PEAR_|DEFAULT_|E_)[_A-Z]*' | \
sort -u | sed 's/\(.*\)/    keyword whole \1 white/g'
```
Add STD(IN|OUT|ERR) by hands.

### DIFF
--- a/misc/syntax/php.syntax
+++ b/misc/syntax/php.syntax
@@ -3056,11 +3056,76 @@ context default
 # predefined constants
 # core
 
-    keyword whole PHP\_\[0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_\] white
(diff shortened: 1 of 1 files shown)


zyv (approved):
Thanks!

PR #4802, opened 2025-10-19
Ticket #4801: invalid utf-8 in prompt

### LINKED ISSUE #4801: Prompt resyncronizes too late after invalid UTF-8

```shell
master
```

### Operating system

```shell
all
```

### How to reproduce

UTF-8 locale, UTF-8 terminal.

Set up prompt in `.bashrc` to print the directory (if you don't have it already), e.g.

```
PS1='\w$ '
```

Create a directory containing an invalid UTF-8 in its name, such as the string `abcdéfghi` encoded in latin1:

```
mkdir $'abcd\351fghi'
```

Enter this directory.

Outside of mc, the directory in the prompt should appear as `abcd�fghi` or so.

Inside mc, in its command line prompt, it's displayed as `abcdhi`: no replacement symbol, and the next two letters `fg` are swallowed.

---

To spice it up, let's use some color changing escape sequences, and put the invalid UTF-8 at the end of the directory name:

```
PS1='\[\e[31m\]\u@\w\[\e[0m\]$ '
```

```
mkdir $'abcd\351'  # abcdé encoded in latin1
```

The prompt in mc appears as `abcd0m$ `. The two stripped bytes are now `ESC` `[`, resulting in the rest of the escape sequence getting interpreted verbatim.

### Expected behavior

Synchronize immediately after an invalid UTF-8, preserve and interpret every valid UTF-8 character (including the escape byte starting the escape sequence).

### Actual behavior

Characters get lost after invalid UTF-8.

### Additional context

_No response_

egmontkob:
## Proposed changes

Don't assume that the prompt is valid UTF-8

* Resolves: #4801

## Checklist

### DIFF
--- a/tests/lib/terminal.c
+++ b/tests/lib/terminal.c
@@ -38,7 +38,7 @@
 static void
 setup (void)
 {
-    str_init_strings (NULL);
+    str_init_strings ("UTF-8");
 }
 
 static void
@@ -81,6 +81,26 @@ START_TEST (test_strip_ctrl_codes)
 }
 END_TEST
 
+// Test the handling of inner and final incomplete UTF-8, also make sure there's no overrun.
+// Ticket #4801. Ideally replacement symbols should be added, but we don't have that yet.
+START_TEST (test_strip_ctrl_codes2)
+{
+    // U+2764 heart in UTF-8, followed by " ábcdéfghíjklnmó\000pqrst" in Latin-1
+    const char s_orig[] = "\342\235\244 \341bcd\351fgh\355jklm\363\000pqrst";
+    ck_assert_int_eq (sizeof (s_orig), 25);
+
+    // copy the entire string, with embedded '\0'
+    char *s = g_malloc (sizeof (s_orig));
+    memcpy (s, s_orig, sizeof (s_orig));
+
+    char *actual = strip_ctrl_codes (s);
+    const char *expected = "\342\235\244 bcdfghjklm";
+
+    ck_assert_str_eq (actual, expected);
+    g_free (s);
+}
+END_TEST
+
 /* --------------------------------------------------------------------------------------------- */
 
 int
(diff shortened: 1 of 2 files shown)


ossilator (commented):
i don't know the entire context, so i can't possibly approve this, but from what i do see it looks sane.

same applies to the followup patch, which i'd squash here.

the autotest doesn't cover combining ctrl codes with botched utf8, which seems like a relevant omission. as you noted, the exact semantics probably aren't all too important, but the boundary conditions should be exercised for robustness.

note that afaict, you're fixing a security hole here: root may be browsing a user's folders, which are prepared to exploit the heap overflow. say hello to a CVE process ...

egmontkob:
Please review.

I decided to go with two separate commits, without squashing them. If for some reason the second commit turns out to be undesired, we can easily revert but we'll still keep the important bugfix of the first.

tests/lib/terminal.c
```diff
@@ -81,6 +81,25 @@ START_TEST (test_strip_ctrl_codes)
 }
 END_TEST
 
+// Test the handling of inner and final incomplete UTF-8, also make sure there's no overrun.
+// Ticket #4801. Ideally replacement symbols should be added, but we don't have that yet.
```
zyv: This comment confuses me in view of the statement in #4801.

> The UI drawing part of mc already knows how to handle it (as it does e.g. in filenames), U+FFFD � replacement symbols appear.

This should probably be amended.

zyv (commented):
Otherwise, I think this should be merged. @aborodin, agree?

tests/lib/terminal.c
egmontkob: Yup I forgot to update that in the second commit, fixed now.

(I know it's a bit weird that I first fix the bug somehow, then the second change fixes it in another way. I did it so because the second fix is not super ideal in my opinion, it's technically a bit ugly to keep the invalid UTF-8 and rely on a later component replacing with replacement symbols. But just in case we'd decide to revert that, I'd still like to see the first fix remain in place, which fixes a buffer overrun without doing anything more complex. Hope that this decision makes sense.)

mc-worker (approved):

tests/lib/terminal.c
zyv: Thank you very much!

Issue #2109, opened 2010-03-17
Vertical lines on the mc panels do not properly stick into the bottom line


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/2109 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `zaytsev` (@zyv) |

<p>
A picture is worth thousand words. If someone does not like how it will look like, it can be made re-definable in the skin file.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [Screenshot.png](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/2109/Screenshot.png) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/2109/Screenshot.png)) by `zaytsev` (@zyv) on <span title="2010-03-17T08:55:09+00:00">`Mar 17, 2010 at 8:55 UTC`</span>

egmontkob:
First draft:  [mc-2109-full-vertical-lines.patch](https://github.com/user-attachments/files/23680916/mc-2109-full-vertical-lines.patch)

There's a bit of code duplication; the first (inner) occurrence is used for Listing format -> Full, the second (outer) is for Brief.

The way I do those gotoyx's is quite ugly I guess (mixing absolute tty coordinates vs. widget-relative coordinates), is there a nicer way?

It wasn't obvious to me how printing the header line (`panel_print_header()`) vs. actual file lines (`format_file()`) keep the same columns. Ideally it would be a third method also keeping the same count, if it's not too much code and business logic duplication. But instead I piggybacked on the header (that one can't be hidden, right?). Not a nice architecture, but okay for a first draft.

And for this reason, I had to move `panel_print_header()` to the middle of `show_dir()`, otherwise these new T shapes would override the directory in the top frame of the panel, and the free disk space info in the bottom of the panel if `show_mini_info=false`. This can be reverted if we come up with a standalone method painting only these T shapes, based on counting the columns.

egmontkob:
The patch breaks the colors of the special symbols in the top frame (show/hide dotfiles, open panel directory history etc.).

egmontkob:
Would you want the same kind of full vertical lines for the "User mini status", too?

Of course taking care that if the two vertical lines happen to align then we need to paint a cross.

[...]

PR #4813, opened 2025-10-24
Ticket #3158: fix ncurses double lines

### LINKED ISSUE #3158: No double lines with ncurses
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3158 |

<p>
Choose a skin with double lines, e.g. "mc -S double-lines". These are properly displayed with the slang build, but are replaced by brighter single lines with ncurses.<br />
</p>
<p>
If the enrivonment suggests that double lines are most likely supported (TERM refers to a graphical emulator and locale is utf-8) then the characters defined in the skin file should be used without being altered in any way.<br />
</p>

### LINKED ISSUE #3820: PuTTy requires NCURSES_NO_UTF8_ACS=1 for ncurses to actually use line-drawing characters
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3820 |
> | **Reporter** | `knudch` (knud.skrald@….info) |
> | **Mentions** | `egmont` (@egmontkob), `kloczek` (kloczko.tomasz@….com), `dickey` (dickey@….com) |

<p>
In a SSH session coming from PuTTy codepage is incorrect eg. line drawings etc is wrong.<br />
Should be UTF-8 but is ISO8859-1.<br />
SSH session from a Linux client MC 4.8.18 is working correct.<br />
</p>
<p>
Version 4.8.17 does not show this bug<br />
4.8.18/19 do..<br />
</p>
<p>
Knud<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [mc_bug1.png](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/3820/mc_bug1.png) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/3820/mc_bug1.png)) by `knudch` (knud.skrald@….info) on <span title="2017-05-02T14:59:17+00:00">`May 2, 2017 at 14:59 UTC`</span>

### LINKED ISSUE #4301: help page broken with "-a" on a restricted shell
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4301 |
> | **Reporter** | `alexander.stohr` (alexander.stohr@….de) |

<p>
use a shell that does not honor extended escape sequences (found this when using <tt>TeraTerm</tt> and a SSH connection to some ARM64 embedded board).<br />
mc version on the target system: 4.8.19<br />
</p>
<p>
as the border frames are "corrupted" i needed to use the command "mc -a" to launch the software and getting a useable screen.<br />
</p>
<p>
when calling help screen i found the big MC logo there to not displaying correctly. also the right border of the help text window is corrupted.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [mc-help-corruption.png](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/4301/mc-help-corruption.png) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/4301/mc-help-corruption.png)) by `alexander.stohr` (alexander.stohr@….de) on <span title="2021-11-12T13:21:01+00:00">`Nov 12, 2021 at 13:21 UTC`</span>

### LINKED ISSUE #4799: Fixup double-line frames and other special symbols in non-UTF-8 locales

We only have 3 skins that are usable with non-UTF-8 locales (i.e. non-UTF-8 terminal charset). These are `default`, `gotar` and `mc46`.

`double-lines` cannot display those double lines (they're unavailable in legacy encodings), it displays uppercase letter there.

All the other skins break at those characters defined in the `[widget-panel]` and `[widget-scrollbar]` sections; plus the ones that use double lines break there too.

Instead of showing broken glyphs, if a non-UTF-8 locale is detected we could replace each such character that's not either 7-bit ASCII or a single line drawing character (or maybe: not convertible to the current locale) with a hardcoded default. That is, double lines would fall back to single lines, fancy Unicode ▸ would fall back to >, etc.

This would be a softer and automatic variant of `mc -a`, which is a bit too agressive, even replaces single line drawing chars with ASCII variants, even though most terminals support these glyphs in a special mode.

.

Maybe a slightly simpler variant is if the skin file defines a `requires_utf8 = true` and then we automatically replace everything with the hardcoded value.

Or just automatically ignore every such value in every skin file and automatically use the hardwired variant if the locale is not UTF-8; after all, they're the same in all non-UTF8-supporting skin files and the hardcoded values would be the same, too.

I'd love to believe that terminals running in non-UTF-8 are extremely rare, so the lowest hanging fruit to fix their broken layout is probably good enough.

### Additional context

_No response_

### LINKED ISSUE #4860: ncurses: Retire double lines -> bold conversion

```shell
current
```

### Operating system

```shell
any
```

### How to reproduce

Our ncurses code has a bit of special treatment for double lines if they are unavailable (e.g. because running with `LC_ALL=en_US`, an ISO-8859-1 locale).

They are converted to single lines plus `A_BOLD` (the SGR 1 escape sequence).

This in turn, depending on the terminal, may result in bold typeface (thicker frame), brighter color, both, or neither.

slang build doesn't have it, but there double lines in 8-bit locales currently result in garbage (`P`, `Q` etc.) anyway.

My pending PR #4813 keeps this boldening behavior for ncurses, and even though fixes slang to draw single lines, doesn't add this feature there.

Therefore, with my PR, the two builds still remain inconsistent.

---

You need truly special circumstances to see this boldening behavior:
- a legacy 8-bit locale, very few users should have nowadays,
- an ncurses build (which I have the _feeling_ is the rarer, distros tend to link it to slang???),
- a non-default skin with double lines,
- a terminal that at least either boldens the lines (many terminals manually paint these line drawing characters and don't make them thicker) or brightens the colors (what quite a few terminals no longer do, also you'd need a frame color not already fully white).

That is, I suspect that extremely few users see this beavhior. Even for them, as the default skin perfectly demonstrates, this is an absolutely nonessential feature, a non-boldened/brightened frame is just fine.

---

The result looks somewhat ugly at "tee" characters, e.g. `╟`. That's because even the little stem (to the right in this example) is subject to boldening/brightening, but the continuation of that line isn't. So the very beginning and end of such long horizontal lines is slightly bolder and/or brighter.

---

Let's remove this feature.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
## Proposed changes

Refactor frame handling, fixing double lines in ncurses build, fixing the look in 8-bit locales.

To be thorougly tested with ncurses, slang, old ncurses, non-wide ncurses etc.; combined with utf-8, koi8-{r,u}, latin-x etc. locales; combined with various skins.

* Resolves: #3158
* Resolves: #3820
* Resolves: #4301 
* Resolves: #4799
* Resolves: #4860

## Checklist

### DIFF
--- a/lib/mcconfig/get.c
+++ b/lib/mcconfig/get.c
@@ -120,6 +120,48 @@ mc_config_get_string (mc_config_t *mc_config, const gchar *group, const gchar *p
 
 /* --------------------------------------------------------------------------------------------- */
 
+// like mc_config_get_string, but fail (use default) if can't be fully converted to the target
+// charset
+gchar *
+mc_config_get_string_strict (mc_config_t *mc_config, const gchar *group, const gchar *param,
+                             const gchar *def)
+{
+    GIConv conv;
+    GString *buffer;
+    gchar *ret;
+    estr_t conv_res;
+
+    ret = mc_config_get_string_raw (mc_config, group, param, def);
+
+    if (!g_utf8_validate (ret, -1, NULL))
+    {
+        g_free (ret);
+        return g_strdup (def);
+    }
+
+    if (mc_global.utf8_display)
+        return ret;
+
+    conv = str_crt_conv_from ("UTF-8");
+    if (conv == INVALID_CONV)
+        return ret;
+
+    buffer = g_string_new ("");
+    conv_res = str_convert (conv, ret, buffer);
+    g_free (ret);
+    str_close_conv (conv);
+
+    if (conv_res != ESTR_SUCCESS)
+    {
+        g_string_free (buffer, TRUE);
+        return g_strdup (def);
+    }
+
+    return g_string_free (buffer, FALSE);
+}
+
+/* --------------------------------------------------------------------------------------------- */
+
 gchar *
 mc_config_get_string_raw (mc_config_t *mc_config, const gchar *group, const gchar *param,
                           const gchar *def)
--- a/lib/skin/lines.c
+++ b/lib/skin/lines.c
@@ -29,6 +29,7 @@
 
 #include "internal.h"
 #include "lib/tty/tty.h"
+#include "lib/strutil.h"
 
 /*** global variables ****************************************************************************/
 
@@ -44,17 +45,104 @@
 /*** file scope functions ************************************************************************/
 /* --------------------------------------------------------------------------------------------- */
 
-static int
-mc_skin_lines_load_frm (mc_skin_t *mc_skin, const char *name)
+static mc_tty_char_t
+unicode_to_mc_acs (gunichar c)
 {
-    int ret;
-    char *frm_val;
+    switch (c)
+    {
+    case 0x2500:  // ─
+        return MC_ACS_HLINE;
+    case 0x2502:  // │
+        return MC_ACS_VLINE;
+    case 0x250C:  // ┌
+        return MC_ACS_ULCORNER;
+    case 0x2510:  // ┐
+        return MC_ACS_URCORNER;
+    case 0x2514:  // └
+        return MC_ACS_LLCORNER;
+    case 0x2518:  // ┘
+        return MC_ACS_LRCORNER;
+    case 0x251C:  // ├
+        return MC_ACS_LTEE;
+    case 0x2524:  // ┤
+        return MC_ACS_RTEE;
+    case 0x252C:  // ┬
+        return MC_ACS_TTEE;
+    case 0x2534:  // ┴
+        return MC_ACS_BTEE;
+    case 0x253C:  // ┼
+        return MC_ACS_PLUS;
 
-    frm_val = mc_config_get_string_raw (mc_skin->config, "Lines", name, " ");
-    ret = mc_tty_normalize_lines_char (frm_val);
-    g_free (frm_val);
+    default:
+        return 0;
+    }
+}
 
-    return ret;
+/* --------------------------------------------------------------------------------------------- */
+
+// "def" is the default in case the skin doesn't define anything. It's not the fallback in case the
+// character cannot be converted to the locale, the fallback is handled by the caller.
+static gboolean
+skin_get_char (mc_skin_t *mc_skin, const char *name, gunichar def, mc_tty_char_t *result)
+{
+    gunichar c;
+    GIConv conv;
+    GString *buffer;
+    estr_t conv_res;
+    char *value_utf8;
+
+    value_utf8 = mc_config_get_string_raw (mc_skin->config, "Lines", name, NULL);
+    if (value_utf8 != NULL)
+    {
+        c = g_utf8_get_char_validated (value_utf8, -1);
+        if (c == (gunichar) (-1) || c == (gunichar) (-2))
+        {
+            g_free (value_utf8);
+            return FALSE;
+        }
+    }
+    else
+    {
+        // the default to be used if not defined in the skin file
+        value_utf8 = g_malloc (7);
+        const int len = g_unichar_to_utf8 (def, value_utf8);
+        value_utf8[len] = '\0';
+        c = def;
+    }
+
+    // in UTF-8 nothing left to do, no ACS business, just return the codepoint
+    if (mc_global.utf8_display)
+    {
+        g_free (value_utf8);
+        *result = c;
+        return TRUE;
+    }
+
+    // use MC_ACS_* values for single line drawing chars
+    const mc_tty_char_t mc_acs = unicode_to_mc_acs (c);
+    if (mc_acs != 0)
+    {
+        g_free (value_utf8);
+        *result = mc_acs;
+        return TRUE;
+    }
+
+    // convert to 8-bit charset, e.g. KOI8-R has all the double line chars
+    conv = str_crt_conv_from ("UTF-8");
+    if (conv == INVALID_CONV)
+    {
+        g_free (value_utf8);
+        return FALSE;
+    }
+
+    buffer = g_string_new ("");
+    conv_res = str_convert (conv, value_utf8, buffer);
+    if (conv_res == ESTR_SUCCESS)
+        *result = (unsigned char) buffer->str[0];
+    str_close_conv (conv);
+    g_string_free (buffer, TRUE);
+    g_free (value_utf8);
+    return conv_res == ESTR_SUCCESS;
 }
 
 /* --------------------------------------------------------------------------------------------- */
@@ -64,35 +152,70 @@ mc_skin_lines_load_frm (mc_skin_t *mc_skin, const char *name)
 void
 mc_skin_lines_parse_ini_file (mc_skin_t *mc_skin)
 {
+    gboolean success;
+
     if (mc_global.tty.slow_terminal)
         mc_skin_hardcoded_space_lines (mc_skin);
     else if (mc_global.tty.ugly_line_drawing)
         mc_skin_hardcoded_ugly_lines (mc_skin);
 
     // single lines
-    mc_tty_frm[MC_TTY_FRM_VERT] = mc_skin_lines_load_frm (mc_skin, "vert");
-    mc_tty_frm[MC_TTY_FRM_HORIZ] = mc_skin_lines_load_frm (mc_skin, "horiz");
-    mc_tty_frm[MC_TTY_FRM_LEFTTOP] = mc_skin_lines_load_frm (mc_skin, "lefttop");
-    mc_tty_frm[MC_TTY_FRM_RIGHTTOP] = mc_skin_lines_load_frm (mc_skin, "righttop");
-    mc_tty_frm[MC_TTY_FRM_LEFTBOTTOM] = mc_skin_lines_load_frm (mc_skin, "leftbottom");
-    mc_tty_frm[MC_TTY_FRM_RIGHTBOTTOM] = mc_skin_lines_load_frm (mc_skin, "rightbottom");
-    mc_tty_frm[MC_TTY_FRM_TOPMIDDLE] = mc_skin_lines_load_frm (mc_skin, "topmiddle");
-    mc_tty_frm[MC_TTY_FRM_BOTTOMMIDDLE] = mc_skin_lines_load_frm (mc_skin, "bottommiddle");
-    mc_tty_frm[MC_TTY_FRM_LEFTMIDDLE] = mc_skin_lines_load_frm (mc_skin, "leftmiddle");
-    mc_tty_frm[MC_TTY_FRM_RIGHTMIDDLE] = mc_skin_lines_load_frm (mc_skin, "rightmiddle");
-    mc_tty_frm[MC_TTY_FRM_CROSS] = mc_skin_lines_load_frm (mc_skin, "cross");
+    success = TRUE /* just for nicer indentation */
+        && skin_get_char (mc_skin, "horiz", 0x2500, &mc_tty_frm[MC_TTY_FRM_HORIZ])
+        && skin_get_char (mc_skin, "vert", 0x2502, &mc_tty_frm[MC_TTY_FRM_VERT])
+        && skin_get_char (mc_skin, "lefttop", 0x250C, &mc_tty_frm[MC_TTY_FRM_LEFTTOP])
+        && skin_get_char (mc_skin, "righttop", 0x2510, &mc_tty_frm[MC_TTY_FRM_RIGHTTOP])
+        && skin_get_char (mc_skin, "leftbottom", 0x2514, &mc_tty_frm[MC_TTY_FRM_LEFTBOTTOM])
+        && skin_get_char (mc_skin, "rightbottom", 0x2518, &mc_tty_frm[MC_TTY_FRM_RIGHTBOTTOM])
+        && skin_get_char (mc_skin, "topmiddle", 0x252C, &mc_tty_frm[MC_TTY_FRM_TOPMIDDLE])
+        && skin_get_char (mc_skin, "bottommiddle", 0x2534, &mc_tty_frm[MC_TTY_FRM_BOTTOMMIDDLE])
+        && skin_get_char (mc_skin, "leftmiddle", 0x251C, &mc_tty_frm[MC_TTY_FRM_LEFTMIDDLE])
+        && skin_get_char (mc_skin, "rightmiddle", 0x2524, &mc_tty_frm[MC_TTY_FRM_RIGHTMIDDLE])
+        && skin_get_char (mc_skin, "cross", 0x253C, &mc_tty_frm[MC_TTY_FRM_CROSS]);
+
+    // if any of them are unavailable, revert all of them to regular single lines
+    if (!success)
+    {
+        mc_tty_frm[MC_TTY_FRM_HORIZ] = MC_ACS_HLINE;
+        mc_tty_frm[MC_TTY_FRM_VERT] = MC_ACS_VLINE;
+        mc_tty_frm[MC_TTY_FRM_LEFTTOP] = MC_ACS_ULCORNER;
+        mc_tty_frm[MC_TTY_FRM_RIGHTTOP] = MC_ACS_URCORNER;
+        mc_tty_frm[MC_TTY_FRM_LEFTBOTTOM] = MC_ACS_LLCORNER;
+        mc_tty_frm[MC_TTY_FRM_RIGHTBOTTOM] = MC_ACS_LRCORNER;
+        mc_tty_frm[MC_TTY_FRM_TOPMIDDLE] = MC_ACS_TTEE;
+        mc_tty_frm[MC_TTY_FRM_BOTTOMMIDDLE] = MC_ACS_BTEE;
+        mc_tty_frm[MC_TTY_FRM_LEFTMIDDLE] = MC_ACS_LTEE;
+        mc_tty_frm[MC_TTY_FRM_RIGHTMIDDLE] = MC_ACS_RTEE;
+        mc_tty_frm[MC_TTY_FRM_CROSS] = MC_ACS_PLUS;
+    }
 
     // double lines
-    mc_tty_frm[MC_TTY_FRM_DVERT] = mc_skin_lines_load_frm (mc_skin, "dvert");
-    mc_tty_frm[MC_TTY_FRM_DHORIZ] = mc_skin_lines_load_frm (mc_skin, "dhoriz");
-    mc_tty_frm[MC_TTY_FRM_DLEFTTOP] = mc_skin_lines_load_frm (mc_skin, "dlefttop");
-    mc_tty_frm[MC_TTY_FRM_DRIGHTTOP] = mc_skin_lines_load_frm (mc_skin, "drighttop");
-    mc_tty_frm[MC_TTY_FRM_DLEFTBOTTOM] = mc_skin_lines_load_frm (mc_skin, "dleftbottom");
-    mc_tty_frm[MC_TTY_FRM_DRIGHTBOTTOM] = mc_skin_lines_load_frm (mc_skin, "drightbottom");
-    mc_tty_frm[MC_TTY_FRM_DTOPMIDDLE] = mc_skin_lines_load_frm (mc_skin, "dtopmiddle");
-    mc_tty_frm[MC_TTY_FRM_DBOTTOMMIDDLE] = mc_skin_lines_load_frm (mc_skin, "dbottommiddle");
-    mc_tty_frm[MC_TTY_FRM_DLEFTMIDDLE] = mc_skin_lines_load_frm (mc_skin, "dleftmiddle");
-    mc_tty_frm[MC_TTY_FRM_DRIGHTMIDDLE] = mc_skin_lines_load_frm (mc_skin, "drightmiddle");
+    success = TRUE /* just for nicer indentation */
+        && skin_get_char (mc_skin, "dhoriz", 0x2550, &mc_tty_frm[MC_TTY_FRM_DHORIZ])
+        && skin_get_char (mc_skin, "dvert", 0x2551, &mc_tty_frm[MC_TTY_FRM_DVERT])
+        && skin_get_char (mc_skin, "dlefttop", 0x2554, &mc_tty_frm[MC_TTY_FRM_DLEFTTOP])
+        && skin_get_char (mc_skin, "drighttop", 0x2557, &mc_tty_frm[MC_TTY_FRM_DRIGHTTOP])
+        && skin_get_char (mc_skin, "dleftbottom", 0x255A, &mc_tty_frm[MC_TTY_FRM_DLEFTBOTTOM])
+        && skin_get_char (mc_skin, "drightbottom", 0x255D, &mc_tty_frm[MC_TTY_FRM_DRIGHTBOTTOM])
+        && skin_get_char (mc_skin, "dtopmiddle", 0x2564, &mc_tty_frm[MC_TTY_FRM_DTOPMIDDLE])
+        && skin_get_char (mc_skin, "dbottommiddle", 0x2567, &mc_tty_frm[MC_TTY_FRM_DBOTTOMMIDDLE])
+        && skin_get_char (mc_skin, "dleftmiddle", 0x255F, &mc_tty_frm[MC_TTY_FRM_DLEFTMIDDLE])
+        && skin_get_char (mc_skin, "drightmiddle", 0x2562, &mc_tty_frm[MC_TTY_FRM_DRIGHTMIDDLE]);
+
+    // if any of them are unavailable, revert all of them to their single counterpart
+    if (!success)
+    {
+        mc_tty_frm[MC_TTY_FRM_DHORIZ] = mc_tty_frm[MC_TTY_FRM_HORIZ];
+        mc_tty_frm[MC_TTY_FRM_DVERT] = mc_tty_frm[MC_TTY_FRM_VERT];
+        mc_tty_frm[MC_TTY_FRM_DLEFTTOP] = mc_tty_frm[MC_TTY_FRM_LEFTTOP];
+        mc_tty_frm[MC_TTY_FRM_DRIGHTTOP] = mc_tty_frm[MC_TTY_FRM_RIGHTTOP];
+        mc_tty_frm[MC_TTY_FRM_DLEFTBOTTOM] = mc_tty_frm[MC_TTY_FRM_LEFTBOTTOM];
+        mc_tty_frm[MC_TTY_FRM_DRIGHTBOTTOM] = mc_tty_frm[MC_TTY_FRM_RIGHTBOTTOM];
+        mc_tty_frm[MC_TTY_FRM_DTOPMIDDLE] = mc_tty_frm[MC_TTY_FRM_TOPMIDDLE];
+        mc_tty_frm[MC_TTY_FRM_DBOTTOMMIDDLE] = mc_tty_frm[MC_TTY_FRM_BOTTOMMIDDLE];
+        mc_tty_frm[MC_TTY_FRM_DLEFTMIDDLE] = mc_tty_frm[MC_TTY_FRM_LEFTMIDDLE];
+        mc_tty_frm[MC_TTY_FRM_DRIGHTMIDDLE] = mc_tty_frm[MC_TTY_FRM_RIGHTMIDDLE];
+    }
 }
 
 /* --------------------------------------------------------------------------------------------- */
--- a/lib/skin/lines.c
+++ b/lib/skin/lines.c
@@ -128,6 +128,18 @@ skin_get_char (mc_skin_t *mc_skin, const char *name, gunichar def, mc_tty_char_t
     }
 
     // convert to 8-bit charset, e.g. KOI8-R has all the double line chars
+
+#ifdef HAVE_NCURSES
+    // ncurses versions before 6.5-20251115 are buggy in KOI8-R locale, see mc #4799.
+    // Claim failure so that the caller will fall back to ACS single lines.
+    // ASCII chars (e.g. with `--stickchars`) need to fall through.
+    if (ncurses_koi8r_double_line_bug && c >= 128)
+    {
+        g_free (value_utf8);
+        return FALSE;
+    }
+#endif
+
     conv = str_crt_conv_from ("UTF-8");
     if (conv == INVALID_CONV)
     {
--- a/lib/tty/tty-ncurses.c
+++ b/lib/tty/tty-ncurses.c
@@ -164,63 +164,99 @@ tty_clip (int *y, int *x, int *rows, int *cols)
 }
 
 /* --------------------------------------------------------------------------------------------- */
-/*** public functions ****************************************************************************/
-/* --------------------------------------------------------------------------------------------- */
 
-int
-mc_tty_normalize_lines_char (const char *str)
+static chtype
+get_maybe_acs (mc_tty_char_t ch)
 {
-    char *str2;
-    int res;
-
-    struct mc_tty_lines_struct
+    switch (ch)
     {
-        const char *line;
-        int line_code;
-    } const lines_codes[] = {
-        { "\342\224\230", ACS_LRCORNER },  // ┌
-        { "\342\224\224", ACS_LLCORNER },  // └
-        { "\342\224\220", ACS_URCORNER },  // ┐
-        { "\342\224\214", ACS_ULCORNER },  // ┘
-        { "\342\224\234", ACS_LTEE },      // ├
-        { "\342\224\244", ACS_RTEE },      // ┤
-        { "\342\224\254", ACS_TTEE },      // ┬
-        { "\342\224\264", ACS_BTEE },      // ┴
-        { "\342\224\200", ACS_HLINE },     // ─
-        { "\342\224\202", ACS_VLINE },     // │
-        { "\342\224\274", ACS_PLUS },      // ┼
+    case MC_ACS_HLINE:
+        return ACS_HLINE;
+    case MC_ACS_VLINE:
+        return ACS_VLINE;
+    case MC_ACS_ULCORNER:
+        return ACS_ULCORNER;
+    case MC_ACS_URCORNER:
+        return ACS_URCORNER;
+    case MC_ACS_LLCORNER:
+        return ACS_LLCORNER;
+    case MC_ACS_LRCORNER:
+        return ACS_LRCORNER;
+    case MC_ACS_LTEE:
+        return ACS_LTEE;
+    case MC_ACS_RTEE:
+        return ACS_RTEE;
+    case MC_ACS_TTEE:
+        return ACS_TTEE;
+    case MC_ACS_BTEE:
+        return ACS_BTEE;
+    case MC_ACS_PLUS:
+        return ACS_PLUS;
 
-        { "\342\225\235", ACS_LRCORNER },  // ╔
-        { "\342\225\232", ACS_LLCORNER },  // ╚
-        { "\342\225\227", ACS_URCORNER },  // ╗
-        { "\342\225\224", ACS_ULCORNER },  // ╝
-        { "\342\225\237", ACS_LTEE },      // ╟
-        { "\342\225\242", ACS_RTEE },      // ╢
-        { "\342\225\244", ACS_TTEE },      // ╤
-        { "\342\225\247", ACS_BTEE },      // ╧
-        { "\342\225\220", ACS_HLINE },     // ═
-        { "\342\225\221", ACS_VLINE },     // ║
-
-        { NULL, 0 },
-    };
-
-    if (str == NULL)
-        return (int) ' ';
-
-    for (res = 0; lines_codes[res].line; res++)
-        if (strcmp (str, lines_codes[res].line) == 0)
-            return lines_codes[res].line_code;
-
-    str2 = mc_tty_normalize_from_utf8 (str);
-    res = g_utf8_get_char_validated (str2, -1);
-
-    if (res < 0)
-        res = (unsigned char) str2[0];
-    g_free (str2);
-
-    return res;
+    default:
+        return ch;
+    }
 }
 
+/* --------------------------------------------------------------------------------------------- */
+
+static inline void
+addch_maybe_acs (mc_tty_char_t ch)
+{
+#ifdef HAVE_NCURSES_WIDECHAR
+    if (mc_global.utf8_display)
+    {
+        wchar_t wch[2] = { ch, 0 };
+        cchar_t ctext;
+
+        setcchar (&ctext, wch, 0, 0, NULL);
+        add_wch (&ctext);
+    }
+    else
+#endif
+        addch (get_maybe_acs (ch));
+}
+
+/* --------------------------------------------------------------------------------------------- */
+
+static inline void
+hline_maybe_acs (mc_tty_char_t ch, int len)
+{
+#ifdef HAVE_NCURSES_WIDECHAR
+    if (mc_global.utf8_display)
+    {
+        wchar_t wch[2] = { ch, 0 };
+        cchar_t ctext;
+
+        setcchar (&ctext, wch, 0, 0, NULL);
+        hline_set (&ctext, len);
+    }
+    else
+#endif
+        hline (get_maybe_acs (ch), len);
+}
+
+/* --------------------------------------------------------------------------------------------- */
+
+static inline void
+vline_maybe_acs (mc_tty_char_t ch, int len)
+{
+#ifdef HAVE_NCURSES_WIDECHAR
+    if (mc_global.utf8_display)
+    {
+        wchar_t wch[2] = { ch, 0 };
+        cchar_t ctext;
+
+        setcchar (&ctext, wch, 0, 0, NULL);
+        vline_set (&ctext, len);
+    }
+    else
+#endif
+        vline (get_maybe_acs (ch), len);
+}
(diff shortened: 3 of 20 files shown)


lib/mcconfig/get.c
```diff
@@ -120,6 +120,44 @@ mc_config_get_string (mc_config_t *mc_config, const gchar *group, const gchar *p
...
+
+    buffer = g_string_new ("");
+    conv_res = str_convert (conv, ret, buffer);
+    str_close_conv (conv);
+
+    if (conv_res != ESTR_SUCCESS)
+    {
+        g_string_free (buffer, TRUE);
+        g_free (ret);
+        return strdup (def);
```
mc-worker: `g_strdup`

lib/mcconfig/get.c
```diff
@@ -120,6 +120,44 @@ mc_config_get_string (mc_config_t *mc_config, const gchar *group, const gchar *p
...
+        return ret;
+
+    buffer = g_string_new ("");
+    conv_res = str_convert (conv, ret, buffer);
+    str_close_conv (conv);
+
+    if (conv_res != ESTR_SUCCESS)
+    {
+        g_string_free (buffer, TRUE);
+        g_free (ret);
```
mc-worker: Can be moved before test to avoid two calls.

lib/skin/lines.c
```diff
@@ -44,17 +45,109 @@
...
 
-    return ret;
+static mc_tty_char_t
+mc_skin_lines_get_tty_char (mc_skin_t *mc_skin, const char *name, mc_tty_char_t def)
+{
+    GIConv conv;
+    GString *buffer;
+    estr_t conv_res;
+
+    gchar *value_utf8 = mc_config_get_string_raw (mc_skin->config, "Lines", name, "");
```
mc-worker: Please split the variable declaration and assignment.

lib/skin/lines.c
```diff
@@ -44,17 +45,109 @@
...
+    const gunichar c = g_utf8_get_char_validated (value_utf8, -1);
+    if (c <= 0)
```
mc-worker: `gunichar`is unsigned, thus this test is incorrect. It should be
```
if (c == (gunchar) (-1) || c == (gunchar) (-2))
```

lib/skin/lines.c
```diff
@@ -44,17 +45,109 @@
...
+    {
+        g_free (value_utf8);
+        return def;
+    }
+
+    // Prefer MC_ACS_* values, even if there's a codepoint for this character in the current locale.
+    // This gives more flexibility to the ncurses or slang implementation in handling them, and also
+    // prevents us from hitting an ncurses bug (mc ticket #4799#issuecomment-3442057420).
+    const mc_tty_char_t mc_acs = unicode_to_mc_acs (c);
+    if (mc_acs)
```
mc-worker: `if (mc_acs != 0)`

lib/tty/tty-ncurses.c
```diff
@@ -163,64 +163,170 @@ tty_clip (int *y, int *x, int *rows, int *cols)
...
 }
 
-/* --------------------------------------------------------------------------------------------- */
-/*** public functions ****************************************************************************/
 /* --------------------------------------------------------------------------------------------- */
 
-int
-mc_tty_normalize_lines_char (const char *str)
+#if defined(NCURSES_WIDECHAR) && NCURSES_WIDECHAR
+static cchar_t *
```
mc-worker: `static const cchar_t *`?

lib/tty/tty.h
```diff
@@ -25,13 +25,50 @@
...
 
+// For frame characters (single or double) it's one of the MC_ACS_* special values.
+// Otherwise the character code in the locale's encoding.
+typedef unsigned int mc_tty_char_t;
+
 /*** enums ***************************************************************************************/
 
+// These values always represent the given line drawing characters, in a common way regardless of
+// the underlying screen library.
+enum mc_tty_char_t
```
mc-worker: `mc_tty_char_t` is defined above already.

lib/tty/tty-slang.c
```diff
@@ -649,65 +586,139 @@ tty_display_8bit (gboolean what)
 
 /* --------------------------------------------------------------------------------------------- */
 
-void
-tty_print_char (int c)
+static int
+get_maybe_wacs (mc_tty_char_t c, gboolean *alt_char)
```
mc-worker: Static functions should be placed in the section "file scope functions".

mc-worker (commented):
I believe, po-files in this PR are unintentionally.

egmontkob:
@mc-worker Thanks for the review! I'll address them later. I'm glad that it's only small details, nothing wrong with the overall refactoring.

Sneaked in a 3rd commit for fun.

The .po files are indeed accidental; somehow they often get modified during make / git rebase / not sure what step exactly, I didn't touch them manually.

Let's wait for a fix to #4817 so that I can apply the same ncurses build fix here in this PR as well.

mc-worker:
This is really resolves #4799 but there are no double lines anymore in mc/S-Lang in koi8-r locale.
No problem if this is expected result.

egmontkob:
> no double lines anymore in mc/S-Lang in koi8-r locale.
> No problem if this is expected result.

This is expected, and documented in the commit message. Let me give you more info:

With the clear separation of logical position ("role") vs. actual glyph (the character taken from the skin to used), and for the latter with the introduction of the screenlibrary-agnostic new type `mc_tty_char_t` which can hold either a char in the locale or a frame drawing char, I faced a dilemma:

Sometimes a character is available in the locale but is not a frame drawing one. Then we use its codepoint. Sometimes it is a frame drawing one but isn't available in the locale (may or may not be available as an ACS), then we use one of the special values denoting frame characters. Sometimes it's neither, then we fall back to a simple ASCII char. But sometimes it's both; in that case which one to choose for `mc_tty_char_t`?

In the first work iterations of the fix, I went with the character code according to the locale. Disadvantages: KOI8-* are heavily broken in [certain ncurses versions](https://github.com/MidnightCommander/mc/issues/4799#issuecomment-3442057420). The presumably slightly optimized method of `SLsmg_draw_hline()` won't be used and we resort to printing each char separately (not a big deal at all).

So I changed my mind and went with preferring the MC_ACS_ constants. The problem is: For double line chars the place where I actually handle these chars I'd need to perform a locale-specific conversion to their codepoint (or do it somewhere once at startup and remember the result). It could be done, no problem, it just needs a bit of extra work that I really didn't feel like putting in.

Or: `mc_tty_char_t` could be a struct of two integers: both the codepoint and the line drawing magic constant, whichever (or maybe both) are available. Again I don't think it's worth it.

---

There's only 1 skin, "double-lines", which so far looked fine with its double lines in KOI8-R with slang. All the other skins either have single lines, or have special Unicode symbols for dropdown arrows and such that are currently broken (fixed by this PR). No prior bugreport about them suggests that hardly anyone cares, presumably hardly anyone uses non-UTF-8 locales anymore.

The ncurses bug going unnoticed for 1.5+ years also suggests that KOI8-R must hardly be used as a locale anymore; I presume other ncurses-based apps suffer from noticeable display issues, not just my unpublished draft of this PR where I faced that bug.

[In fact, I'm somewhat tempted to file an issue to remove support for locales other than UTF-8 and plain ASCII, if this simplifies things in mc. (I don't want to remove Alt-E's ability to change the encoding assumed for filenames or file contents, I'm talking about the charset of the host terminal.) While some pieces of software out there might still struggle with UTF-8, I see no reason for anyone for not having already swithed to UTF-8 by default, reverting back to legacy encodings only for the particular tasks where it's still necessary. Many of the new terminal emulators (and there are oh such plenty of them!) don't even support any other encoding than UTF-8.]

And one more thing: KOI8-U contains most, but not all of the double line drawing characters. Accordingly, current mc with slang and the "double-lines" skin looks slightly broken, at a few of these chars (e.g. `╢` is a `b` instead). If I added the extra special code to use the actual codepoints for double line drawing chars in non-UTF-8 if available, i.e. if I fixed this regression you pointed out with KOI8-R, then I'd also re-introduce the broken behavior of KOI8-U. Or I would have to noticeably complicate the code, "batch up" those characters: try to convert all 10 of them to the current locale and if any of them fails then fallback all of them to single-lines ones; it's really not worth it.

Hope this makes sense.

misc/skins/sand256.ini
```diff
@@ -61,21 +61,21 @@
 [Lines]
     horiz = ─
     vert = │
-    lefttop = ┌
-    righttop = ┐
-    leftbottom = └
-    rightbottom = ┘
+    lefttop = ╭
```
zyv, line 59: Just to clarify, is this intentional?

zyv:
> So I changed my mind and went with preferring the MC_ACS_ constants. The problem is: For double line chars the place where I actually handle these chars I'd need to perform a locale-specific conversion to their codepoint (or do it somewhere once at startup and remember the result). It could be done, no problem, it just needs a bit of extra work that I really didn't feel like putting in.
> 
> Or: `mc_tty_char_t` could be a struct of two integers: both the codepoint and the line drawing magic constant, whichever (or maybe both) are available. Again I don't think it's worth it.

Given the amount of work (as it appears to me) that went into this fix, doing a cache table does not sound like a major addition to that. However, I guess I should let **you** be the judge of where you want to stop with this...

misc/skins/sand256.ini
egmontkob: See the commit log.

I think it would be cool to have at least one skin demonstrate the use of rounded corners, just for the sake of it, and to give an idea to future skin developers.

With the changes in this PR, this now falls back to regular line drawing in non-UTF-8 locales rather than displaying garbage, so no damage is done.

This was the first 256-color skin, created by me, and honestly it's not a great skin, probably one of the worst ones (if not _the worst_ we have), so I hope noone would mind this change.

egmontkob:
> doing a cache table does not sound like a major addition to that

It wouldn't be a _major_ addition, but it would be some addition just for the sake of one specific charset KOI8-R, while having to be careful not to use any double line chars in KOI8-U.

For me it's not the amount of code; it's the amount of code relative to its usefulness. The code I wrote addresses all 8-bit charsets and even works around an ncurses bug, these are generic things. A slight improvement for just one speficic charset, KOI8-R, is just way too specific.

[Do people out there still use KOI-R at all, combined with the too-dark-blue-to-my-taste "double-lines" skin (the only one where currently double lines show up with slang without other display problems), for this to _really_ matter?]

egmontkob:
... oh, and the fix would be slang-only. Or it could fix it on ncurses too, subject to it not being broken, which means I'd have to go into the business of checking ncurses's version. Which is easy at compile time, but the runtime version might be different. For runtime there's `curses_version()` but I have to manually parse and compare that string. It's just getting really ugly.

zyv:
> [Do people out there still use KOI-R at all, combined with the too-dark-blue-to-my-taste "double-lines" skin (the only one where currently double lines show up with slang without other display problems), for this to _really_ matter?]

Just to address this point, there still seems to be quite a following of KOI8-R on Russian-speaking Linux admin forums, and Andrew seems to be using it. I'm not aware of any living users of KOI8-U for that matter. I'm not using anything other than UTF-8 these days. Note that I'm just trying to answer this question, not to convince you one way or another.

> ... oh, and the fix would be slang-only. Or it could fix it on ncurses too, subject to it not being broken, which means I'd have to go into the business of checking ncurses's version. Which is easy at compile time, but the runtime version might be different. For runtime there's `curses_version()` but I have to manually parse and compare that string. It's just getting really ugly.

Okay, I think this is the real argument for not doing it and stopping where you stopped.

misc/skins/sand256.ini
zyv: Ah, okay, sorry, unlike @ossilator, I actually mostly look at the "Files changed" tab, and then look at per-commit steps if something looks unclear and/or unclean.

So the reason I'm asking is, I'd actually prefer more skins to use that if we add it to one skin. I wouldn't necessarily want to change the skins submitted by others, but I think all 16M skins are yours, and some 256-color ones. Would you mind making all of your skins use them?

egmontkob: Honestly, I don't like the look of rounded corners that much. :D

Plus, there's no rounded double lines, and my `gray*256` and `seasons*16M` skins use double lines, I'd prefer to keep it that way. 

That's why I'd like to keep it "hidden" in an ugly skin of mine, if this makes sense.

lib/mcconfig/get.c
egmontkob: Done.

lib/mcconfig/get.c
egmontkob: Done.

lib/skin/lines.c
egmontkob: Done.

lib/skin/lines.c
egmontkob: Nice catch, fixed.

lib/skin/lines.c
egmontkob: Done.

lib/tty/tty-ncurses.c
egmontkob: Done.

lib/tty/tty-slang.c
egmontkob: Done.

lib/tty/tty.h
egmontkob: Fixed.

misc/skins/sand256.ini
zyv: Okay, thanks for the explanation.

egmontkob:
The line art "Midnight" logo in the F1 help page is broken in ncurses + UTF-8.

egmontkob:
> The line art "Midnight" logo in the F1 help page is broken in ncurses + UTF-8.

Fixed.

For the logo, the `xnc.hlp` file uses the "ACS notation" (tons of letters l q w k x ...), between custom mc-specific markers.

As much as I'd rather see real Unicode instead, it's not easily doable. The entire file is first converted to the locale's encoding, and then formatted (lines wrapped etc.) for the UI. Obviously, during charset conversion these chars would get lost if converted to 8-bit.

So instead I just added a new method that converts these letters to the new mc-specific constants, and then pass them to the unified new slang/ncurses methods that can handle them. One less slang vs. ncurses branching outside of the tty code, yay!

egmontkob:
> there are no double lines anymore in mc/S-Lang in koi8-r locale

I might actually fix this, just because why not :)

> For runtime there's curses_version() but I have to manually parse

I figured out it's enough to do two `strcmp()`s against the first and last broken ncurses version string. (It'll break if ncurses ever gets to version 6.40, which it won't, and if does then it'll be high time to drop that workaround.:))

mc-worker (approved):
Cool!

misc/skins/sand256.ini
ossilator: fwiw, this would be neatly addressed by #3169 ...

egmontkob: mc devs weren't keen on that idea back then, and I don't want to resurrect that discussion, it's just not that important for me. I've accepted that these remain defined by the skins.

egmontkob:
Bit cleanup uploaded. Even more weird chars in that ugly skin. Addressing another minor issue with the help pages's logo vs. stickchars.

No double lines in koi8-r yet.

egmontkob:
Double-lines + koi8-r pushed.

To be cleaned up (comments, silence a gcc warning, commit log) tomorrow.

egmontkob:
In the previous version, double lines with koi8-r sometimes works, sometimes doesn't. [Incorrect online glib documentation](https://gitlab.gnome.org/GNOME/glib/-/issues/2778), yay!

The next push (soon) will fix it.

egmontkob:
Minor comment updates; rebased.

zyv:
> Minor comment updates; rebased.

So is that now ready to merge in as far as you are concerned?

egmontkob:
Some of the method names (e.g. whether to begin with `mc` or `tty`) I'm not necessarily satisfied with, also whether things to into `tty` or `tty-internal` and how each one references the other's header (which is not necessarily bad), also how things are already inconsistent (e.g. `mc_tty_normalize_from_utf8()` is in `tty-internal.h` and `tty.c`), ideas how to clean up are welcome

But as for the big picture, I've done what I wanted, so it's only your feedback that I would address.

zyv:
> Some of the method names (e.g. whether to begin with `mc` or `tty`) I'm not necessarily satisfied with, also whether things to into `tty` or `tty-internal` and how each one references the other's header (which is not necessarily bad), also how things are already inconsistent (e.g. `mc_tty_normalize_from_utf8()` is in `tty-internal.h` and `tty.c`), ideas how to clean up are welcome
> 
> But as for the big picture, I've done what I wanted, so it's only your feedback that I would address.

How can I actually properly test this? I have built mc with ncurses and with everything UTF-8. As I ssh into Alpine from macOS using Terminal, it looks like it does with S-Lang (which I guess is a good thing).

Now, all I could think of was to `export LC_ALL=C` on macOS, ssh into Alpine, and `export LC_ALL=C` before I started mc. The result is as shown below. I guess this is not expected, is it?

Screenshot 2025-11-04 at 09 35 20

egmontkob:
Yup, this definitely doesn't look good.

Can you please send a `script` log? Also, is this with ncurses or slang?

Is this a regression, does current mc show up nicely in the same setup?

Do you have another terminal emulator (e.g. iTerm2) to test with?

---

How to test the PR in general:

Try all :stuck_out_tongue: combinations of (I know it's hopeless, just test a few randomly)

- Ancient ncurses in narrow or wide build, new-ish ncurses in narrow or wide build, slang.

- Various locales. Generate locale definitions for let's say `en_US` (ISO-8859-1), `ru_RU.KOI8-R`, `uk_UA.KOI8-U`. Verify with `LC_ALL=... locale charmap`. Always make sure to switch the terminal emulator's charset to the same value, otherwise it's expected that everything falls apart big time.

- Default skin, double line skins, fancy unicode char skins, new rounded corner skin.

- Various terminal emulators.

egmontkob:
In that screenshot, _some_ of the horizontal line characters do appear properly, _some_ become a space, while _some_ become an `R`. So it's not a consistent faulty behavior or a consistent bad replacement of characters.

Also, the first horizontal line in a sequence always shows up correctly; e.g. near the left edge of the error box, or after the word "Error", or at the top of the panels before those many spaces, or at the bottom before those many `R`s.

This makes me suspect that something with the REP "repeat the preceding graphic character n times" escape sequence goes wrong. The graphic library (which one?) emits one horizontal line, followed by this escape sequence to get it repeated.

Maybe the terminal doesn't support it, while terminfo claims it does. Maybe it doesn't work with ACS line drawing. Something along these lines, that's my current guess.

egmontkob:
Those `R`s at the bottom... I suspect it's the last letter of the word "UP--DIR" painted a line above. Maybe your terminal repeats the last non-ACS character.

Can you please confirm this theorem? Move the cursor down, e.g. to the next entry `.cache`, press Ctrl-L to repaint the screen, do you now see tons of `e` there?

Try to do this in the right panel, probably it's the right one that's printed last, so that gets repeated?

zyv:
> Those `R`s at the bottom... I suspect it's the last letter of the word "UP--DIR" painted a line above. Maybe your terminal repeats the last non-ACS character.
> 
> Can you please confirm this theorem? Move the cursor down, e.g. to the next entry `.cache`, press Ctrl-L to repaint the screen, do you now see tons of `e` there?
> 
> Try to do this in the right panel, probably it's the right one that's printed last, so that gets repeated?

Your theory is right on the spot!

Screenshot 2025-11-04 at 14 54 21

zyv:
> Yup, this definitely doesn't look good.
> 
> Can you please send a `script` log? Also, is this with ncurses or slang?

This is ncurses. I don't know how to make a script log, but if you tell me, I can make one.

> Is this a regression, does current mc show up nicely in the same setup?

Stock mc 4.8.31 (S-Lang) looks normal with Terminal set to ISO and LC_ALL=C:

Screenshot 2025-11-04 at 15 00 38

With LC_ALL=en_US it looks like this:

Screenshot 2025-11-04 at 15 02 00

I didn't check ncurses build before this PR (yet).

> Do you have another terminal emulator (e.g. iTerm2) to test with?

No.

> How to test the PR in general:
> 
> Try all 😛 combinations of (I know it's hopeless, just test a few randomly)
> 
> * Ancient ncurses in narrow or wide build, new-ish ncurses in narrow or wide build, slang.
> * Various locales. Generate locale definitions for let's say `en_US` (ISO-8859-1), `ru_RU.KOI8-R`, `uk_UA.KOI8-U`. Verify with `LC_ALL=... locale charmap`. Always make sure to switch the terminal emulator's charset to the same value, otherwise it's expected that everything falls apart big time.
> * Default skin, double line skins, fancy unicode char skins, new rounded corner skin.
> * Various terminal emulators.

Here is macOS Terminal set to ISO-8859-1, export LC_ALL=en_US on macOS and Alpine, and ssh into Alpine with mc from this PR, so similar to stock with S-Lang:

Screenshot 2025-11-04 at 14 57 26

egmontkob:
rebased, added fix for #4860.

zyv:
With LC_ALL=en_US and Terminal set to Latin-1, it's not getting any better...

Screenshot 2025-11-06 at 13 32 15
Screenshot 2025-11-06 at 13 32 31

With LC_ALL=C still looks as posted before.

I guess we can claim that it's a combination of Alpine and macOS Terminal, and after I have ascertained that it's not a regression (which I don't think it is), I think I have to try on Fedora...

egmontkob:
@zyv Somehow I totally missed your replies 2 days ago.

As for the missing horizontal lines, or `RRRRRRRRRRR` looking ones, but other than that an overall correct layout:

> Your theory is right on the spot!

:-)

I think it boils down to ncurses emitting the REP escape sequence, and Terminal.app incorrectly handling them.

I am fairly certain that an ncurses build without my patch results in the same corrupted behavior (but could you please teset it?), and I can't think of anything we could do about it, the bug is below us in the stack.

Workarounds include:
- build with slang
- switch to a terminal that isn't buggy (iTerm2 should be in pretty much all aspect superior to Terminal.app, or at least was ~15 years ago :))
- install a new terminfo description which removes the REP capability so that ncurses doesn't use it
- use `--stickchars`

If the same bug occurred with UTF-8, I'd have two ideas to try:

- `export NCURSES_NO_UTF8_ACS=1` before starting up mc
- modify mc to use the Unicode values rather than `ACS_*`

But you said the bug occurs in the C locale where ACS (alternate character set) is the only approach to paint line drawing characters, so these ideas couldn't work there.

egmontkob:
> I don't know how to make a script log, but if you tell me, I can make one.

Not relevant anymore where I suggested this, but might become handy later (maybe even for the other issue that you posted and I'll take a look in a second).

`script` logs the data flow from the application towards the terminal. Start `script`, you're dropped in a subshell. Start mc, reproduce the bug. Quit mc. Quit that shell that was started by `script` (don't forget this step! - or pass parameters to `script` to launch `mc` directly without that shell). Look at the created `typescript` file, it's literally all the data that mc sent to the terminal.

On mac, usage might be somewhat different. 

If you ssh, you might do it on either side, as ssh itself doesn't modify the traffic, the logged output should be the same.

(Make sure the data doesn't contain anything private before you'd attach. E.g. sensitive filenames listed in the right panel, covered in the screenshot by the F1 help page, these will be there.)

egmontkob:
Just realized that this PR, namely the big change titled "Clean up and fix frame drawing", accidentally fixes rendering in PuTTY / pterm :-)

Previously you had to either `export NCURSES_NO_UTF8_ACS=1` or enable PuTTY's Window -> Translation -> Enable VT100 line drawing even in UTF-8 mode. Otherwise, by default, you'd get `lqqqqqqqqqqqk`-style horizontal lines.

If built against ncursesw and running under an UTF-8 locale, this patch changes mc to use the widechar method and constant rather than the 8-bit one.

`NCURSES_NO_UTF8_ACS=1` only affects how ncurses maps its legacy 8-bit `ACS_*` constants, used in legacy 8-bit methods like `addch()`, to output sequences if the locale is UTF-8: whether to emit ACS or UTF-8.

If Unicode-compatible `add_wch()` or friends are used with `WACS_*` in the source, and the locale is also UTF-8, then luckily ncurses always emits raw UTF-8.

egmontkob:
As for the three more severely broken screenshots:

They are somewhat similar to the screenshot of #4301, in the sense that there are a lot of `â` (U+00E2) characters. Yet, while over there lines that are affected become longer, here they become shorter.

`â` is the first byte of the UTF-8 line drawing characters, if interpreted in Latin-1.

So I'm quite certain that mc emits UTF-8 while the terminal expects Latin-1.

The next two bytes usually disappear, probably because they are usually in the 0x80 .. 0x9F range which are unprintable in Latin-1. (That other ticket used a Windows-x encoding where these are printable characters.)

Could you please double-check the locale on the _remote_ site? What does `locale charmap` say before launching mc, or even from mc's subshell?

Is this a regression introduced by this PR, or is it the same in 4.8.33 or current master, with the same build options?

egmontkob:
Updated the commit log to mention the "accidental" PuTTY fix. No code change.

zyv:
This is `LC_ALL=en_US` and Latin1 encoding in Konsole.

Screenshot 2025-11-07 at 21 02 18

I thought I could try `gnome-terminal`, but it doesn't render properly under virtualization. :(

zyv:
> As for the three more severely broken screenshots:
> 
> They are somewhat similar to the screenshot of #4301, in the sense that there are a lot of `â` (U+00E2) characters. Yet, while over there lines that are affected become longer, here they become shorter.
> 
> `â` is the first byte of the UTF-8 line drawing characters, if interpreted in Latin-1.
> 
> So I'm quite certain that mc emits UTF-8 while the terminal expects Latin-1.

I did one more test now: installing iTerm2, switching it to a single-byte mode, and using `LC_ALL=C` on both sides. I think that the conclusion is that `â` stuff is caused by your suggestion to use `en_US` on both sides. The only thing that really seems to work to make it single-byte is `C`, not sure which side (macOS or Alpine) is the problem. It may be Alpine, since I seem to remember that musl doesn't support anything else other than UTF-8 for locales the last time I tried to fix some bugs on Alpine.

In any case, when both sides and the terminal agree on single bytes, I don't see double lines, and also, somehow F1 doesn't open. Neither of those is a regression. But what I'm really left wondering is, how can I make this PR finally make some difference for me?

```
alpine:~$ mc -V
GNU Midnight Commander 4.8.31
Built with GLib 2.80.0
Built with libssh2 1.11.0
With builtin Editor
Virtual File Systems:
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64;
 ```

Screenshot 2025-11-08 at 08 43 27


 ```
 alpine:~$ ~/src/mc/build/install/bin/mc -V
GNU Midnight Commander 4.8.33-434-g9b2ed0a2a
Built with GLib 2.80.5 (using GLib 2.80.5)
Built with NCursesw 6.4.20240420 (using ncurses 6.4.20240420)
With builtin editor
With mouse support on xterm
Virtual File Systems:
Data types:
 ```

Screenshot 2025-11-08 at 08 44 11

**Comment by zyv** [comment]
New test on Fedora, this time comparing stock with S-Lang, master with ncurses, and this PR with ncurses.

I think I understand the intent of the PR to make double lines non-broken in that they should be replaced with single if not available, is that right? But why then it seems to work on master already, and seems to be broken with S-Lang o_O

```
zaytsev@fedora:~$ mc -V
GNU Midnight Commander 4.8.32
Built with GLib 2.82.2
With builtin editor
Virtual File Systems:
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64;
```

Screenshot 2025-11-08 at 09 02 03


```
zaytsev@fedora:~$ ~/src/mc/build-ncurses/install/bin/mc -V
GNU Midnight Commander 4.8.33-433-g2ebb1dd50
Built with GLib 2.82.5 (using GLib 2.82.5)
Built with NCursesw 6.5.20240629 (using ncurses 6.5.20240629)
With builtin editor
With support for X11 events
Virtual File Systems:
Data types:
```
Screenshot 2025-11-08 at 09 02 43



```
zaytsev@fedora:~$ ~/src/mc-test/build-ncurses/install/bin/mc -V
GNU Midnight Commander 4.8.33-439-gefc6c6106
Built with GLib 2.82.5 (using GLib 2.82.5)
Built with NCursesw 6.5.20240629 (using ncurses 6.5.20240629)
With builtin editor
With support for X11 events
Virtual File Systems:
Data types:
```

Screenshot 2025-11-08 at 09 03 13

**Comment by egmontkob** [comment]
(I had this email thread muted accidentally, that's why I missed your comments the last time, and now too.)

There are so many different setups, so many different variables (which terminal, what locale, what mc version, which screen library...) and so many bugs in your screenshots that honestly I'm totally lost.

Is there any set of circumstances where this PR, but apart from that identical setup (same terminal, same OS, same locale, same screen drawing library) causes a regression? If so, what's that setup, and what's the before and after screenshot?

If there is such a case, it needs to be fixed as part of this PR. If there's none, just remaining (but already existing) bugs then I think they should be tracked down independently and not hold back this PR.

> It may be Alpine, since I seem to remember that musl doesn't support anything else other than UTF-8

It could be, I don't know. What does `LC_ALL=en_US locale` and `LC_ALL=en_US locale charmap` report on Alpine?

**Comment by egmontkob** [comment]
> when both sides and the terminal agree on single bytes, I don't see double lines

Most of the 8-bit charsets don't contain double lines, and they aren't available in the special ACS mode either. So there's literally no way to get them displayed. In such cases they should be replaced by single lines.

An exception is KOI8-R, that _does_ contain all the double lines we need. With such a locale, you should see double lines – except if using ncurses from about 1.5 years ago until very recently, the bug that you forwarded to Thomas.

> But what I'm really left wondering is, how can I make this PR finally make some difference for me?

Fair question :) E.g. ncurses didn't have double lines so far (not even in UTF-8), now it has whenever the terminal supports it (except with KOI8-R in those broken ncurses versions, they should revert to single lines). Another thing is that in Latin-1 and similar locale all the fancy colored skins had display glitches at dropdown arrows and such.

Please see the linked tickets for steps to reproduce the bugs that this PR should fix.

**Comment by egmontkob** [comment]
> I think I understand the intent of the PR to make double lines non-broken in that they should be replaced with single if not available, is that right?

It's one of the intents, yes.

> But why then it seems to work on master already, and seems to be broken with S-Lang o_O

It already works _sometimes_, but not necessarily in every setup. In some other cases it might be broken, or unnecessarily downgrade to single lines when double ones are available.

As for S-Lang: The screenshot where you asked this shows 4.8.32, the question is: is it still broken with this PR?

File: lib/tty/tty-internal.c [hunk 10]
```diff
@@ -35,16 +35,20 @@
...
 
+#include "tty.h"
 #include "tty-internal.h"
 
 /*** global variables ****************************************************************************/
 
 /* pipe to handle SIGWINCH */
 int sigwinch_pipe[2];
 
+GHashTable *double_line_map = NULL;
```
mc-worker: This variable is not global, it is used only in this file.

lib/tty/tty-internal.c
```diff
@@ -108,3 +112,94 @@ tty_destroy_winch_pipe (void)
...
+        0x2554,  // ╔
+        0x2557,  // ╗
+        0x255A,  // ╚
+        0x255D,  // ╝
+        0x255F,  // ╟
+        0x2562,  // ╢
+        0x2564,  // ╤
+        0x2567,  // ╧
+    };
+    gunichar double_lines_local[G_N_ELEMENTS (double_lines_unicode)];
```
mc-worker: `G_N_ELEMENTS (double_lines_unicode)` is used 3 times in this function. What about a constant?

lib/tty/tty-internal.c
```diff
@@ -108,3 +112,94 @@ tty_destroy_winch_pipe (void)
...
+
+/* --------------------------------------------------------------------------------------------- */
+
+gunichar
+tty_double_line_map_lookup (mc_tty_char_t double_line)
+{
+    if (double_line_map == NULL)
+        return 0;
+
+    void *direct_value = g_hash_table_lookup (double_line_map, GINT_TO_POINTER (double_line));
```
mc-worker: Please move the variable declaration to the begin of block.

lib/tty/tty-internal.c
egmontkob: Indeed, will fix. (Probably leftover of earlier iterations of the work.)

lib/tty/tty-internal.c
egmontkob: It's evaluated compile-time, so no runtime win from placing it in a variable. I feel like extracting it would just introduce a level of indirection that wouldn't make it simpler to read the code, the opposite (a little bit). I'd prefer to keep it this way.

lib/tty/tty-internal.c
egmontkob: Style guide says either to move the declaration to the beginning of the block, or to use `const` to store an intermediate value. This is indeed an intermediate (and it's unclear to me why it's needed, but this silences a gcc warning that I couldn't silence just with explicit casting).

Which solution is preferred, non-const declaration at the top, or `const` in the middle of the code? I'm leaning towards the latter.

egmontkob: ... mind you, the `const` in `const void *` is not the same as the `const` in `const int`. It would be `const void const *` or something like that, but C is quite badly sloppy here, only C++ cleans this up. So maybe _this_ is the reason moving to the top would be preferred here?

zyv: > Which solution is preferred, non-const declaration at the top, or `const` in the middle of the code? I'm leaning towards the latter.

I've had a long argument about that with Andrew, and my position is that const is always safer and reduces the mental load because you can assume that the variable will not change, at least by design (okay, let's forget all the limitations of const in C for a moment...). Placing it near the usage in the middle of the code helps to reduce scope and avoid side effects by mistake if declaring early. I think in the end he was able to accept that.

What I wasn't able to get through was to join the declarations and initialization for non-const variables. So I had to agree to a compromise, and we still have this stuff at the beginning of the block. The main argument is that having things you use declared in one place is better for cognitive load, as in the times of Pascal, which of course is not really valid if you use modern IDEs and move towards immutability anyways, but so it is.

egmontkob: Let me try `const gconstpointer`. It's "const" in both senses, and is a closer match to the signature of the glib method as well as the following explicit cast (which both use `gpointer`, which, as far as I can tell, is simply glib's preferred way of saying `void *`).

egmontkob:
@zyv, as far as I can tell, all your broken screenshots are either with macOS Terminal.app which we suspect doesn't handle REP correctly, or using an older mc.

Is there any setup where this PR causes a regression to you?

zyv (approved):
Okay, I tested it once more on Fedora, this time in Konsole with a single-byte locale (`ru_RU.koi8r`), and as it so happens, I have the wrong version of ncurses, so I couldn't get it to display double lines either.

I agree that the issues that surfaced in my naive and limited testing either stem from macOS Terminal, or limitations of my testing setup (Linux VMs with Alpine and Fedora), or both. Likewise, it seems none of those are regressions, but rather pre-existing issues. I even seem to remember that I've already seen the issue with help not opening on Alpine already, but never got time to document and/or investigate.

I'm still way too overwhelmed to be able to provide any more extensive testing (and this isn't going to change anytime soon, unfortunately), but code-wise, it now looks good to me, and is definitively a net improvement.

I think, under the circumstances, this PR should be merged, unless someone has strong objections.

zyv:
> /rebase

Alright, fat chance. I think this PR has maintainer edits disabled :( I can't rebase it.

egmontkob:
Hang on for a sec, don't merge yet.

I just had an idea how to slightly refactor the slang double line drawing hack to be less hacky.

Let me give it a try.

egmontkob:
rebased (don't merge)

egmontkob:
Okay, here's how it should've looked like from the beginning.

By moving the "double -> single fallback if double is not available" from the tty code to the skin code, the big refactoring became noticeably simpler; plus, double lines in koi8-r work out of the box, no need for another convoluted change on top just to fix the koi8-r + double line case (still need a tiny workaround against the ncurses bug to revert to single lines).

Let's give it another thorough round of testing and review, please.

zyv:
I can only offer to repeat some basic tests on the systems mentioned above.

egmontkob:
Possible remaining cleanups on top of these changes (maybe in this PR, maybe in a followup):

---

Maybe rename the `MC_TTY_FRM_*` constants and `mc_tty_frm` array to `MC_SKIN_FRM_*` and `mc_skin_frm`, or similar. They really don't have much to do with the tty, they are about the skin, caching for efficientcy whatever characters the skin defines.

(While at it, I'd prefer a generic "frm" -> "frame" rename, "frm" could stand for "form", "from" etc. Saving two letters reduced readability.)

---

`tty.[ch]` have three methods remaining that are aware of the skin (and ideally it shouldn't have any, it should implement whatever the caller requests and not know anything about skins) -- (((a remotely parallel analogy to the ongoing PR where widgets shouldn't translate the received strings))).

`tty_print_one_hline()` is unused both before and after this PR. `tty_print_one_vline()` is used at about a dozen places, it can easily be inlined at all of them.

The third remaining one would be `tty_draw_box()`, a pretty generic box drawing method that is aware of the skin, and is complex enough not to be feasible to inline. Is there a better place for this, like a generic drawing helper method, above the tty layer?

zyv:
> Possible remaining cleanups on top of these changes (maybe in this PR, maybe in a followup)

I'm sorry, but could you please clarify what do you currently expect from me? Should I try to squeeze some time to have another look at the code and do some final tests, and merge if I see no problems, or should I wait for further changes? I'm asking because my schedule is getting even tighter before I'm off.

egmontkob:
If you don't see any issue, I'd like to see it merged.

I couldn't find any regression, apart from having double lines in ncurses now triggering #4892 on Solaris way more frequently.

Followup cleanups, if at all, might come later in a separate change. They would be internal cleanups, with no visible outcome.

lib/tty/tty-internal.c
egmontkob: It's all obsolete now, a bigger refactoring of the pending PR made this part of the code go away.

egmontkob:
@mc-worker The code changed quite a bit since you looked at it, another round of review would be appreciated!

egmontkob:
rebased; enabled koi8-r double lines with newest ncurses

zyv:
UTF-8, left is master, right is this PR

Screenshot 2025-11-20 at 14 45 38

KOI8-R

Screenshot 2025-11-20 at 14 44 46

Screenshot 2025-11-20 at 14 41 56

egmontkob:
Thanks for the screenshots!

The two blue pairs: Thick lines are removed as per #4860 (note the bad look of short bold stems near the bottom of the panels). You can modify the skin definitions to contain the thick line drawing Unicode chars, although they probably look even thicker.

If you'd prefer to revert to the previous look with bolder and brighter lines (which used to be for ncurses only), the proper way would be to add a new color definition for the panel frames, and specify the `bold` attribute there. This would then work for both ncurses and slang. Note that the exact look depends on the terminal, many terminals don't bolden the frame chars.

Second blue pair, right hand: No double lines because ncurses is buggy here with KOI8-R. Download latest ncurses (6.5.20251115), `./configure --with-shared --without-cxx --without-cxx-binding && make`, then point `LD_LIBRARY_PATH` to its `lib` directory, you should get proper double lines. Or use slang.

Yellow screenshot: A few unsupported characters are now detected and replaced with ASCII. This skin has also been modified to use rounded corners and dashed lines, you should see those in UTF-8. In KOI8-R their lack is detected, hence the fallback to ASCII.

Does this make sense?

zyv:
> Does this make sense?

Sure, sorry that I was not able to leave more comments after testing. What I meant to say is that I've now tested the PR in multiple and single-byte locales as compared to master and couldn't see any regressions, but on the contrary, the current behavior seems to be an improvement, so I don't see this as a reason against merging.

egmontkob:
Glad to hear it!

That being said, I guess it might make sense to extract the frame colors and attributes to new skin properties? Editor already has "editframe", analogously there could be core / frame, dialog / dframe, error / errdframe, help / helpframe, etc.

Low prio optional followup work in a separate PR, maybe only if people complain about the frame no longer being as prominent as it used to be on ncurses.

zyv:
> That being said, I guess it might make sense to extract the frame colors and attributes to new skin properties? Editor already has "editframe", analogously there could be core / frame, dialog / dframe, error / errdframe, help / helpframe, etc.

You know, if we are talking about eye candy, what always bothered me throughout the last decade is that the panel columns do not cleanly join the top and bottom lines. FAR Manager doesn't have this problem. It really feels like a splinter - nagging, but not painful enough to let everything go and take care of it right away. So far, I wasn't able to get my act together to do anything about it :(

516881613-31c3544a-f155-4a27-9b34-a29abb5c07c8

egmontkob:
> what always bothered me throughout the last decade is

I know: #2109 :)

I always considered it a feature to have a more airy look than in those other commanders. But of course I wouldn't mind to see it "fixed".

egmontkob:
(reabsed)

egmontkob:
so... okay to merge? :)

zyv:
> so... okay to merge? :)

Maybe it's a good idea for you to do it while I'm still around?

egmontkob:
Okay, I'm pressing the big green button on this page and see what happens. Then I'll update the wiki.

PR #4816, opened 2025-10-24
Ticket #4815: Fix no colored prompt after switching panels off for the first time

### LINKED ISSUE #4815: No colored prompt after switching panels off first time

```shell
git master
```

### Operating system

```shell
any with subshell support
```

### How to reproduce

1. Set the `PS1` variable to a colored prompt (see `~/.bashrc` on most distributions)
2. Launch mc in a terminal with subshell support
3. Press `Ctrl+O`
4. A black screen appears, showing only a white prompt
5. Press `Enter`
6. The colored prompt appears from now on

### Expected behavior

After pressing `Ctrl+O`, the colored prompt should appear

### Actual behavior

See above

### Additional context

_No response_

horkykuba:
Fix the issue by strictly distinguishing between `mc_prompt`, which should be stripped using `strip_ctrl_codes()`, and `subshell_prompt`, which should always hold the original, non-stripped `PS1`. Make `mc_prompt` heap-based.

* Resolves: #4815

### DIFF
--- a/src/filemanager/layout.c
+++ b/src/filemanager/layout.c
@@ -974,7 +974,6 @@ setup_cmdline (void)
     const WRect *r = &mw->rect;
     int prompt_width;
     int y;
-    char *tmp_prompt = (char *) mc_prompt;
 
     if (!command_prompt)
         return;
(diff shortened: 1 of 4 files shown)


src/filemanager/layout.c
```diff
@@ -984,32 +983,26 @@ setup_cmdline (void)
     {
         // Workaround: avoid crash on FreeBSD (see ticket #4213 for details)
         if (subshell_prompt != NULL)
-            tmp_prompt = g_string_free (subshell_prompt, FALSE);
-        else
-            tmp_prompt = g_strdup (mc_prompt);
-        (void) strip_ctrl_codes (tmp_prompt);
+        {
+            g_free (mc_prompt);
+            mc_prompt = g_strdup (subshell_prompt->str);
```
mc-worker: You do know the string length, why are you not using it?
```
mc_prompt = g_strndup (subshell_prompt->str, subshell_prompt->len);
```

src/main.c
```diff
@@ -419,7 +419,7 @@ main (int argc, char *argv[])
         enable_bracketed_paste ();
 
         // subshell_prompt is NULL here
-        mc_prompt = (geteuid () == 0) ? "# " : "$ ";
+        mc_prompt = g_strdup ((geteuid () == 0) ? "# " : "$ ");
```
mc-worker, line 422: There is no `g_free (mc_prompt)` later.

src/filemanager/layout.c
horkykuba: Done

src/main.c
horkykuba: Done

horkykuba:
@aborodin thanks, updated

mc-worker (approved):
Thanks!

zyv (approved):
Thank you!

PR #4822, opened 2025-10-25
Tickets #4137 & #4821: true color support with ncurses and S-Lang color detection

### LINKED ISSUE #4137: TERM "*-direct" values are not recognized as True Color-capable
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4137 |
> | **Reporter** | `RogueScholar` (admin@….net) |
> | **Keywords** | `truecolor`, `terminfo`, `color` |

<p>
Using latest versions of Konsole and Yakuake terminal emulators with True Color (24-bit or Direct RGB) color support announced via the pair of environment variables:<br />
</p>
<div class="code"><pre> <span class="nv">TERM</span><span class="o">=</span>konsole-direct
 <span class="nv">COLORTERM</span><span class="o">=</span>truecolor
</pre></div><p>
With MC configured to use either 256-color or 16M-color skins reports on invocation that the terminal is capable of only basic 16-color palettes and has reverted to the default skin accordingly. Prepending 'TERM=konsole-256color' to the invocation while leaving the COLORTERM variable set as before provides the expected behavior, with skins of all color bit-depths able to be set and subsequently displaying as designed. A test with TERM set to the more generic 'xterm-direct' value produced the same results as 'konsole-direct' where MC does not recognize direct RGB colors. This leads me to conclude that MC lacks recognition for the "-direct" values for TERM which were implemented in <a class="ext-link" href="https://invisible-island.net/ncurses/ncurses-slang.html#compare_dots">S-Lang 2.3.1 built with ncurses &gt;=6.1</a>.<br />
</p>
<h3 id="Systeminfo">System info</h3>
<ul><li>Operating System: Kubuntu 20.10
</li><li>KDE Plasma Version: 5.20.0
</li><li>KDE Frameworks Version: 5.76.0
</li><li>Qt Version: 5.15.0
</li><li>Kernel Version: 5.8.0-23-generic
</li><li>OS Type: 64-bit
</li></ul><h3 id="Hardware">Hardware</h3>
<ul><li>Model: Lenovo <tt>ThinkPad</tt> T420s
</li><li>Processor: Intel® Core™ i7-2620M CPU @ 2.70GHz
</li><li>Memory: 8 GiB of RAM
</li><li>Graphics Processor: NVS 4200M/PCIe/SSE2
</li></ul><h3 id="RelevantSoftwareVersions">Relevant Software Versions</h3>
<ul><li>libncursesw6: 6.2
</li><li>libslang2: 2.3.2 (built using unicode-data 13.0.0)
</li><li>konsole: 20.08.2
</li><li>yakuake: 20.08.1
</li></ul><h3 id="Debuggingcommandoutput">Debugging command output</h3>
<div class="code"><pre>$ <span class="nv">LC_MESSAGES</span><span class="o">=</span>C mc -V

  GNU Midnight Commander 4.8.25-55-g5bf714b
  Built with GLib 2.64.2
  Built with S-Lang 2.3.2 with terminfo database
  With <span class="nb">builtin</span> Editor and Aspell support
  With support <span class="k">for</span> background operations
  With support <span class="k">for</span> X11 events
  Virtual File Systems:
   cpiofs, tarfs, sfs, extfs, ftpfs, sftpfs, fish
  Data types:
   char: 8<span class="p">;</span> int: 32<span class="p">;</span> long: 64<span class="p">;</span> void *: 64<span class="p">;</span> size_t: 64<span class="p">;</span> off_t: 64<span class="p">;</span>

$ mc --configure-options

   <span class="s1">'--build=x86_64-linux-gnu'</span> <span class="s1">'--prefix=/usr'</span> <span class="s1">'--includedir=${prefix}/include'</span>
   <span class="s1">'--mandir=${prefix}/share/man'</span> <span class="s1">'--infodir=${prefix}/share/info'</span>
   <span class="s1">'--sysconfdir=/etc'</span> <span class="s1">'--localstatedir=/var'</span> <span class="s1">'--libexecdir=/usr/lib'</span>
   <span class="s1">'--srcdir=.'</span> <span class="s1">'--disable-maintainer-mode'</span> <span class="s1">'--disable-dependency-tracking'</span>
   <span class="s1">'--disable-silent-rules'</span> <span class="s1">'--with-x'</span> <span class="s1">'--with-screen=slang'</span> <span class="s1">'--enable-aspell'</span>
   <span class="s1">'--disable-rpath'</span> <span class="s1">'--disable-static'</span> <span class="s1">'build_alias=x86_64-linux-gnu'</span>
   <span class="s1">'CFLAGS=-g -O2 -fdebug-prefix-map=/build/mc-4.8.25~git20201017=.
   -fstack-protector-strong -Wformat -Werror=format-security'</span>
   <span class="s1">'LDFLAGS=-Wl,-Bsymbolic-functions -Wl,-z,relro'</span>
   <span class="s1">'CPPFLAGS=-Wdate-time -D_FORTIFY_SOURCE=2'</span>
</pre></div>

### LINKED ISSUE #4821: Add true color support with ncurses

Currently there is no true color support when using ncurses library.

Use true color support of ncurses library. It is supported by ncurses if terminfo entry includes `RGB` capability. This is set in `-direct`, `-direct16`, and `-direct256` terminfo variants.

_No response_

### Additional context

_No response_

horkykuba:
## Proposed changes

Enable true color support in ncurses. TERM must be set to a relevant terminfo entry, i.e. *-direct, *-direct16, or *-direct256.

If the terminal does not support 256 colors but does support true color, allow the use of 256-color skins by adding a translation layer that computes direct color values from 256-color palette. So both 256-color and true color mode can be used at the same time.

Also fix true color detection in S-Lang. S-Lang checks for RGB terminfo capability to enable true color, so we should check for it as well. This means the COLORTERM variable doesn't need to be set when using direct terminfo variants. This also applies to simultaneous 256-color and true color mode.

Update FAQ.

* Resolves: #4137
* Resolves: #4821

## Checklist

### DIFF
--- a/doc/FAQ
+++ b/doc/FAQ
@@ -403,47 +403,51 @@ Frequently Asked Questions
    If you get colors, be happy.
 
    If your terminal stays black and white, your terminal doesn't support
-   color.  You might want to upgrade to a terminal which compatible with
+   color.  You might want to upgrade to a terminal which is compatible with
    the ANSI color sequences.
 
    If your terminal goes completely black, see the next question.
 
    More detailed answer:
 
-   Check that your terminal supports color.  color_xterm, rxvt and Linux
-   console do support, most other terminals don't.  You can test color
-   support with following simple C program:
+   Check that your terminal supports color.  Most terminals do. You can
+   test color support with the following command:
 
-    #include <stdio.h>
+        printf "\033[32m Hello color world! \033[m\n"
 
-    int main (void){
-        printf ("\033[32m Hello world! \033[m\n");
-        return 0;
-    }
+   You can test 256 color support with the following command:
 
-   Compile and run it.  If you see "Hello world!" text in green your
-   terminal supports color, otherwise not (however, for color_xterm see
-   also the next question).
+        printf "\033[38;5;120m Hello 256-color world! \033[m\n"
 
-   Check whether you are using Ncurses or the S-Lang library (type
-   "mc -V" to find out).
+   You can test true color support with the following command:
 
-   With S-Lang library you can force color support by setting the
-   environment variable COLORTERM to any value.
+        printf "\033[38;2;0;200;0m Hello true color world! \033[m\n"
 
-   If you use ncurses library, check that your terminfo database
-   supports color.  If not, you should install one of the enhanced
-   terminfo databases included in GNU Midnight Commander source
-   distribution.
+   If you see the text in green, your terminal supports the respective
+   color mode.
 
-   You might want to set the TERM environment variable so that you are
-   using the correct terminfo database or termcap entry.
+   Check that you are using the proper TERM variable for your terminal.
+   If not, set it accordingly. You can use the 'toe -a' command to list
+   all available terminfo entries.
 
-   If you use color_xterm (or rxvt) the correct value might be
-   xterm-color, xtermc or simply xterm.
+   If your terminal supports 256 colors, the correct entry name may be
+   appended with -256color. If your terminal supports true color, it
+   should be appended with -direct, -direct16, or -direct256. The best
+   backwards compatibility option is -direct256 variant, which provides
+   both 256-color and true color support simultaneously. The -direct16
+   variant supports bright versions of the basic colors along with true
+   color support, in contrast to the -direct variant, which supports
+   only basic 8 colors together with true color when used in a skin.
 
-   If you use Linux console the correct value for TERM is linux or
-   console.
+   If there is no 256-color or true-color terminfo variant for your
+   terminal, even though your terminal supports it, send an e-mail to
+   ncurses/terminfo maintainers (bug-ncurses@gnu.org) and ask them to
+   add one.
+
+   With the S-Lang library (you can check by 'mc -V'), you can force
+   color support by setting the environment variable COLORTERM to any
+   value, and specifically force true color support by setting it to
+   'truecolor'.
 
 4.5 My color_xterm goes completely (or partially) black!
 
--- a/doc/FAQ
+++ b/doc/FAQ
@@ -432,12 +432,12 @@ Frequently Asked Questions
 
    If your terminal supports 256 colors, the correct entry name may be
    appended with -256color. If your terminal supports true color, it
-   should be appended with -direct, -direct16, or -direct256. The best
-   backwards compatibility option is -direct256 variant, which provides
-   both 256-color and true color support simultaneously. The -direct16
-   variant supports bright versions of the basic colors along with true
-   color support, in contrast to the -direct variant, which supports
-   only basic 8 colors together with true color when used in a skin.
+   should be appended with -direct, -direct16, or -direct256. All three
+   variants provide both 256-color and true color support simultaneously.
+   The latter two variants make no difference in mc, and they support
+   basic colors including their bright versions along with true color
+   support, in contrast to the -direct variant, which supports only basic
+   8 colors together with true color when used in a skin.
 
    If there is no 256-color or true-color terminfo variant for your
    terminal, even though your terminal supports it, send an e-mail to
--- a/lib/tty/color-internal.c
+++ b/lib/tty/color-internal.c
@@ -155,7 +155,7 @@ parse_256_or_true_color_name (const char *color_name)
                 i = (h[0] << 20) | (h[0] << 16) | (h[1] << 12) | (h[1] << 8) | (h[2] << 4) | h[2];
             else
                 i = (h[0] << 20) | (h[1] << 16) | (h[2] << 12) | (h[3] << 8) | (h[4] << 4) | h[5];
-            return (1 << 24) | i;
+            return FLAG_TRUECOLOR | i;
         }
     }
 
@@ -178,7 +178,7 @@ tty_color_get_name_by_index (int idx)
             return color_table[i].name;
 
     // Create and return the strings in "colorNNN" or "#rrggbb" format.
-    if ((idx >= 16 && idx < 256) || (idx & (1 << 24)) != 0)
+    if ((idx >= 16 && idx < 256) || (idx & FLAG_TRUECOLOR) != 0)
     {
         char name[9];
 
--- a/lib/tty/color-internal.c
+++ b/lib/tty/color-internal.c
@@ -241,3 +241,37 @@ tty_attr_get_bits (const char *attrs)
 }
 
 /* --------------------------------------------------------------------------------------------- */
+
+int
+convert_256color_to_truecolor (int color)
+{
+    int r, g, b;
+
+    // Invalid color
+    if (color > 255)
+        return 0;
+
+    if (color >= 232)  // Gray scale
+        r = g = b = (color - 231) * 10 + 8;
+    else if (color >= 16)  // 6x6x6 color cube
+    {
+        color -= 16;
+
+        r = (color / (6 * 6) % 6);
+        r = r > 0 ? r * 40 + 55 : 0;
+
+        g = (color / 6 % 6);
+        g = g > 0 ? g * 40 + 55 : 0;
+
+        b = (color % 6);
+        b = b > 0 ? b * 40 + 55 : 0;
+    }
+    else  // We don't convert basic 16 colors as they are terminal-dependent and user-configurable
+        return color;
+
+    color = FLAG_TRUECOLOR | (r << 16) | (g << 8) | b;
+
+    return color;
+}
+
+/* --------------------------------------------------------------------------------------------- */
--- a/lib/tty/color-ncurses.c
+++ b/lib/tty/color-ncurses.c
@@ -38,6 +38,7 @@
 
 #include "lib/global.h"
 
+#include "tty.h"
 #include "tty-ncurses.h"
 #include "color.h"  // variables
 #include "color-internal.h"
@@ -53,6 +54,7 @@
 /*** file scope variables ************************************************************************/
 
 static GHashTable *mc_tty_color_color_pair_attrs = NULL;
+static int overlay_colors = 0;
 
 /* --------------------------------------------------------------------------------------------- */
 /*** file scope functions ************************************************************************/
@@ -179,8 +181,9 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
         ibg = mc_color_pair->bg;
         attr = mc_color_pair->attr;
 
-        // In legacy color mode, change bright colors into bold
-        if (!tty_use_256colors (NULL) && !tty_use_truecolors (NULL))
+        // If we have 8 indexed colors only, change foreground bright colors into bold and
+        // background bright colors to basic colors
+        if (COLORS <= 8 || (tty_use_truecolors (NULL) && overlay_colors <= 8))
         {
             if (ifg >= 8 && ifg < 16)
             {
@@ -191,11 +194,31 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
             if (ibg >= 8 && ibg < 16)
             {
                 ibg &= 0x07;
-                // attr | = A_BOLD | A_REVERSE ;
             }
         }
 
+        // Shady trick: if we don't have the exact color, because it is overlaid by backwards
+        // compatibility indexed values, just borrow one degree of red. The user won't notice :)
+        if ((ifg & FLAG_TRUECOLOR) != 0)
+        {
+            ifg &= ~FLAG_TRUECOLOR;
+            if (ifg != 0 && ifg <= overlay_colors)
+                ifg += (1 << 16);
+        }
+
+        if ((ibg & FLAG_TRUECOLOR) != 0)
+        {
+            ibg &= ~FLAG_TRUECOLOR;
+            if (ibg != 0 && ibg <= overlay_colors)
+                ibg += (1 << 16);
+        }
+
+#if NCURSES_VERSION_PATCH >= 20170401 && defined(NCURSES_EXT_COLORS) && defined(NCURSES_EXT_FUNCS) \
+    && defined(HAVE_NCURSES_WIDECHAR)
+        init_extended_pair (mc_color_pair->pair_index, ifg, ibg);
+#else
         init_pair (mc_color_pair->pair_index, ifg, ibg);
+#endif
         mc_tty_color_save_attr (mc_color_pair->pair_index, attr);
     }
 }
@@ -231,7 +254,19 @@ tty_use_256colors (GError **error)
 {
     (void) error;
 
-    return (COLORS == 256);
+    overlay_colors = tty_tigetnum ("CO", NULL);
+
+    if (COLORS != 256 && !(COLORS > 256 && overlay_colors == 256))
+    {
+        g_set_error (error, MC_ERROR, -1,
+                     _ ("\nIf your terminal supports 256 colors, you need to set your TERM\n"
+                        "environment variable to match your terminal, perhaps using\n"
+                        "a *-256color or *-direct256 variant. Use the 'toe -a'\n"
+                        "command to list all available variants on your system.\n"));
+        return FALSE;
+    }
+
+    return TRUE;
 }
 
 /* --------------------------------------------------------------------------------------------- */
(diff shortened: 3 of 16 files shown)


egmontkob:
Just checked what `xterm-direct256` does, and accordingly, made a new comment to #4137.

I see the

    Shady trick [...] just borrow one degree of red. The user won't notice :)

bit in your change. I guess this is the best we can do, given the underlying library's limitations.

doc/FAQ
```diff
@@ -403,47 +403,51 @@ Frequently Asked Questions
...
-   also the next question).
+        printf "\033[38;5;120m Hello 256-color world! \033[m\n"
 
-   Check whether you are using Ncurses or the S-Lang library (type
-   "mc -V" to find out).
+   You can test true color support with the following command:
 
-   With S-Lang library you can force color support by setting the
-   environment variable COLORTERM to any value.
+        printf "\033[38;2;0;200;0m Hello true color world! \033[m\n"
```
egmontkob, line 424: Just FYI:

The official (ITU-T T.416) escape sequences would be `\033[38:5:120m` and `\033[38:2::0:200:0m`.

The official separator character is `:`, and there's an (unused) color-space-id (if I remember correctly) parameter between `2` and the red component.

The widespread but unofficial use of `;` as separator, along with the omission of color-space-id, is more widely supported. There are quite a few terminals that support these but not the official forms, while I'm not aware of any that would do it the other way around. So, for the purposes of the user testing the terminal (and assuming that ncurses/slang would also use the more widespread but unofficial codes), it makes sense to use them here, but it's not de jure correct.

lib/tty/color-slang.c
```diff
@@ -245,11 +253,15 @@ tty_use_truecolors (GError **error)
...
-    if (colorterm == NULL
-        || (strcmp (colorterm, "truecolor") != 0 && strcmp (colorterm, "24bit") != 0))
+    if (!((tty_tgetflag ("RGB") && tty_tgetnum ("Co") == 16777216)
+          || (colorterm != NULL
+              && (strcmp (colorterm, "truecolor") == 0 || strcmp (colorterm, "24bit") == 0))))
     {
         g_set_error (error, MC_ERROR, -1,
-                     _ ("Set COLORTERM=truecolor if your terminal really supports true colors."));
+                     _ ("\nIf your terminal supports true colors, you need to set your TERM\n"
+                        "environment variable to a *-direct, *-direct16, or *-direct256 variant.\n"
```
egmontkob: I think they should be listed in decreasing order of preference, i.e. 256 first, then 16, then simple (8) direct. (Same for ncurses's similar message.)

lib/tty/color-ncurses.c
```diff
@@ -231,17 +246,36 @@ tty_use_256colors (GError **error)
...
 {
-    // Not yet supported in ncurses
-    g_set_error (error, MC_ERROR, -1, _ ("True color not supported with ncurses."));
-    return FALSE;
+    if (COLORS != 16777216)
+    {
+        g_set_error (error, MC_ERROR, -1,
+                     _ ("\nIf your terminal supports true colors, you need to set your TERM\n"
+                        "environment variable to a *-direct, *-direct16, or *-direct256 variant.\n"
+                        "Use the 'toe' command to list all available variants on your system.\n"));
```
egmontkob: I think you could say `'toe -a'` to save the user a manpage lookup.

egmontkob (commented):
I'm not going to do a detailed review. As for the big picture, I think it's as good as it gets (given ncurses's limitations), and it survived nicely some stress-testing for me (like using a 256-color skin with direct16, rejected with an error message as expected).

Thanks a lot!

horkykuba:
> Just checked what `xterm-direct256` does, and accordingly, made a new comment to #4137.
> 
> I see the
> 
> ```
> Shady trick [...] just borrow one degree of red. The user won't notice :)
> ```
> 
> bit in your change. I guess this is the best we can do, given the underlying library's limitations.

Yes, unfortunately. I’m trying to figure out better solution with @ThomasDickey, but the discussion is still not completed - see:

https://lists.gnu.org/archive/html/bug-ncurses/2025-09/msg00032.html

(the rest of this thread is there: https://lists.gnu.org/archive/html/bug-ncurses/2025-08/msg00101.html)

zyv:
I wonder if the build failure on Solaris is again due to the `libmc` shared object nonsense (see #4806).

~~Unfortunately~~, there is so much activity around the repository now that I totally fell behind on my mail, let alone was able to do something other than skimming through the messages and answering when possible without much thinking...

horkykuba:
> I wonder if the build failure on Solaris is again due to the `libmc` shared object nonsense (see #4806).
> 
> ~Unfortunately~, there is so much activity around the repository now that I totally fell behind on my mail, let alone was able to do something other than skimming through the messages and answering when possible without much thinking...

I completely understand :) That’s because ncurses doesn’t recognize init_extended_pair(), which means it’s using a version of ncurses older than 6.1 (from 2018).

Maybe #ifdef that?

doc/FAQ
horkykuba: Yes, this is exact reason I wrote the commands in this form. Many terminals still support only ; form, for example PuTTY. Unfortunately, the ; form is ambiguous to concatenating more SGR commands, which is the standard.

lib/tty/color-ncurses.c
```diff
@@ -179,23 +181,36 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
...
+        // Shady trick: if we don't have the exact color, because it is overlaid by backwards
+        // compatibility indexed values, just borrow one degree of red. The user won't notice :)
+        if (ifg & (1 << 24))
+        {
+            ifg &= ~(1 << 24);
+            if (ifg != 0 && ifg <= overlay_colors)
+                ifg += (1 << 16);
+        }
+
+        if (ibg & (1 << 24))
```
mc-worker: Same.

lib/tty/color-ncurses.c
```diff
@@ -231,17 +246,36 @@ tty_use_256colors (GError **error)
...
 
 /* --------------------------------------------------------------------------------------------- */
 
 gboolean
 tty_use_truecolors (GError **error)
 {
-    // Not yet supported in ncurses
-    g_set_error (error, MC_ERROR, -1, _ ("True color not supported with ncurses."));
-    return FALSE;
+    if (COLORS != 16777216)
```
mc-worker: It would be great to create a constant or preprocessor macro with comment for this magic number.

lib/tty/color-ncurses.c
```diff
@@ -179,23 +181,36 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
...
-        init_pair (mc_color_pair->pair_index, ifg, ibg);
+        if (ibg >= 8 && ibg < 16)
+        {
+            ibg &= 0x07;
+            // attr | = A_BOLD | A_REVERSE ;
+        }
+
+        // Shady trick: if we don't have the exact color, because it is overlaid by backwards
+        // compatibility indexed values, just borrow one degree of red. The user won't notice :)
+        if (ifg & (1 << 24))
```
mc-worker: Please use explicit tests:
```if ((ifg & (1 << 24)) != 0)```

File: lib/tty/color-slang.c (hunk 2 above)
Comment (horkykuba) [reply lib/tty/color-slang.c]: Done

File: lib/tty/color-ncurses.c (hunk 3 above)
Comment (horkykuba) [reply lib/tty/color-ncurses.c]: Done

File: lib/tty/color-ncurses.c (hunk 5 above)
Comment (horkykuba) [reply lib/tty/color-ncurses.c]: Done

File: lib/tty/color-ncurses.c (hunk 6 above)
Comment (horkykuba) [reply lib/tty/color-ncurses.c]: Done

File: lib/tty/color-ncurses.c (hunk 4 above)
Comment (horkykuba) [reply lib/tty/color-ncurses.c]: Done

**Comment by horkykuba** [comment]
> Maybe #ifdef that?

Done

**Comment by horkykuba** [comment]
It appears that Solaris build still fails because although it uses recent ncurses, but it is configured to use ABI 5, which is incompatible with init_extended_pair(). Included into #ifdef.

**Comment by horkykuba** [comment]
Fixed compatibility code in 8-color mode, improved documentation, included commit with minor improvement of true color skins (differentiate directories and files)

**Comment by horkykuba** [comment]
Fixed #ifdefs for checking availability of ncurses extended colors

**Comment by horkykuba** [comment]
Added a commit fixing a bug that caused shadows not to work with ncurses

**Comment by zyv** [comment]
> Added a commit fixing a bug that caused shadows not to work with ncurses

See #4817 for @egmontkob's notes on the conditional. I'm sorry, but I'm confused about you solving this problem by just "dropping" support for non-wide ncurses.

Could you please explain whether it was even "supported" / "working" before for some definition of "working"? If it wasn't working or isn't possible to support it, then dropping it hard might be the reasonable thing to do. But if not, I would actually prefer keeping it working with shadows disabled by changing the conditional appropriately.

The background is that according to my knowledge, most of the distributions do use S-Lang. However, most "exotic" systems like AIX either come with their own curses or ship some very old version of ncurses like Solaris. S-Lang is either not available or undesirable because it needs to be built in the first place and pulls in further dependencies.

This is why I'm happy to have curses/ncurses support working at least to some degree, with however degraded functionality it might be, and I was even questioning if we want to support S-Lang in the long run as the second library. The alternative consideration was to drop both ncurses and S-Lang and roll our own (see #3264), but I personally neither have the competence nor capacity to write and maintain a screen library just for mc.

**Comment by horkykuba** [comment]
> > Added a commit fixing a bug that caused shadows not to work with ncurses
> 
> See #4817 for @egmontkob's notes on the conditional. I'm sorry, but I'm confused about you solving this problem by just "dropping" support for non-wide ncurses.
> 
> Could you please explain whether it was even "supported" / "working" before for some definition of "working"? If it wasn't working or isn't possible to support it, then dropping it hard might be the reasonable thing to do. But if not, I would actually prefer keeping it working with shadows disabled by changing the conditional appropriately.

See my commit message, where I explain it in more detail:

> Fix a bug in preprocessor statement that caused shadows to be always
> disabled in ncurses, instead of intended behavior of being disabled only
> when the non-wide ncurses library was used.
> 
> If only the non-wide ncurses library is found, throw an error during
> ./configure instead of warning, since mc is not compatible with the
> non-wide version.
> 
> Remove the ENABLE_SHADOWS preprocessor macro as it is redundant.

The problem is that the current mc code depends heavily on the wide ncurses functions, so effectively mc doesn't support the non-wide version of ncurses. This is why I'm dropping the conditional - there is currently no way to compile the code with non-wide ncurses without major code changes.

**Comment by egmontkob** [comment]
I tried to see if mc works with non-wide ncurses.

tl;dr: Yes it does.

Longer version: It's not a separate package on Ubuntu 25.10, so I manually moved away the `/usr/include/ncursesw` directory and the `/usr/lib/x86_64-linux-gnu/libncurses*w*` files, as well as `/usr/lib/x86_64-linux-gnu/pkgconfig/ncurses*w.pc`.

~For some reason yet to be tracked down, `./configure --with-screen=ncurses` believes that I do have ncursesw, and places `#define HAVE_NCURSESW_NCURSES_H 1` and `#define HAVE_NCURSESW_TERM_H 1`, causing the build to fail on trying to `#include <ncursesw/ncurses.h>`. So I manually fixed `config.h`, commented out those two faulty defines. (Note: setting to 0 doesn't work.)~ [Update: #4825 this was a ccache bug]

~Apart from this little manual intervention,~ it compiles fine.

This is recent master at e32631a6cae83dfb2bb8fe0b821b0d21adf318d9, with the first (faulty) attempt to fix NCURSES_WIDECHAR detection. At my pending 3158 / 4813 line drawing work it fails, probably due to an older (even more faulty) variant of this check, but I'll make sure to fix it.

I am not 100% satisfied with the idea of dropping narrow ncurses support due to the introduction of truecolor support. Mind you, my pending line drawing would also become _slightly_ simpler if we dropped it. However, as long as these ifdefs are reasonably maintainable, I'd prefer if truecolor could be ifdef'ed to depend on widechar, while letting narrowchar still work with reduced (256-color) functionality. I mean, I won't shed a tear if mc drops it in order to become more easily maintainable, but it should be a well-thought-out decision; claiming that it doesn't work anyway (apart from a small autodetection issue) is not true.

(Side note: widechar ncurses has apparently been available since 2000. Yet, it's only built by default on a simple `./configure && make` since 6.4-20231021.)

**Comment by zyv** [comment]
Your last commit reverts #4800. Check it out first. Also see the continued discussion in #4817.

**Comment by horkykuba** [comment]
> I tried to see if mc works with non-wide ncurses.
> 
> tl;dr: Yes it does.

My bad. I accidentally checked mc's dependency on wide ncurses with your double-line patch merged, and it showed about 8 wide-dependent functions :) @zyv sorry, you were right.

mc can be indeed compiled with non-wide ncurses. So I reverted the conditional enabling of shadows and fixed it properly instead. To enable this, I added an autoconf check for wide/non-wide ncurses in a separate commit.

**Comment by horkykuba** [comment]
> Your last commit reverts #4800. Check it out first. Also see the continued discussion in #4817.

Not really, it changes the conditional to autoconf generated macro.

**Comment by egmontkob** [comment]
@horkykuba You rock! (I've verified your widechar detection.)

@zyv You overlooked this, there's another commit underneath similar to the one you suggested, combined with my suggestion to use a more verbose name.

@zyv Could we please land these two commits in master ASAP?
- m4.include: Add autoconf check for wide/non-wide ncurses
- Fix shadows not working with ncurses

My line drawing work also depends on that; that way I could go on in parallel to the beefier truecolor review.

**Comment by zyv** [comment]
> > Your last commit reverts #4800. Check it out first. Also see the continued discussion in #4817.
> 
> Not really, it changes the conditional to autoconf generated macro.

I'm sorry, I'm getting really confused and overwhelmed by the email flow :-/

> @zyv Could we please land these two commits in master ASAP?
> 
> * m4.include: Add autoconf check for wide/non-wide ncurses
> * Fix shadows not working with ncurses
> 
> My line drawing work also depends on that; that way I could go on in parallel to the beefier truecolor review.

If you guys could please extract this change as a separate PR, I will merge it immediately. Then you can rebase your other WIP stuff on top of that...

**Comment by horkykuba** [comment]
> If you guys could please extract this change as a separate PR, I will merge it immediately. Then you can rebase your other WIP stuff on top of that...

Done, #4826.

**Comment by horkykuba** [comment]
Added commit with compatibility translation from 256 colors to true colors if the terminal supports true colors but doesn't support 256 colors.

File: lib/tty/color-internal.c [hunk 7]
```diff
@@ -241,3 +241,36 @@ tty_attr_get_bits (const char *attrs)
...
+
+int
+convert_256color_to_truecolor (int color)
+{
+    int r, g, b;
+
+    // Invalid color
+    if (color > 255)
+        return 0;
+    else if (color >= 232)  // Gray scale
```
ossilator: 'else' after jump statements is pointless, and therefore discouraged by some coding styles (https://midnight-commander.org/coding-style/ doesn't have anything to say about it).

lib/tty/color-ncurses.c
```diff
@@ -231,17 +258,57 @@ tty_use_256colors (GError **error)
 {
     (void) error;
 
-    return (COLORS == 256);
+    overlay_colors = tty_tigetnum ("CO", NULL);
+
+    if (COLORS == 256 || (COLORS > 256 && overlay_colors == 256))
+        return TRUE;
+    else if (tty_use_truecolors (NULL))
```
ossilator: ditto

lib/tty/color-ncurses.c
```diff
@@ -231,17 +258,57 @@ tty_use_256colors (GError **error)
...
+    {
+        need_convert_256color = TRUE;
+        return TRUE;
+    }
+    else
```
ossilator: ...

lib/tty/color-slang.c
```diff
@@ -214,15 +222,28 @@ tty_set_normal_attrs (void)
...
+    int colors, overlay_colors;
 
-    ret = (SLtt_Use_Ansi_Colors && SLtt_tgetnum ((char *) "Co") == 256);
+    colors = tty_tigetnum ("colors", "Co");
+    overlay_colors = tty_tigetnum ("CO", NULL);
 
-    if (!ret)
+    if (SLtt_Use_Ansi_Colors && (colors == 256 || (colors > 256 && overlay_colors == 256)))
+        return TRUE;
+    else if (tty_use_truecolors (NULL))
```
ossilator: ...

lib/tty/color-slang.c
```diff
@@ -214,15 +222,28 @@ tty_set_normal_attrs (void)
...
+    {
+        need_convert_256color = TRUE;
+        return TRUE;
+    }
+    else
```
ossilator: ...

ossilator (commented):
i just skimmed it (and don't intend to go further).

horkykuba:
> > Just checked what `xterm-direct256` does, and accordingly, made a new comment to #4137.
> > I see the
> > ```
> > Shady trick [...] just borrow one degree of red. The user won't notice :)
> > ```
> > bit in your change. I guess this is the best we can do, given the underlying library's limitations.
> 
> Yes, unfortunately. I’m trying to figure out better solution with @ThomasDickey, but the discussion is still not completed - see:
> 
> https://lists.gnu.org/archive/html/bug-ncurses/2025-09/msg00032.html
> 
> (the rest of this thread is there: https://lists.gnu.org/archive/html/bug-ncurses/2025-08/msg00101.html)

So, we have confirmed the new capability names for separate RGB, intended to be included in terminfo and ncurses by @ThomasDickey in the future:

https://lists.gnu.org/archive/html/bug-ncurses/2025-10/msg00086.html

It follows today's Emacs implementation.

lib/tty/color-internal.c
horkykuba: Done

lib/tty/color-ncurses.c
horkykuba: Done

lib/tty/color-slang.c
horkykuba: Done

horkykuba:
Made changes suggested by @ossilator and improved documentation in comments

zyv (commented):
What do you think @mc-worker & @egmontkob ? I'm not a color-user, but code-wise I don't think I have anything to request, so I'm wondering if we can merge it now.

egmontkob:
I only skimmed over the first iteration the code, on the high level it looked about as reasonable (incl. workarounds for ncurses's limitations) as it could be, also survived a bit of testing from me.

Also, I know I shouldn't go personal, but based on his debut contributions I have a pretty good trust in horkykuba's work.

Yup let's land it and see how it goes.

horkykuba:
Thank you guys.

Fyi, I also plan to add support for `setb24/setf24`, which @ThomasDickey has confirmed as future official capabilities for ncurses/terminfo:

> So, we have confirmed the new capability names for separate RGB, intended to be included in terminfo and ncurses by @ThomasDickey in the future:
> 
> https://lists.gnu.org/archive/html/bug-ncurses/2025-10/msg00086.html
> 
> It follows today's Emacs implementation.

However, I'll do that in a separate PR.

zyv:
@horkykuba, thank you for your contribution!

PR #4826, opened 2025-10-27
Fix shadows not working with ncurses

### LINKED ISSUE #4817: Fixing build with ncurses-5.9 (#4800) broke shadows

```shell
a5909f24b800fbac51a7f8049eddaf9a960ebee1
```

### Operating system

```shell
all
```

### How to reproduce

No shadows on ncurses-6.5 after the said commit / PR.

Ossi correctly pointed out during review that the condition is fishy; it is indeed broken.

I assume that in ncurses itself, where this was taken from, something else is going on. Widec support might already be enabled and he's just gathering a few more conditions when it also needs to be enabled; or something like that.

(Also, in my pending PR 4813, line characters break if I apply this same change.)

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

horkykuba:
## Proposed changes

Fix shadows not working with ncurses

Fix a bug in preprocessor statement that caused shadows to be always disabled in ncurses, instead of intended behavior of being disabled only when the non-wide ncurses library was used.

To be able to do this, also add autoconf check for wide/non-wide ncurses.

* Resolves: #4817

### DIFF
--- a/lib/tty/tty-ncurses.h
+++ b/lib/tty/tty-ncurses.h
@@ -27,7 +27,7 @@
 #endif
 
 /* do not draw shadows if NCurses is built with --disable-widec */
-#if defined(NCURSES_WIDECHAR) && (NCURSES_WIDECHAR - 0 < 1)
+#ifdef HAVE_NCURSES_WIDECHAR
 #define ENABLE_SHADOWS 1
 #endif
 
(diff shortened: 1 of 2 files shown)


zyv (approved):
🙏

PR #4831, opened 2025-10-28
Ticket #3869: actually use the .netrc password for ftp vfs

### LINKED ISSUE #3869: ftp: fail to retrieve password from .netrc when user is in URL
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3869 |
> | **Reporter** | `cri` (cri@….org) |

<p>
I have a proper entry in <tt>~/.netrc</tt> for USERNAME, PASSWORD and HOST.<br />
</p>
<p>
If I try to access the URL <a class="ext-link" href="ftp://HOST/PATH">ftp://HOST/PATH</a>, mc correctly retrieves USERNAME and PASSWORD from <tt>.netrc</tt> and I can log into the FTP host.<br />
</p>
<p>
If I try to access the URL <a class="ext-link" href="ftp://USERNAME@HOST/PATH">ftp://USERNAME@HOST/PATH</a>, mc displays a popup window, asking for the password. Shouldn't it retrieve the password from <tt>.netrc</tt>?<br />
</p>
<p>
I had a look at <a href="https://midnight-commander.org/browser/src/vfs/ftpfs/ftpfs.c#L371">https://midnight-commander.org/browser/src/vfs/ftpfs/ftpfs.c#L371</a><br />
</p>
<p>
shouldn't it be <br />
</p>

```
...`&& path_element->password == NULL`
```
<p>
instead of<br />
</p>

```
...`&& path_element->password != NULL`
```
<p>
?<br />
</p>
<p>
Or am I missing something?<br />
</p>

Vrihub:
## Proposed changes

This fixes #3869, when one wants to log into an ftp server only providing username and hostname, and relies on mc to retrieve the password from the .netrc file. Without this, mc retrieves the password from the .netrc file, but doesn't actually use it to login, and asks the user for a password instead.

Please review, this works for me, but I'm not 100% sure it's the best solution.

* Resolves: #3869

## Checklist

### DIFF
--- a/src/vfs/ftpfs/ftpfs.c
+++ b/src/vfs/ftpfs/ftpfs.c
@@ -371,12 +371,13 @@ ftpfs_correct_url_parameters (const vfs_path_element_t *velement)
 
         ftpfs_netrc_lookup (path_element->host, &new_user, &new_passwd);
 
-        // If user is different, remove password
-        if (new_user != NULL && strcmp (path_element->user, new_user) != 0)
-            MC_PTR_FREE (path_element->password);
+        // Use the password if it matches the user
+        if (new_user != NULL && strcmp (path_element->user, new_user) == 0)
+            path_element->password = new_passwd;
+        else
+            g_free (new_passwd);
 
         g_free (new_user);
-        g_free (new_passwd);
     }
 
     return path_element;


src/vfs/ftpfs/ftpfs.c
```diff
@@ -374,6 +374,9 @@ ftpfs_correct_url_parameters (const vfs_path_element_t *velement)
         // If user is different, remove password
         if (new_user != NULL && strcmp (path_element->user, new_user) != 0)
             MC_PTR_FREE (path_element->password);
+        
+        if (path_element->password == NULL)
+          path_element->password = g_strdup (new_passwd);
```
mc-worker: There is no reason to duplicate `new_passwd` that is just free'd below.

zyv: Well, I think this is the safest and easiest way to code that, isn't it? What would be your way of doing it?

I'm more concerned about this introducing a logical mistake. I think this if-block should go above the previous one. Otherwise, it just reverts the effect of the previous condition, but not only for the same user but also for a different user.

Vrihub: You're right, there's a logical error, but I think the correct way is using `else`, see my latest fixup commit. I've tested the following cases and mc behaves as expected:
- `ftp://host`: uses user and passwd from .netrc
- `ftp://user@host`: uses `user`'s passwd from .netrc 
- `ftp://user_not_in_netrc@host`: asks for `user_not_in_netrc`'s password

Please just confirm that it's ok to use `g_strdup()`, I'm not sure about that, I've used it in analogy with line 364 `path_element->user = g_strdup ("anonymous");`

Vrihub: > Please just confirm that it's ok to use `g_strdup()`, I'm not sure about that, I've used it in analogy with line 364 `path_element->user = g_strdup ("anonymous");`

Well, if I just do `path_element->password = new_passwd;`, the ftp login fails, and mc segfaults.

mc-worker: > Well, if I just do `path_element->password = new_passwd;`, the ftp login fails, and mc segfaults.

Because of use-after-free. You can take `new_passwd` in the one branch and free it in the other one.

src/vfs/ftpfs/ftpfs.c
```diff
@@ -374,6 +374,9 @@ ftpfs_correct_url_parameters (const vfs_path_element_t *velement)
         // If user is different, remove password
         if (new_user != NULL && strcmp (path_element->user, new_user) != 0)
             MC_PTR_FREE (path_element->password);
```
mc-worker: Well, this is pointless because `path_element->password` is NULL (see line 367).

src/vfs/ftpfs/ftpfs.c
```diff
@@ -374,6 +374,9 @@ ftpfs_correct_url_parameters (const vfs_path_element_t *velement)
...
+        else
+            // otherwise, use it
+            path_element->password = g_strdup (new_passwd);
```
zyv: So in the end, the patch becomes like that, agree?

```suggestion
        // Use the password if it matches the user
        if (new_user != NULL && strcmp (path_element->user, new_user) == 0)
            path_element->password = g_strdup (new_passwd);
```

src/vfs/ftpfs/ftpfs.c
zyv: I actually think that branching free is less safe than duplicating the string. The performance penalty of a few microseconds is negligible, because it isn't in a loop, but if later more code is added, whether to free or not has to be correctly considered in all branches. As we can see, we have a lot of bugs creeping in due to subtle errors in conditions. Using `g_strdup` always, even if it's not really needed, is more like ownership transfer semantics in C++. So it's harder to make a mistake when the code is modified later.

src/vfs/ftpfs/ftpfs.c
Vrihub: I confirm this version passes the three test cases I described in https://github.com/MidnightCommander/mc/pull/4831#discussion_r2470776424

Re the other implementation details (string duplication etc), I'll trust you, because my C knowledge is quite limited.

zyv (approved):

zyv:
@mc-worker ok?

src/vfs/ftpfs/ftpfs.c
mc-worker: > I actually think that branching free is less safe than duplicating the string. The performance penalty of a few microseconds is negligible, because it isn't in a loop, but if later more code is added, whether to free or not has to be correctly considered in all branches. As we can see, we have a lot of bugs creeping in due to subtle errors in conditions. Using `g_strdup` always, even if it's not really needed, is more like ownership transfer semantics in C++. So it's harder to make a mistake when the code is modified later.

I disagree.

zyv: So how do you want to have it? Like in the first or second fixup?

mc-worker: The first fixup is definitely wrong. It has a memory leak. The second one is OK .

zyv: Alright, if that's what you wanted, let's do it like this, even though I think that this is more dangerous for the reason above - C doesn't have a borrow checker.

PR #4857, opened 2025-11-03
Tickets #4847 / #4848: cons.saver cleanups

### LINKED ISSUE #4847: cons.saver setgid paranoia: Fail if !HAVE_STRUCT_STAT_ST_RDEV

```shell
4.8.33, git
```

### Operating system

```shell
Linux
```

### How to reproduce

In setgid `cons.saver`, if `! HAVE_STRUCT_STAT_ST_RDEV` then we blindly assume minor number 1 with a "FIXME" in the source.

This code is Linux-specific, and Linux does have `st_rdev`.

In the unlikely case that it's not available, let's fail (either buildtime or runtime) instead of making faulty assumptions.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

### LINKED ISSUE #4848: cons.saver setgid paranoia: Don't open arbitrary file

```shell
4.8.33, git
```

### Operating system

```shell
Linux
```

### How to reproduce

Setgid `cons.saver` checks that the passed device name is at most 15 bytes and begins with `/dev/`, and then proceeds and opens it for reading.

This allows one to create a symlink `/dev/../tmp/x` pointing wherever, and get `cons.saver` open it for reading.

Of course `cons.saver` will immediately check if it's indeed a terminal device and quit otherwise. Still, I'm not happy this amount of slack in a setgid tool.

Let's check that the passed string strictly matches the pattern `/dev/tty[0-9]+`.

Or rather: Maybe just pass the plain number as parameter, and let `cons.saver` prepend `/dev/tty` and `/dev/vcsa`.

While at it, we could remove the devfs-specific code, devfs has been obsoleted and replaced by udev in Linux 2.6 or so.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
## Proposed changes

Check the passed filename more rigorously before opening it.
    
Verify the major and minor number of the ttyN and vcsaN devices.
    
Don't make an arbitrary assumption about the device number if st_rdev is unavailable.
    
Drop support for devfs device names.

Use major() and minor() macros rather than hardcoded masks.

* Resolves: #4848
* Resolves: #4847

## Checklist

### DIFF
--- a/src/consaver/cons.saver.c
+++ b/src/consaver/cons.saver.c
@@ -69,7 +69,7 @@
 #endif
 #include <termios.h>
 
-#include "lib/unixcompat.h"  // STDERR_FILENO
+#include "lib/unixcompat.h"  // STDERR_FILENO, major(), minor()
 
 #define LINUX_CONS_SAVER_C
 #include "cons.saver.h"
@@ -202,9 +202,9 @@ main (int argc, char **argv)
     if (fstat (console_fd, &st) < 0 || !S_ISCHR (st.st_mode))
         die ();
 #ifdef HAVE_STRUCT_STAT_ST_RDEV
-    if ((st.st_rdev & 0xff00) != 0x0400)
+    if (major (st.st_rdev) != 4)
         die ();
-    if ((st.st_rdev & 0x00ff) != console_minor)
+    if (minor (st.st_rdev) != console_minor)
         die ();
 #endif
     if (st.st_uid != uid)
@@ -220,9 +220,9 @@ main (int argc, char **argv)
     if (fstat (vcsa_fd, &st) < 0 || !S_ISCHR (st.st_mode))
         die ();
 #ifdef HAVE_STRUCT_STAT_ST_RDEV
-    if ((st.st_rdev & 0xff00) != 0x0700)
+    if (major (st.st_rdev) != 7)
         die ();
-    if ((st.st_rdev & 0x00ff) != 128 + console_minor)
+    if (minor (st.st_rdev) != 128 + console_minor)
         die ();
 #endif
 
--- a/src/consaver/cons.saver.c
+++ b/src/consaver/cons.saver.c
@@ -163,11 +163,11 @@ int
 main (int argc, char **argv)
 {
     unsigned char action = 0, console_flag = 3;
-    int console_fd, vcsa_fd, console_minor, buffer_size;
+    int console_fd, vcsa_fd, buffer_size;
+    unsigned int console_minor;
     struct stat st;
     uid_t uid, euid;
     char *buffer, *tty_name, console_name[16], vcsa_name[16];
-    const char *p, *q;
     struct winsize winsz;
 
     close (STDERR_FILENO);
@@ -176,7 +176,17 @@ main (int argc, char **argv)
         die ();
 
     tty_name = argv[1];
-    if (strnlen (tty_name, 15) == 15 || strncmp (tty_name, "/dev/", 5))
+
+    // make sure the parameter is /dev/tty1 .. /dev/tty63, extract the number
+    if (sscanf (tty_name, "/dev/tty%u", &console_minor) != 1)
+        die ();
+    if (console_minor < 1 || console_minor > 63)
+        die ();
+
+    // sscanf allows negative sign, leading zeros, integer overflow, or arbitrary suffix;
+    // verify that the canonical form was given
+    snprintf (console_name, sizeof (console_name), "/dev/tty%u", console_minor);
+    if (strncmp (console_name, tty_name, sizeof (console_name)) != 0)
         die ();
 
     setsid ();
@@ -185,7 +195,8 @@ main (int argc, char **argv)
 
     if (seteuid (uid) < 0)
         die ();
-    console_fd = open (tty_name, O_RDONLY);
+
+    console_fd = open (console_name, O_RDONLY);
     if (console_fd < 0)
         die ();
     if (fstat (console_fd, &st) < 0 || !S_ISCHR (st.st_mode))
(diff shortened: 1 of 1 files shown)


egmontkob:
Could you guys please extremely thoroughly, bit-by-bit review this change? It's a setgid binary, so it deserves extra scrutiny; try to think with the mind of an attacker :)

src/consaver/cons.saver.c
```diff
@@ -67,6 +67,12 @@
 #ifdef HAVE_FCNTL_H
 #include <fcntl.h>
 #endif
+#ifdef MAJOR_IN_MKDEV
```
mc-worker: We already have this in `lib/unixcompat.h` included below.

src/consaver/cons.saver.c
```diff
@@ -176,7 +182,24 @@ main (int argc, char **argv)
         die ();
 
     tty_name = argv[1];
-    if (strnlen (tty_name, 15) == 15 || strncmp (tty_name, "/dev/", 5))
+
+    // make sure the parameter is /dev/tty1 .. /dev/tty63, extract the number
+    if (strncmp (tty_name, "/dev/tty", 8))
```
mc-worker: Please use explicit test:
```strncmp (tty_name, "/dev/tty", 8) != 0```

**Comment by zyv** [comment]
> /rebase, addressed both feedback points

Just so that you better understand what's going on, the comment with the `/rebase` command (and this should be the only content of this comment) starts my rebase bot here:

https://github.com/MidnightCommander/mc/blob/master/.github/workflows/command-rebase.yml

I use it to perform the final rebase with autosquash before merges. This makes it possible for me to avoid pulling in remote changes locally, checking out the branches, performing the rebase manually, and force pushing, so it saves a lot of time and effort.

It does not bring you anything, though, and you don't have to use it (unless you know very well what and why you are doing this).

**Comment by egmontkob** [comment]
Thanks for the heads-up, I didn't know this. I thought this was merely a convention from you that I also started to follow :)

**Review by zyv — COMMENTED** [review]
I can't provide a worthy review here, but I'm happy to take care of merging if others agree. Superficially, it looks good to me, but this is of course exactly not the type of assessment that is needed in this instance.

**Comment by egmontkob** [comment]
Latest update: Just an empty lined added to make the code more consistent.

File: src/consaver/cons.saver.c [hunk 3]
```diff
@@ -176,7 +176,24 @@ main (int argc, char **argv)
         die ();
 
     tty_name = argv[1];
-    if (strnlen (tty_name, 15) == 15 || strncmp (tty_name, "/dev/", 5))
+
+    // make sure the parameter is /dev/tty1 .. /dev/tty63, extract the number
+    if (strncmp (tty_name, "/dev/tty", 8) != 0)
```
ossilator: this whole block can be replaced with a single sscanf().

ossilator (commented):
the first commit should be atomized (at least to the degree that no lines overlap).

src/consaver/cons.saver.c
egmontkob: I am not entirely certain how robust sscanf is against attacks in a setgid program.

At the very least, we should exclude additional trailing characters such as `/dev/tty4asdf`, for that my usual solution is `sscanf("/dev/tty%d%c")` and then verify in the return value that `%c` did _not_ result in a conversion, but I'm not sure how standard of a pattern this is, or how this would handle tons of leading zeros in the number, or integer overflow, or whatnot; it'd probably accept 007 as a synonym for 7, (2^32+N) as a synonym of N, etc., which is not necessarily a problem, but still not a behavior I'm happy with in security sensitive context.

That's why, given the setgid nature, I preferred manually checking the bytes. Now, _maybe_ security experts wouldn't agree with me and would also advise to use standard methods, I don't know.

egmontkob:
> the first commit should be atomized (at least to the degree that no lines overlap).

Could you please elaborate what you mean? Split into multiple smaller commits? And what do you mean by "no lines overlap"?

I'd find it really nontrivial to split it into smaller pieces, most of the changes depend on other ones. I couldn't do it without introducing (and testing, including its security aspects) intermediate codes that would be gone in the next commit.

src/consaver/cons.saver.c
ossilator: sscanf is one of the most basic libc functions, so i'd say it's rather robust.

instead of %c, you could also use %n and check that the char there is zero. but the %c variant produces less code.

the trade-off of hand-coding it is that it's quite some code with lots of conditionals. esp. in security, KISS is paramount.

it's ok to accept some technically invalid input if you can prove that it can do no harm.

ossilator:
it would be quite easy to pull the udev removal ahead and put the rdev stuff last. but upon a closer look i see that neither of this can be done with only git gui. your choice.

src/consaver/cons.saver.c
egmontkob: If I accept `/dev/tty007` or `/dev/tty4294967303` or even `/dev/tty-4294967289` as synonyms of `/dev/tty7`, I guess I don't need the `%c` or the `%n` trick because I can just as equally accept `/dev/tty7asdf`. No point in being pedantic about trailing garbage if I'm not equally pedantic about the canonical representation of the number.

However, after the sanity checks, the code opens `argv[1]` as the tty device. If the argument is `/dev/tty7/../../tmp/blah` then that. Which, in practice, is not a problem because `/dev/tty7` is not a directory so it cannot point to a user-controlled location.

Still, in security-sensitive areas, I don't feel good about looseness which can later be proved not to be able to cause problem. I'm more in favor of strict behavior at every step, which `sscanf` on its own cannot provide.

In order to remain fully strict, I can see two possibilities: one is the manual code that I have, the other is `sscanf` followed by `sprintf`ing back the parsed number, and check that the two strings are equal.

egmontkob:
udev removal can technically easily be separated (don't worry about the work needed, I don't use git gui anyway, I'm happy to do the work manually if I agree with the goal). However, it results in a totally weird code after that commit that checks for the `/dev/` prefix somewhere, and for the followup letter to have to be `t` elsewhere. This change on its own shouldn't pass a review, a careful reviewer should ask to cleanup this up.

Now the next philosophical question (and I don't know mc devs' or your take on this) is to what extent it's okay to leave unpolished things in a commit which is fixed by a subsequent one in the same PR. I'm generally not a fan of this (although my pending PR for ncurses double line drawing has something like this, actually breaks something which the next change fixes because it's quite a big additional change). But generally my preference is for every commit to polish these kinds of things within itself.

Anyway, thinking along these lines, I might slightly prefer not to split the udev removal to a separate commit, but I might change my mind.

The rdev stuff: it's a functional change, I no longer read and use the device minor number but I read and verify it against the expected value. I don't see how this could be separated.

ossilator:
yeah, don't produce invalid/messy intermediate states - that would directly contradict the idea of atomic commits.

if you can't think of a commit order that wouldn't require lots of churn, then forget the request.

ossilator:
huh, shit, something went wrong with ctrl-enter. :innocent:

egmontkob:
s/udev/devfs in previous comments

egmontkob:
Pushed a new version. `sscanf()` + `sprintf()` verification indeed looks cleaner to me, thanks for the suggestion!

(Also renamed `tty_name` vs. `console_name` to their previous usage, so the diff against master should be a bit smaller.)

ossilator (approved):
i'm somewhat confident that this doesn't make things worse. :smiling_face_with_tear:

PR #4784, opened 2025-10-11
4625 slow zsh macos

egmontkob:
## Proposed changes

Fix slow startup on macOS with zsh, and other related problems also affecting other platforms and shells.

* Resolves: #4625
* Resolves: #4780
* Resolves: #4548
* Resolves: #4723

## Checklist

👉 Our coding style can be found here: https://midnight-commander.org/coding-style/ 👈

- [x] I have referenced the issue(s) resolved by this PR (if any)
- [x] I have signed-off my contribution with `git commit --amend -s`
- [x] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

Issue #4844, opened 2025-11-01
Sometimes ` ^H` appears in front of the prompt


egmontkob:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Midnight Commander version and build configuration

```shell
- 4.8.33
- 4.8.33-402-g723392f61
```

### Operating system

```shell
Ubuntu 25.10
```

### Is this issue reproducible using the latest version of Midnight Commander?

- [x] I confirm the issue is still reproducible with the latest version of Midnight Commander

### How to reproduce

Create a shell prompt that executes slowly, e.g. place this in `~/.bashrc`:

    PS1='$(sleep 1)'"$PS1"

Start up mc. (Wait patiently, it feeds a multi-line init string to the subshell causing multiple prompts.)

Press Ctrl-O to hide the panels.

### Expected behavior

Subshell prompt should appear as normal.

### Actual behavior

` ^H` (space, caret, H) is displayed in front of the prompt.

### Additional context

`src/subshell/common.c` has this:

    write_all (mc_global.tty.subshell_pty, " \b", 2);  // Hack to make prompt reappear

Apparently this is sent when the tty line is still in cooked mode, and the kernel echoes it back in caret-notation.

Not sure if we can find a better hack instead. Nor is it clear to me why it would necessarily trigger a prompt repaint in the shell.

zyv:
FWIW, I noticed that `^H` appears with `/usr/bin/dash` as a shell on Alpine when pressing Ctrl+O :-/

PR #4859, opened 2025-11-03
Subshell initialization cleanups

### LINKED ISSUE #3792: mc overrides user subshell prompt for tcsh
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3792 |
> | **Reporter** | `citycat4` (llcdeltahw@….com) |

<p>
For ages I have tcsh as login shell. And for ages have in .cshrc this prompr setting<br />
<em>set prompt = "%{\033[1;32m%}${mch:q}:%{\033[1;34m%}[$USER]%{\033[1;36m%} \!&gt;%{\033[0m%} "</em><br />
Prior to 4.8.16 mc uses this format of prompt. Since 4.8.16 it ignores my prompt and uses their own, which I cannot change. To make mc to use my own prompt I forced to edit src/subshell/common.c.<br />
</p>

### LINKED ISSUE #4634: Clean up subshell precmd and PS1 handling

### LINKED ISSUE #4793: bash `-rcfile` vs. `--rcfile`

```shell
4.8.33
```

### Operating system

```shell
any
```

### How to reproduce

bash 1.x had `-rcfile`, bash 2.0 changed that to `--rcfile`.

As per https://github.com/MidnightCommander/mc/issues/4599#issuecomment-3383340171 I'd like to drop support for bash 1.x.

Even though the single-dash variant is still supported, let's switch to the official documented double-dash.

(Let's also check if we're using obsolete flags with any other shell.)

### Expected behavior

x

### Actual behavior

x

### Additional context

_No response_

### LINKED ISSUE #4824: Broken PS1 prompt when using mksh

```shell
GNU Midnight Commander 4.8.33
Built with GLib 2.84.1
Built with libssh2 1.11.1
With builtin editor
Virtual File Systems:
Data types:
```

### Operating system

```shell
Linux clip 6.17.4-0-stable #1-Alpine SMP PREEMPT_DYNAMIC 2025-10-21 08:38:12 x86_64 GNU/Linux
```

### How to reproduce

1. Set your shell to mksh
2. Set PS1 variable. Fox example to: `PS1='$PWD $ '` 
3. Run Midnight Commander
4. Observe the PS1 prompt displayed by the subshell

### Expected behavior

The PS1 prompt displayed by the subshell should be displayed properly. Variable and command substitution should work as expected.

### Actual behavior

The PS1 prompt displayed by the subshell is not displayed correctly. For the example I gave above (`PS1='$PWD $ '`) it literally displays the string '$PWD $ ' without properly substituting the variable. No variable or command substitutions work.

### Additional context

Seems related to feature #4634. Opening this up as a separate bug as the behavior I'm describing is in fact a bug in the current behavior of Midnight Commander and not a desired new feature.

Midnight Commander seems to be doing a lot of unnecessary convoluted things with respect to the PS1 prompt under mksh. As a temporary workaround adding the following to my mkshrc after the point where PS1 is defined fixes things for now by preventing Midnight Commander from overwriting the PS1 variable and preventing it from using its own broken `precmd()` function (makes PS1 read-only):
```
# hack to fix Midnight Commander with mksh
if [ "$MC_SID" ]
then
    mc_fixed_precmd() { pwd >&9; kill -STOP $$; }
    PS1="\$(mc_fixed_precmd)$PS1"
    typeset -r PS1
fi
```

On my system file descriptor 9 always seems to be correct for the workaround but your mileage may vary.

egmontkob:
## Proposed changes

Some of the easy bits extracted from enormous and still WIP PR #4784. Plus a few new ones.

Mostly cosmetical, although there are a few small functional changes / fixes.

Let's land these please, and I'll rebase / continue 4784 on top of these.

* Resolves: #3792
* Resolves: #4634
* Resolves: #4793
* Resolves: #4824

## Checklist

### DIFF
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -1181,44 +1181,27 @@ init_subshell_precmd (void)
 {
     /*
      * Fallback precmd emulation that should work with virtually any shell.
-     * No real precmd functionality is required, no support for \x substitutions
-     * in PS1 is needed. For convenience, $HOME is replaced by ~ in PS1.
      *
-     * The following example is a little less fancy (home directory not replaced)
-     * and shows the basic workings of our prompt for easier understanding:
+     * Explanation of the indirect hop via $MC_PRECMD:
      *
-     * " precmd() {"
-     *     " echo \"$USER@$(hostname -s):$PWD\";"
-     *     " pwd >&%d;"
-     *     " kill -STOP $$;"
-     * " };"
-     * " PRECMD=precmd;"
-     * " PS1='$($PRECMD)$ '\n",
+     * Scenario: The user exports PS1 and then invokes a sub-subshell (e.g. "sh").
      *
-     * Explanations:
+     * This would lead to the sub-subshell stopping (=frozen mc):
+     *     PS1='$(pwd >&%d; kill -STOP $$)...'
      *
-     * A: This leads to a stopped subshell (=frozen mc) if user calls "ash" command
-     *    "PS1='$(pwd >&%d; kill -STOP $$)\\u@\\h:\\w\\$ '\n",
+     * This would lead to an error message like "sh: mc_precmd: not found":
+     *     mc_precmd() { pwd >&%d; kill -STOP $$; }
+     *     PS1='$(mc_precmd)...'
      *
-     * B: This leads to "sh: precmd: not found" in sub-subshell if user calls "ash" command
-     *    "precmd() { pwd >&%d; kill -STOP $$; }; "
-     *    "PS1='$(precmd)\\u@\\h:\\w\\$ '\n",
-     *
-     * C: This works if user calls "ash" command because in sub-subshell
-     *    PRECMD is undefined, thus evaluated to empty string - no damage done.
-     *    Attention: BusyBox must be built with FEATURE_EDITING_FANCY_PROMPT to
-     *    permit \u, \w, \h, \$ escape sequences. Unfortunately this cannot be guaranteed,
-     *    especially on embedded systems where people try to save space, so let's use
-     *    the fallback version.
+     * A hop via $MC_PRECMD works because in the sub-subshell MC_PRECMD is undefined (assuming the
+     * user did not export this one), thus evaluated to empty string - no damage done.
      */
-    static const char *precmd_fallback = " MC_PS1_SAVED=\"$PS1\";"  // Save custom PS1
-                                         " precmd() {"
-                                         "   echo \"$MC_PS1_SAVED\";"
+    static const char *precmd_fallback = " mc_precmd() {"
                                          "   pwd >&%d;"
                                          "   kill -STOP $$;"
                                          " };"
-                                         " PRECMD=precmd;"
-                                         " PS1='$($PRECMD)'\n";
+                                         " MC_PRECMD=mc_precmd;"
+                                         " PS1='$($MC_PRECMD)'\"$PS1\"\n";
 
     switch (mc_global.shell->type)
     {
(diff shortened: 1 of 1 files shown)


zyv:
I have to admit that I've completely lost track of the whole subshell business by now, but didn't we (in #4634)

1. Decide that we shouldn't override what the user selected as PS1, as the snippets have a difficult history and evolved through many misunderstandings, like the initial desire to have a uniform prompt across shells, then subsequent edits that broke that anyways, etc.
2. And also come to the conclusion that it's impossible to have `PS1` that is not set, and if it's somehow unset actively by the user, then it's their problem, so the fallbacks for unset PS1 per shell should be removed?

If yes and you agree, then I think that this aptly titled PR is the right place to do that. I'm making a non-line comment, because you certainly know which bits I'm referring to and it's easier to keep the conversation not bound to one line.

egmontkob:
1. Yup that definitely remains in that other PR, I just wasn't 100% confident moving them to this "easy bits" PR, they might be a bit more heavy-weight. But I can move here if you feel like it deserves to.

2. I might have missed that bit; indeed we can refactor accordingly but that change hasn't been written yet.

zyv:
> 1. Yup that definitely remains in that other PR, I just wasn't 100% confident moving them to this "easy bits" PR, they might be a bit more heavy-weight. But I can move here if you feel like it deserves to.
> 2. I might have missed that bit; indeed we can refactor accordingly but that change hasn't been written yet.

My understanding was that you (rightfully) decided to limit the scope of the "big" PR to the fixes around the lockups / synchronization story and move refactorings and optimization to a separate "easy bits" PR, so I would appreciate it if you could address both of these items here as well. Or maybe open yet another topical PR if you prefer to merge in even smaller steps. Whatever works for you. I'm just a bit confused about what will be done where and when and if at all, and I didn't want to miss anything, so I asked.

egmontkob:
> 2. And also come to the conclusion that it's impossible to have PS1 that is not set, and if it's somehow unset actively by the user, then it's their problem, so the fallbacks for unset PS1 per shell should be removed?

So, this affects the fallback variant and ksh, am I right? And for the fallback, the large 5-line if-then-else can be removed because MC_PWD will no longer be used at all?

While at it, it's time to fix #4824 as well, I guess.

Hang on...

egmontkob:
Few new commits on the top to address the remaining issues. The fallback code becomes much simpler, I hope it doesn't introduce some regression in some weird cases.

Anyway, some of it is newly written code, so let's give it some testing.

Also sneaked in another tiny cleanup commit midway-ish about BASH_VERSION vs. BASH_VERSINFO.

src/subshell/common.c
```diff
@@ -1256,8 +1234,7 @@ init_subshell_precmd (void)
 
     case SHELL_KSH:
         // pdksh based variants support \x placeholders but not any "precmd" functionality
```
zyv: I think it would be nice to remove the comments about placeholders, since we are not trying to provide fallbacks or override user prompts anymore. If they remain, I think this would be just confusing.

src/subshell/common.c
zyv: See R1236.

zyv: See R1236.

zyv (approved):
I have reviewed this and have done some testing on Alpine.

| Shell | Version | PS1 & PWD | Persistence | Comments |
|--------|--------|--------|--------|--------|
| ash | BusyBox v1.36.1 | ✅  | ➖  | PS1 with PWD |
| bash | GNU bash, version 5.2.26(1)-release | ✅  | ✅ | PS1 with PWD |
| dash | dash-0.5.12-r2 | ✅ | ➖ | PS1: no PWD; #4844 |
| fish | version 3.7.1 | ✅ | ✅ | PS1 with PWD, colored on Ctrl+O |
| mksh | @(#)MIRBSD KSH R59 2020/10/31 | ✅ | ➖ | PS1: no PWD |
| tcsh | 6.24.12 (Astron) | ✅ | ➖ | PS1: no PWD | 
| zsh | zsh 5.9 | ✅ | ✅ | PS1: no PWD | 

I think that everything is working as expected. I don't think I can help more, and barring the comments that I think should be removed now (see line comments), this is a great job and makes the code simpler and more maintainable. Thank you!

@d3m3t3r, maybe you could test this on your shells...

src/subshell/common.c
egmontkob: Done (all three; plus also simplified the comment at BusyBox).

egmontkob:
> I have reviewed this and have done some testing on Alpine.

Thanks a lot for testing (and for reviewing of course)!

mc-worker (approved):

PR #4865, opened 2025-11-09
Ticket #4865: add %v and %V macros to user menu

Bad-ptr:
## Proposed changes

%v and %V macros are similar to %t and %T, but expands to full names of tagged files.

## Checklist


zyv (changes requested):
Thank you for your contribution!

1. Could you please sign off your commits as requested in the PR template?
2. Could you please add a test as requested in the PR template?

Unfortunately, I'm not able to take changes like this without a unit test, and unit-testing this particular function doesn't appear to be too hard. I also don't have the capacity to write the test for you.

You can find an example for a test, which I wrote to test another `usermenu` function some time ago here:

https://github.com/MidnightCommander/mc/tree/master/tests/src

src/usermenu.c
```diff
@@ -948,22 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
 
     case 't':
     case 'u':
+    case 'v':
     {
         GString *block = NULL;
         int i;
+        const char *cwd;
```
mc-worker: Please reduce the scope of this variable.

src/usermenu.c
```diff
@@ -948,22 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
...
+        char *qcwd;
+
+        qcwd = NULL;
```
mc-worker: Why not the variable initialization?

src/usermenu.c
```diff
@@ -948,22 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
...
+
+            qcwd = quote_func (cwd, FALSE);
+        }
+
+        block = g_string_sized_new (64);
+
         for (i = 0; i < panel->dir.len; i++)
             if (panel->dir.list[i].f.marked != 0)
             {
+                if (qcwd != NULL)
```
mc-worker, line 977: Please move this block after variable declaration.

src/usermenu.c
```diff
@@ -979,6 +1008,8 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
                 if (c_lc == 'u')
                     do_file_mark (panel, i, 0);
             }
+        if (qcwd != NULL)
```
mc-worker: This test is redundant, `g_free` makes it itself.

src/usermenu.c
```diff
@@ -948,22 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
...
+
+        qcwd = NULL;
 
         if (panel == NULL)
         {
             result = NULL;
             goto ret;
         }
 
+        if ('v' == c_lc)
```
mc-worker: Please change to
``` c_lc == 'v'```

File: src/usermenu.c [hunk 6]
```diff
@@ -948,22 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
...
+        }
+
+        block = g_string_sized_new (64);
+
         for (i = 0; i < panel->dir.len; i++)
             if (panel->dir.list[i].f.marked != 0)
             {
+                if (qcwd != NULL)
+                {
+                    if (block == NULL)
```
mc-worker: This condition is always false because `block` was created before this loop.

src/usermenu.c
```diff
@@ -948,22 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
...
+                    }
+                }
+
                 char *tmp;
 
                 tmp = quote_func (panel->dir.list[i].fname->str, FALSE);
+
                 if (tmp != NULL)
                 {
                     if (block == NULL)
```
mc-worker: Now this condition is always false too.

src/usermenu.c
mc-worker: Done

src/usermenu.c
```diff
@@ -948,37 +948,51 @@ expand_format (const Widget *edit_widget, char c, gboolean do_quote)
...
+                    g_string_append (block, tmp);
+                    g_free (tmp);
                     g_string_append_c (block, ' ');
                 }
 
                 if (c_lc == 'u')
                     do_file_mark (panel, i, 0);
             }
+        g_free (qcwd);
         result = block == NULL ? NULL : g_string_free (block, block->len == 0);
```
mc-worker: Block is never NULL here.

Bad-ptr:
Updated

@zyv I added tests

zyv:
> Updated
> 
> @zyv I added tests

Oh, the tests make me very happy. Thank you very much!

zyv (approved):
This looks good to me now. Thank you!

mc-worker (approved):

Issue #4507, opened 2023-10-14
Shell link shows incorrect file names with cyrillic or diacritic symbols


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4507 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `lnx` (reg@….lv) |

<p>
When using Shell link, filenames with cyrillic or diacritic symbols are displayed as chains of numbers. It is not possible to perform operations with such files.<br />
example contents of some directory:<br />
</p>

```
ls -gn
-rw-r--r-- 1 1000 0 Oct 14 13:11 aaa
-rw-r--r-- 1 1000 0 Oct 14 13:26 āāā
-rw-r--r-- 1 1000 0 Oct 14 13:11 bbb
lrwxrwxrwx 1 1000 6 Oct 14 13:26 fff -> āāā
lrwxrwxrwx 1 1000 6 Oct 14 13:26 sss -> ššš
-rw-r--r-- 1 1000 0 Oct 14 13:26 ššš
-rw-r--r-- 1 1000 0 Oct 14 13:26 ввв
-rw-r--r-- 1 1000 0 Oct 14 13:26 ггг
lrwxrwxrwx 1 1000 6 Oct 14 13:26 ддд -> ššš
```
<p>
how the same directory is displayed via shell link (ssh) to localhost:<br />
</p>

```
/..                     │UP--DIR│Oct 14 13:44
│ 304201304201304201    │      0│Oct 14 13:26
│ 305241305241305241    │      0│Oct 14 13:26
│ 320262320262320262    │      0│Oct 14 13:26
│ 320263320263320263    │      0│Oct 14 13:26
│@320264320264320264    │      6│Oct 14 13:26
│ aaa                   │      0│Oct 14 13:11
│ bbb                   │      0│Oct 14 13:11
│@fff                   │      6│Oct 14 13:26
│@sss                   │      6│Oct 14 13:26
```
<p>
Problem is with <strong>fish_list_perl()</strong> in <strong>/usr/lib/mc/fish/ls</strong>:<br />
S_IMODE and S_IFMT subroutines are not being imported from <em>Fcntl</em> <br />
due to line 128 <strong><em>import Fcntl ":mode" unless defined &amp;S_ISLNK;</em></strong><br />
</p>
<p>
[...]

PR #4867, opened 2025-11-09
vfs: shell: Add support of UTF-8 encoding for SSH connections

KuzinAndrey:
Most linux distributions today use UTF-8 encoding for theirs filesystems, because of that use LANG=C and LC_ALL=C make filenames unreadable and unusable on SSH connections (it fill names by question marks).

Change all mentions of LANG=C to LANG=C.UTF-8, also use as default env.

### DIFF
--- a/src/vfs/shell/shell.c
+++ b/src/vfs/shell/shell.c
@@ -459,6 +459,9 @@ shell_set_env (int flags)
 
     ret = g_string_sized_new (256);
 
+    g_string_append (ret, "LANG=C.UTF-8; export LANG; ");
+    g_string_append (ret, "LC_ALL=C.UTF-8; export LC_ALL; ");
+
     if ((flags & SHELL_HAVE_HEAD) != 0)
         g_string_append (ret, "SHELL_HAVE_HEAD=1 export SHELL_HAVE_HEAD; ");
 
@@ -631,7 +634,7 @@ shell_open_archive_int (struct vfs_class *me, struct vfs_s_super *super)
 
     // Set up remote locale to C, otherwise dates cannot be recognized
     if (shell_command (me, super, WAIT_REPLY,
-                       "LANG=C LC_ALL=C LC_TIME=C; export LANG LC_ALL LC_TIME;\n"
+                       "LANG=C.UTF-8 LC_ALL=C.UTF-8 LC_TIME=C; export LANG LC_ALL LC_TIME;\n"
                        "echo '### 200'\n",
                        -1)
         != COMPLETE)


zyv:
This reminds me of #4507.

Our file listing support for FISH is organized as follows: 

1. We try Perl if available.
2. We try `ls -lan` & `sed` if available.
3. We try `ls -lan` as the last resort.

We need formats and messages in POSIX locale to be able to parse them reliably. However, the Perl script can return filenames without garbling them even in POSIX locale.

Apparently, your host does not support Perl, so you are seeing this problem due to fallback scripts. Is this correct?

If yes, then the question is how to best fix it. Your solution has several problems:

1. To my knowledge, many systems do not have `C.UTF-8` locale at all or have `C.utf-8` or `C.utf8` locales (how compliant is that, and what to comply with is a different question). What will be the behavior on such systems, would it fall back to `C` or not? I wasn't able to figure out quickly what should be the behavior according to the standard: https://pubs.opengroup.org/onlinepubs/9799919799/ . Can you find out more?
2. Non-UTF displays like systems running with KOI8-R are still supported by mc. That is, we have a concept of file system locale and display locale. Display is what your terminal running mc has, and I/O or filesystem is what it interacts with underneath. If your change is accepted as is, then I would assume it will break FISH on such systems.

We could make this conditional on mc display locale being UTF-8, which is maybe not ideal, but would also avoid breakage.

We could also check that filesystem encoding for FISH is UTF-8, and whatever goes to mc itself is then converted in display locale properly. I think this is a better way. I'm not familiar with the code there; you have to investigate. Or maybe @aborodin can say something about that if he knows.

And lastly, is there any way to limit the damage, like only defining some `LC_` variable that influences filename output to `C.UTF-8`?

Sorry, but this cannot be accepted as is without much investigation and research. Making changes like that to scripts to support "poor" systems is especially dangerous. I hope you have time and energy to go deeper. I cannot support much more.

KuzinAndrey:
This issue i have with Alpine Linux busybox version of ls (and may be this understanble). In my environment i make some tests:
```sh
router:~# touch прикол.txt
router:~# ls -Qlan | grep txt
-rw-r--r--    1 0        0                0 Nov  9 14:17 "прикол.txt"

router:~# echo $LANG
C.UTF-8

router:~# LANG=C ls -Qlan | grep txt
-rw-r--r--    1 0        0                0 Nov  9 14:17 "????????????.txt"

router:~# cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.22.2
PRETTY_NAME="Alpine Linux v3.22"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
```

On Debian 12 this looks differently for `LANG=C`:
```
user@localhost:/$ touch прикол.txt

user@localhost:/$ ls -Qlan | grep txt
-rw-r--r-- 1 4013 4014        0 Nov  9 14:16 "прикол.txt"

user@localhost:/$ echo $LANG
C.UTF-8

user@localhost:/$ LANG=C ls -Qlan | grep txt
-rw-r--r-- 1 4013 4014        0 Nov  9 14:16 "\320\277\321\200\320\270\320\272\320\276\320\273.txt"

user@localhost:/$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

And also in Ubuntu :
```
root@s1:~# touch прикол.txt

root@s1:~# ls -Qlan | grep txt
-rw-r--r--  1 0 0         0 Nov  9 14:29 "прикол.txt"

root@s1:~# echo $LANG
C.UTF-8

root@s1:~# LANG=C ls -Qlan | grep txt
-rw-r--r--  1 0 0         0 Nov  9 14:29 "\320\277\321\200\320\270\320\272\320\276\320\273.txt"

root@s1:~# cat /etc/os-release 
NAME="Ubuntu"
VERSION="18.04.6 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.6 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

So this gives me the following conclusions:
- **most** modern systems has С.UTF-8 as default codepage (and this didn't break anything for me)
- busybox utilities has defective functionality - this understandable (they support only really need things)
- for my case `mc` already broken - and i fix it (this just work for me already now)
- forget about dinosaurs technology koi8-r - they all already died (If I find even one such person, I will offer him to revert this commit himself)
- Perl already died too for modern OS on my opinion (i get really annoyed when the package manager installs the Perl interpreter as a dependency of MidnightCommander, i've had enough of ugly Python everywhere, i recommend use only Shell or C-lang wherever possible to make less dependencies for project, especially for remote code executions)

By comments in code LANG=C and LC_TIME=C was used for make parseable file creation date, but this make unusable filenames in my case. I think this part of code may be rewritten with `ls --full-time` supported by common `ls` and `busybox` version and don't touch LANG=/LC_TIME= at all and use the remote host default value. But i don't want to make this deep diving.

But as always this looks differently too:
```
# common version
$ ls -Qlan --full-time | grep txt
-rw-r--r-- 1 4013 4014        0 2025-11-09 14:16:38.151720438 +0300 "прикол.txt"

# busybox version
router:~# ls -Qlan --full-time | grep txt
-rw-r--r--    1 0        0                0 2025-11-09 14:17:14 +0300 "прикол.txt"
```

zyv:
> * **most** modern systems has С.UTF-8 as default codepage (and this didn't break anything for me)
> * busybox utilities has defective functionality - this understandable (they support only really need things)

Thank you for the additional research! So, the reason why the filenames work correctly now on the systems without Perl, but with GNU ls is because it encodes code points in file names as escape sequences if the locale is set to POSIX, which is actually great.

Thus one way to solve this issue would be to add this functionality to BusyBox. I don't see a reason as to why Denys would refuse it. I think this would be an ideal solution to the problem. It should not be too hard to code this up either. How about that?

One of the goals of shell protocol support in mc (as opposed to sftp, which sadly is also in bad shape) is to work with any crippled and/or legacy system that system administrators have to work with, including all flavors of BusyBox, BSD, Solaris, AIX, etc. Thus, the fact that **modern** systems support UTF-8, which is great, doesn't help us much at all.

> * forget about dinosaurs technology koi8-r - they all already died (If I find even one such person, I will offer him to revert this commit himself)

One of the maintainers of mc (not me) is still running all his systems on single-byte locale, and specifically KOI8-R, and it is important for him that it keeps working.

When I'm explaining something to you or asking to do something, it is not because I find it funny to annoy other people. It is because there are some constraints or reasons why I can't make certain decisions.

> By comments in code LANG=C and LC_TIME=C was used for make parseable file creation date, but this make unusable filenames in my case. I think this part of code may be rewritten with `ls --full-time` supported by common `ls` and `busybox` version and don't touch LANG=/LC_TIME= at all and use the remote host default value. But i don't want to make this deep diving.

Sadly, `--full-time` is not POSIX, and even if it were, it wouldn't work on most legacy systems.

---

Anyways, I'd suggest first trying to get an escaping patch into BusyBox. This will help tons of people, also those not even using mc.

Otherwise, I'm fine to accept this patch with an extra check that the current display encoding **is** UTF-8 and some reference to how the systems behave if an invalid locale (in their view) is entered. Do they fall back to `C`? If yes, then it's great.

If you don't have time and/or interest in working on that, I'm sorry about this, but there is not much else I can do at the moment.

KuzinAndrey:
Already in master branch of busybox - https://github.com/mirror/busybox/commit/4f3a56dc12a2126bdacff73f1cfe586d06e800c0
but not in latest release 1.37.0 :(

zyv:
> Already in master branch of busybox - [mirror/busybox@4f3a56d](https://github.com/mirror/busybox/commit/4f3a56dc12a2126bdacff73f1cfe586d06e800c0) but not in latest release 1.37.0 :(

Thanks for the confirmation! I think that's great news. As soon as the new release is made and spreads, this particular issue will then "fix itself".

For now, I think you have the following options:

1. Carry this patch yourself or install a custom shell script like explained in the documentation: https://github.com/MidnightCommander/mc/blob/master/src/vfs/shell/helpers/README.shell
2. Install Perl on the target machine.
3. Use SFTP, even if it's not in good shape, see #3654.

I'll close this PR for now.

PR #4882, opened 2025-11-13
Fix broken arj support.

moose-kazan:
[Here](https://github.com/MidnightCommander/mc/pull/4842#issuecomment-3477631213)  @zyv mentioned broken arj helper script and lack of tests for it.

I try to fix this problem. I was rewrote helper script in perl language and fix regexes. Also i was added some tests. For now it work correctly on my test arj-files.


tests/src/vfs/extfs/helpers-list/data/uarj.README
```diff
@@ -0,0 +1,5 @@
+uarj.input is created with wimtools from uarj.arj
```
zyv:
```suggestion
uarj.input is created with arj from uarj.arj
```
?

zyv (approved):
Thank you for fixing this, and it's great to have one more extfs script covered with at least a basic test.

PR #4886, opened 2025-11-14
Ticket #4886: fix broken alz support

moose-kazan:
Support for ALZ archives was broken. Code required unalz v0.50, which very old.

I was rewrote code to support actual version (tested with v0.65) and test was added.


zyv (approved):
Thank you for fixing this extfs module and for adding a test!

PR #4869, opened 2025-11-09
Bugfix: N_ macro used instead of _

Strahinja:
## Proposed changes

Currently, `N_` macro is errorneously used instead `_` in some places related to the About box (see issue #4868). This causes related messages in the About boxes of MC and mcedit to be displayed in English instead of localized. This PR aims to correct this error.

* Resolves: #4868 

## Checklist

👉 Our coding style can be found here: https://midnight-commander.org/coding-style/ 👈

- [x] I have referenced the issue(s) resolved by this PR (if any)
- [x] I have signed-off my contribution with `git commit --amend -s`
- [ ] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #4887, opened 2025-11-14
Ticket #4868: avoid nested gettext calls and fix untranslated strings

### LINKED ISSUE #4868: Some strings in the About boxes are not translated in MC and mcedit

```shell
GNU Midnight Commander 4.8.33-428-g08f77a932
Built with GLib 2.84.4 (using GLib 2.84.4)
Built with libssh2 1.11.0
With builtin editor
With mouse support on xterm
With support for X11 events
Virtual File Systems:
Data types:
```

### Operating system

```shell
OpenBSD mithlond.local 7.8 GENERIC.MP#54 amd64
```

### How to reproduce

1. Set `LANG` environment variable to any language with translated About box messages, eg. `export LANG=sr_RS.UTF-8`
2. Run MC and/or mcedit, open About box
3. Some of the strings are not translated, namely:
   * `Detected display codepage` ([.po file translation](https://github.com/MidnightCommander/mc/blob/37bf3376d8a209e306908bb0fe20c3b48e72422e/po/sr.po#L1826)) and `Selected source (file I/O) codepage` (currently not translated, but if translated in .po file exhibits the same behavior as the other two) for main MC,
   * `A user friendly text editor\nwritten for the %s.` ([.po file translation](https://github.com/MidnightCommander/mc/blob/37bf3376d8a209e306908bb0fe20c3b48e72422e/po/sr.po#L1583)) for mcedit.

### Expected behavior

The messages should be translated to the set language.

### Actual behavior

The messages are shown in English.

### Additional context

The problem seems to lie in the following code:

https://github.com/MidnightCommander/mc/blob/37bf3376d8a209e306908bb0fe20c3b48e72422e/src/editor/editwidget.c#L143

https://github.com/MidnightCommander/mc/blob/37bf3376d8a209e306908bb0fe20c3b48e72422e/src/filemanager/boxes.c#L501-L504

When `N_` macro is replaced with `_`, the translated messages are shown as expected, in the set language.

egmontkob:
## Proposed changes

Make widgets take already translated strings. This refactoring helps prevent some untranslated or double-translated strings.

* Resolves: #4868
* Resolves: #4869

## Checklist


zyv (approved):

zyv:
@Strahinja, would you be so kind to have a look at this and test it? Thank you!

lib/search/search.c
```diff
@@ -312,6 +312,19 @@ mc_search_is_type_avail (mc_search_type_t search_type)
 const mc_search_type_str_t *
 mc_search_types_list_get (size_t *num)
 {
+#ifdef ENABLE_NLS
+    static gboolean i18n_flag = FALSE;
+
+    if (!i18n_flag)
+    {
+        size_t i;
+        for (i = 0; i < G_N_ELEMENTS (mc_search__list_types); i++)
```
mc-worker: ```for (size_t i = 0; ```

lib/widget/quick.h
```diff
@@ -213,8 +213,8 @@
 
 #define QUICK_BUTTONS_OK_CANCEL                                                                    \
     QUICK_START_BUTTONS (TRUE, TRUE),                                                              \
-        QUICK_BUTTON (N_ ("&OK"), B_ENTER, NULL, NULL),                                            \
-        QUICK_BUTTON (N_ ("&Cancel"), B_CANCEL, NULL, NULL)
+        QUICK_BUTTON (_ ("&OK"), B_ENTER, NULL, NULL),                                            \
```
mc-worker: Looks like an Indentation fault of `\ ` .

egmontkob:
@mc-worker done (both).

@Strahinja Yup could you please check, especially the commit I modified but still retained you as author: Does it look good to you? There was a change on master which made me have to adjust it. Also I changed the commit wording to match mc's style. Please let us know if it looks okay to you.

mc-worker (approved):
There is no need to write same "Ticket #XXX" in the every commit other than the first one.

Strahinja:
Sorry I didn't get back earlier, I have a day job. I might check this out today or tomorrow.

Strahinja:
I checked out this fork and it seems to work correctly with regards to the issue. On an unrelated note, I got this warning compiling under OpenBSD 7.8:
```
utilunix.c: In function 'mc_realpath':
utilunix.c:1067:13: warning: 'strchr' reading 1 or more bytes from a region of size 0 [-Wstringop-overread]
 1067 |         p = strchr (p, PATH_SEP);
      |             ^~~~~~~~~~~~~~~~~~~~
```
(The file _utilunix.c_ in question is in the _lib_ subdirectory in the repository.)

zyv:
> I checked out this fork and it seems to work correctly with regards to the issue. On an unrelated note, I got this warning compiling under OpenBSD 7.8:
> 
> ```
> utilunix.c: In function 'mc_realpath':
> utilunix.c:1067:13: warning: 'strchr' reading 1 or more bytes from a region of size 0 [-Wstringop-overread]
>  1067 |         p = strchr (p, PATH_SEP);
>       |             ^~~~~~~~~~~~~~~~~~~~
> ```
> 
> (The file _utilunix.c_ in question is in the _lib_ subdirectory in the repository.)

I think that this is a false positive. 

Apparently, the `strchr` DFA model doesn't understand that `p` cannot be `NULL` after removing the prefix, because the prefix is guaranteed to be there due to an explicit `if` check above. I wonder if something like that would help the DFA:

```c
master % git diff
diff --git a/lib/utilunix.c b/lib/utilunix.c
index 4ae316daa..c4f844dd4 100644
--- a/lib/utilunix.c
+++ b/lib/utilunix.c
@@ -1063,7 +1063,7 @@ mc_realpath (const char *path, char *resolved_path)
     // ignore encoding: skip "#enc:"
     if (g_str_has_prefix (p, VFS_ENCODING_PREFIX))
     {
-        p += strlen (VFS_ENCODING_PREFIX);
+        p += sizeof VFS_ENCODING_PREFIX - 1;
         p = strchr (p, PATH_SEP);
         if (p != NULL)
         {
```

If not, then I'm afraid the only way to mute the warning would be to add a useless and confusing NULL check.

```c
master % git diff
diff --git a/lib/utilunix.c b/lib/utilunix.c
index 4ae316daa..8fdd16677 100644
--- a/lib/utilunix.c
+++ b/lib/utilunix.c
@@ -1064,7 +1064,8 @@ mc_realpath (const char *path, char *resolved_path)
     if (g_str_has_prefix (p, VFS_ENCODING_PREFIX))
     {
         p += strlen (VFS_ENCODING_PREFIX);
-        p = strchr (p, PATH_SEP);
+        if (p != NULL)
+            p = strchr (p, PATH_SEP);
         if (p != NULL)
         {
             if (!absolute_path && p[1] != '\0')
```

I wouldn't do that though, and either report it to gcc, or just ignore it. We don't activate `-Wstringop-overread` in our build system, so whoever does that (OpenBSD) should deal with false positives in my opinion.

PR #4897, opened 2025-11-17
Ticket #4870: editor: Show the character code in more conventional format

### LINKED ISSUE #4870: mcedit: Character code in header bar weirdnesses

```shell
4.8.33
```

### Operating system

```shell
Linux
```

### How to reproduce

Multiple closely related issues, with the way the code of the character under the cursor is shown in mcedit's header bar.

- If the codepage is UTF-8, the decimal value is shown in at least 4 digits, with leading zeros if necessary. This is very unusual for decimal numbers, and there's nothing special at the boundary of 999-1000 or 9999-10000 to justify the width of at least 4. Moreover, e.g. for the letter `e` being `0101`, letter `é` being `0233` etc. it really confusingly looks like as if it was octal notation. I recommend to have no enforced minimum size and no leading zeros, just like it's already the case with non-UTF-8 codepage.

- The hex code is shown with at least 3 hex digits (e.g. `0x0E9`), both with legacy 8-bit and UTF-8 codepages. This is, again, highly unusual. I recommend fixed two digits for non-UTF-8 codepages (e.g. `0xE9`) and the standard `U+` notation with at least 4 digits (e.g. `U+00E9`) for UTF-8. (If one day we'll have proper combining accent support then this could become multiple codepoints.)

- In UTF-8, lone bytes that are invalid UTF-8 are denoted in the exact same way as Unicode characters of the given value. E.g. create a 3-byte file whose hexdump is e9 c3 a9. This is a lone 0xE9 byte (invalid UTF-8; `é` in Latin-1) followed by U+00E9 (`é`) in UTF-8. Using UTF-8 codepage, the first is displayed as a highlighted dot, the second is displayed as an `é`, as expected. Yet, as you move the cursor, the character code is equally displayed as "0233 0x0E9" for both of them. Suggestion: show "0xE9" for lone bytes and "U+00E9" for proper UTF-8-encoded Unicode chars.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
Show hex byte value in 2 digits, decimal value in at most 3 digits. Show Unicode value in standard U+xxxx notation with at least 4 digits.

In UTF-8 mode, distinguish proper Unicode values from invalid UTF-8 single bytes.

* Resolves: #4870

## Checklist

### DIFF
--- a/src/editor/editdraw.c
+++ b/src/editor/editdraw.c
@@ -95,45 +95,64 @@ printwstr (const char *s, int len)
 
 /* --------------------------------------------------------------------------------------------- */
 
-static inline void
-status_string (WEdit *edit, char *s, int w)
-{
-    char byte_str[16];
+#define CHAR_CODE_BUF_SIZE BUF_TINY
 
-    /*
-     * If we are at the end of file, print <EOF>,
-     * otherwise print the current character as is (if printable),
-     * as decimal and as hex.
-     */
+static char *
+format_character_code (WEdit *edit)
+{
+    char *buf;
+
+    buf = g_malloc (CHAR_CODE_BUF_SIZE);
+
     if (edit->buffer.curs1 >= edit->buffer.size)
-        strcpy (byte_str, "<EOF>     ");
+        strncpy (buf, "<EOF>   ", CHAR_CODE_BUF_SIZE);
     else if (edit->utf8)
     {
         unsigned int cur_utf;
-        int char_length = 1;
+        int char_length = 0;
 
         cur_utf = edit_buffer_get_utf (&edit->buffer, edit->buffer.curs1, &char_length);
         if (char_length > 0)
-            g_snprintf (byte_str, sizeof (byte_str), "%04u 0x%03X", (unsigned) cur_utf,
-                        (unsigned) cur_utf);
+            if (cur_utf < 128)
+                g_snprintf (buf, CHAR_CODE_BUF_SIZE, "%3u 0x%02X", cur_utf, cur_utf);
+            else
+            {
+                char uplus[CHAR_CODE_BUF_SIZE];
+                g_snprintf (uplus, CHAR_CODE_BUF_SIZE, "U+%04X", cur_utf);
+                g_snprintf (buf, CHAR_CODE_BUF_SIZE, "%8s", uplus);
+            }
         else
         {
+            // cursor is over some invalid UTF-8 fragment
             cur_utf = edit_buffer_get_current_byte (&edit->buffer);
-            g_snprintf (byte_str, sizeof (byte_str), "%04d 0x%03X", (int) cur_utf,
-                        (unsigned) cur_utf);
+            g_snprintf (buf, CHAR_CODE_BUF_SIZE, "%3u 0x%02X", cur_utf, cur_utf);
         }
     }
     else
     {
-        unsigned char cur_byte;
+        unsigned int cur_byte;
 
         cur_byte = edit_buffer_get_current_byte (&edit->buffer);
-        g_snprintf (byte_str, sizeof (byte_str), "%4d 0x%03X", (int) cur_byte, (unsigned) cur_byte);
+        g_snprintf (buf, CHAR_CODE_BUF_SIZE, "%3u 0x%02X", cur_byte, cur_byte);
     }
+
+    return buf;
+}
+
+#undef CHAR_CODE_BUF_SIZE
+
+/* --------------------------------------------------------------------------------------------- */
+
+static inline void
+status_string (WEdit *edit, char *s, int w)
+{
+    char *character_code;
+
+    character_code = format_character_code (edit);
 
     // The field lengths just prevent the status line from shortening too much
     if (edit_options.simple_statusbar)
-        g_snprintf (s, w, "%c%c%c%c %3ld %5ld/%ld %6ld/%ld %s %s",
+        g_snprintf (s, w, "%c%c%c%c %3ld %5ld/%ld %6ld/%ld [%s] %s",
                     edit->mark1 != edit->mark2 ? (edit->column_highlight ? 'C' : 'B') : '-',  //
                     edit->modified != 0 ? 'M' : '-',                                          //
                     macro_index < 0 ? '-' : 'R',                                              //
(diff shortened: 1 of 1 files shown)


src/editor/editdraw.c
```diff
@@ -95,45 +95,61 @@ printwstr (const char *s, int len)
 
 /* --------------------------------------------------------------------------------------------- */
 
-static inline void
-status_string (WEdit *edit, char *s, int w)
+static char *
+format_character_code (WEdit *edit)
 {
-    char byte_str[16];
+    const int bufsize = 16;
```
zyv:
```suggestion
    enum { bufsize = 16 };
```

Not awesome, but probably better than `#define`?

src/editor/editdraw.c
```diff
@@ -95,45 +95,61 @@ printwstr (const char *s, int len)
...
 
         cur_utf = edit_buffer_get_utf (&edit->buffer, edit->buffer.curs1, &char_length);
         if (char_length > 0)
-            g_snprintf (byte_str, sizeof (byte_str), "%04u 0x%03X", (unsigned) cur_utf,
-                        (unsigned) cur_utf);
+            if (cur_utf < 128)
+                g_snprintf (buf, bufsize, "%3u 0x%02X", cur_utf, cur_utf);
+            else
+            {
+                char uplus[16];
```
zyv:
```suggestion
                char uplus[bufsize];
```

zyv (commented):
Can't test.

src/editor/editdraw.c
egmontkob: I guess the problem is that my code has a VLA?

I really don't like this enum hack. How about BUF_TINY (64)?

egmontkob:
> Can't test.

What do you mean?

src/editor/editdraw.c
zyv: I didn't like that the buffer size was repeated two times.

Unfortunately, `const int` is a run-time constant, so you can't use it as a fixed array size; you have to use a macro substitution (like you did now, picking the pre-defined `TINY` buffer size) - or go for VLA arrays, which I think is a GNU extension. Or else, if you want to stay within the type system level, use an enum.

What I didn't like with the `#define` option is that it can be changed at one place and forgotten at another. They are usually used for defining fixed-size arrays, and thus only once - at other places one could then use `sizeof`.

Therefore, I suggested to keep it as you did, but pull the `bufsize` to the compile-time, all while keeping it at the type system level. This way, if the number is changed, it cannot be wrongly not updated at some place.

Thus, I think that the enum option actually is much safer and has more advantages than any other options, except for the possibly confusing way it looks.

If I had to choose between different options, I would rate them like this for the reasons above:

1. Enum option
2. Your first solution
3. The way it looks like now

zyv:
> > Can't test.
> 
> What do you mean?

I just wanted to make it explicit that I've looked at the code, but don't have the time to test it. An automated test would have been nice - actually, after you did the refactoring, the function has become testable, but, of course, it's too much to demand, especially given that it's not your code and you're fixing a problem here...

src/editor/editdraw.c
egmontkob: An earlier draft contained a `static char buf[16]` in the callee, returning its address. That way I could use `sizeof`. As discussed in the main issue, you're not in favor of this approach, e.g. it could break we ever go multithreaded. I also understand your preference of using the same pattern across the callbase: callee allocates, caller frees.

Unfortunately I'm afraid there's no solution to this problem that'd perfect address all aspects.

My problem with the `enum` approach is that it's the first time I'm ever in my life seeing `enum` being used like that, I don't think it's standard practice, and it's also against `enum`'s semantics. If we really look in this direction, and if you're not okay with using `BUF_TINY` all throughout (why not?), then another brand new `#define` at the top of this method, and `#undef` at the end, would IMO be preferable to `enum` -- at least in the sense that it's a much more typical pattern.

Alternatively, just remove all those `n`s from the function names and the passed lengths. I mean, it's pretty obvious from the format strings that we're printing roughly 10-ish bytes into a 64-byte buffer. Just allocate the buffer and not be paranoid about overrun?

src/editor/editdraw.c
```diff
@@ -95,45 +95,60 @@ printwstr (const char *s, int len)
 
 /* --------------------------------------------------------------------------------------------- */
 
-static inline void
-status_string (WEdit *edit, char *s, int w)
+static char *
+format_character_code (WEdit *edit)
```
mc-worker, line 101: Probably, there is better to create a static tiny buffer instead of allocating/deallocating it in the heap every time.

```
static const char *
format_character_code (WEdit *edit)
{
    static char buf[BUF_TINY];
```

egmontkob:
> An automated test would have been nice

I did manually test the code. Automated test: I guess I'd have to set up a mock editor widget that returns various things under the cursor, including invalid UTF-8, or setting the source codepage to some 8-bit ones. Honestly I don't think it's worth the required efforts.

src/editor/editdraw.c
egmontkob: That's what the first draft (before creating the PR) did, but I reworked as per https://github.com/MidnightCommander/mc/issues/4870#issuecomment-3521190856.

Shall I go back to that solution?

(Let's not waste too much time discussing how it's the best to allocate a buffer of 16 bytes, especially if the two main developers would prefer a different solution... we have more important things to work on.)

zyv: > Probably, there is better to create a static tiny buffer instead of allocating/deallocating it in the heap every time.

Yes, this is possible, but I explicitly spoke against this because leaking local static buffers outside makes testing and multithreading more difficult, and heap allocations cost nothing these days, especially outside of tight loops.

> (Let's not waste too much time discussing how it's the best to allocate a buffer of 16 bytes, 

I've just committed a fixup that I can live with. If everybody else can live with it, then let's move on.

> especially if the two main developers would prefer a different solution... we have more important things to work on.)

Yeah, that's the problem when there isn't one benevolent dictator that decides for everyone, for better or worse :( In your situation, you wouldn't even know whom to listen to. At least, I'm trying to provide supporting arguments for my opinions, trying to sway rather than intimidate you into the followership :)

egmontkob: > I've just committed a fixup that I can live with. If everybody else can live with it, then let's move on.

`BUF_SIZE` sounds frighteningly generic to me, doesn't it conflict with one of the system headers? If you rename to a bit more mc-specific or context-specific then I'm fine with this change.

Nicpicking: empty lines between the loads of dashes and the define/undef?

zyv: OK? 59374aa

egmontkob: 👍 thanks!

zyv (approved):

Issue #2836, opened 2012-06-28
Faster directory load in certain (specific) cases


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/2836 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `gotar` (gotar@….pl) |

<p>
If there's a need to enter directory containing thousands of entries, calling lstat() for every one of them is very expensive ...and sometimes not necessary. My suggestion:<br />
</p>
<ol class="arabiczero"><li>make use of AC_STRUCT_DIRENT_D_TYPE to check for dirent.d_type existence for the rest to work at all,
</li><li>if:
<ul><li>user has set some new option (like 'quick_cd' to be &gt;0),
</li><li>current panel view doesn't include UID/GID/mode/*time etc. (higher levels of 'quick_cd' may further narrow it down, I think quick_cd=1 should strip on size only, 2 for size and *time, and 3 ignore all the attributes),
</li></ul></li><li>then:
<ul><li>handle_dirent() for d_type==DT_DIR should skip mc_lstat(), instead put some predefined values (0s) when called from do_load_dir(),
</li><li>keep it as it is for do_reload_dir(),
</li></ul></li><li>alt-t for such partial directory listing should reload directory when required (if some not loaded attrs need to be displayed).
</li></ol><p>
This way quick_cd=1 omits lstat() for directories, when current panel view requests type/name/size - only size is lost and this is usually useless (and fast to retrieve via ctrl-space). With higher levels one can expand this behaviour up to entirely disabling lstat() for directories. Still, ctrl-r would bring all the data back.<br />
</p>

egmontkob:
#4891 / #4901 makes the recursive "Find file" operation faster by using d_type.

Here we're loading a single directory only (well, two: one for each panel), so the speedup, if any, is less noticeable.

By default we need to stat() every file that we display to show its size and timestamp. In order to skip that, we'd need a Brief or similar layout that doesn't show file details, and a similar User mini status format. Plus, we'd need to re-read the directory contents whenever the display format changes, we don't currently do that. This already means nontrivial code changes to slightly optimize a hardly-ever-performance-critical and somewhat-rare case.

> mc_lstat() might be called only for visible files

[...]

PR #4901, opened 2025-11-22
Ticket #4891: Speed up Find file by using d_type

### LINKED ISSUE #4891: Find file: use d_type for significant speedup

```shell
4.8.33, git
```

### Operating system

```shell
Linux
```

### How to reproduce

F9 -> Command -> Find File, enable Find recursively.

Specify a filename, but no content to search for.

strace reveals that every entry is `stat()`ed, in order to figure out whether we need to descend to that.

On OSes where `struct dirent` has a `d_type` field, and on filesystems where it returns the actual value rather than `DT_UKNOWN`, this `stat()` could be omitted.

In some cases (directories with many files inside) this could result in an enormous speedup.

### Expected behavior

.

### Actual behavior

.

### Additional context

Related: #2836

egmontkob:
On supporting systems, use struct dirent's d_type field to save a stat call determining the entry's type and thus whether we need to descend to that during recursive file search. This can result in massive speedup.

## Proposed changes

* Resolves: #4891

## Checklist

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -266,6 +266,8 @@ AC_STRUCT_ST_BLOCKS
 AC_CHECK_MEMBERS([struct stat.st_blksize, struct stat.st_rdev, struct stat.st_mtim, struct stat.st_mtimespec, struct stat.st_mtimensec])
 gl_STAT_SIZE
 
+AC_STRUCT_DIRENT_D_TYPE
+
 AH_TEMPLATE([sig_atomic_t],
             [/* Define to `int' if <signal.h> doesn't define.])
 AH_TEMPLATE([SIG_ATOMIC_VOLATILE_T],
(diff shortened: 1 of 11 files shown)


egmontkob:
On Linux + ext4 (or presumably many other file system types) the "Find file" operation should work just like before, except noticeably faster (especially when you're only looking for a filename, not file contents).

Presumably on mac, too (it also has `d_type`, I strongly suspect the actual file system supports it too).

I've tested on Solaris too, there `d_type` is not available, there should be no change in functionality or speed. (You can test on Linux by undef'ing `HAVE_STRUCT_DIRENT_D_TYPE` in `config.h`.)

mc-worker (approved):

zyv (approved):
This is very nice indeed! Thank you! 👍 

Is there more to #2836, or can it be closed by this PR as well?

egmontkob:
#2836 is quite different, let's keep that open for now (or maybe we can close as wontfix after a brief discussion over there).

PR #4907, opened 2025-11-25
Ticket #3160: Use new colors for listbox's selected entry

### LINKED ISSUE #3160: Dialog hotfocus color misused in listbox
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3160 |

<p>
The highlighted item in listboxes use DLG_COLOR_HOT_FOCUS if the widget is focused, and DLG_COLOR_FOCUS if the widget is not focused.<br />
</p>
<p>
This is clearly a misuse of the semantics. "Hot" is supposed to stand for hotkey (aka. shortcut key) color, a single character within a string. This feature is unused in listboxes. Using the hotkey color for the whole line makes it underlined in the "sand256" skin which was not intended there. In #2165 I'm working on a config option to use underline for hotkeys with any skin, this bug is blocking that.<br />
</p>
<p>
I'm not sure if making a distinction whether the widget is focused or not is important, I don't think so. How can a listbox widget be unfocused at all? If it can be, isn't it obvious by other means (e.g. other window overlapping it?)<br />
</p>
<p>
I think the right approach is to always use DLG_COLOR_FOCUS. (In this case there's still a confusion that what is called "focused" in the skin definiton is "selected" in the source, and "focused" in the source means something totally different.)<br />
</p>
<p>
The patch also fixes the "sand256" skin, it used to have an explicit underline for popup boxes (e.g. F2 user menu) to make it consistent with the broken listbox (e.g. Alt+H history), it's now removed, from now on underline is used solely for hotkeys in this skin.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [mc-3160-dialog-hotfocus-misuse.patch](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/3160/mc-3160-dialog-hotfocus-misuse.patch) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/3160/mc-3160-dialog-hotfocus-misuse.patch)) by `egmont` (@egmontkob) on <span title="2014-03-16T15:09:58+00:00">`Mar 16, 2014 at 15:09 UTC`</span>

egmontkob:
## Proposed changes

Introduce new colors dselnormal and dselfocus for unfocused and focused
listbox's selected entry, forked off from the here misused dfocus and
dhotfocus, respectively.

Adjust skins not the underline listbox's selected entry.

* Resolves: #3160

## Checklist


egmontkob:
This change, if accepted, will slightly break the look of 3rd party skins, they'll need to be adjusted to introduce these two new colors. This won't be the first, nor the last such occasion, since we don't have proper fallback / compatibility mechanism in place.

I updated the skins using this quick-n-dirty one-off script:

```
#!/bin/bash

while IFS='' read line; do
  echo "$line"

  if [[ $line =~ " dfocus" ]]; then
    dfocline="$line"
  elif [[ $line =~ " dhotfocus" ]]; then
    dhotfocline="$line"
  fi

  if [[ $line =~ " dhotfocus" ]]; then
    echo "${dfocline/dfocus/dselnormal}"
    echo "${dhotfocline/dhotfocus/dselfocus}"
  fi
done <$1 >$1.tmp

mv "$1.tmp" "$1"
```

egmontkob:
I did not touch the documentation, nor the supported keywords for the `--colors` cmdline option. See #4908.

mc-worker (approved):

PR #4909, opened 2025-11-25
Ticket #3158: Remove obsolete note about ncurses in the manuals

egmontkob:
Double lines now work under ncurses.

It was not even ncurses's fault, we shouldn't have blamed it.

This commit removes the obsolete and incorrect section from the manual, including all translations where it appeared.

Related to already fixed ticket #3158.

## Checklist


egmontkob:
(Double lines in ncurses don't work in KOI8-R locale, combined with ncurses versions before 6.5-20251115, due to an ncurses bug. Single lines are used instead. It's such a minor use case and will quickly vanish as distros upgrade their ncurses that I see no point in mentioning it in the manuals.)

PR #4912, opened 2025-11-25
Ticket #4911: menu padding

### LINKED ISSUE #4911: Uneven padding in menus

```shell
4.8.33, git
```

### Operating system

```shell
.
```

### How to reproduce

Image

Image

... also, if the menu needs to be pushed to the left because otherwise it would overflow to the right from a narrow terminal window, then the accepted mouse click x coordinates are off by 1.

### Expected behavior

A right padding of 1, matching the left padding and looking visually nice.

### Actual behavior

Right padding is sometimes 0, sometimes 2.

### Additional context

_No response_

egmontkob:
## Proposed changes

F9 menu:
- Remove the unused (and not Unicode compatible) first letter feature
- Fix the amount of right padding
- Fix horizontal mouse coordinates

* Resolves: #4911 

## Checklist

### DIFF
--- a/lib/widget/menu.c
+++ b/lib/widget/menu.c
@@ -741,10 +741,10 @@ menubar_mouse_on_menu (const WMenuBar *menubar, int y, int x)
     menu = MENU (g_list_nth_data (menubar->menu, menubar->current));
     left_x = menu->start_x;
     right_x = left_x + menu->max_entry_len + 2;
-    if (right_x > w->cols)
+    if (right_x > w->cols - 1)
     {
-        left_x = w->cols - (menu->max_entry_len + 2);
-        right_x = w->cols;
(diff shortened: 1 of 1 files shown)


egmontkob:
Shall I maybe keep the "first letter" feature, just in case we wish to use it one day?

Making it Unicode compatible to show a tick symbol is as easy as changing an unsigned char to gunichar.

mc-worker:
> Shall I maybe keep the "first letter" feature, just in case we wish to use it one day?

Yes, I think this feature should be kept, for example, to show menu item as checked/unchecked.

egmontkob:
> Yes, I think this feature should be kept

Done.

mc-worker (approved):

PR #4916, opened 2025-11-28
Ticket #4915: Fix the help window's title color

### LINKED ISSUE #4915: master: Help's title color broken

```shell
4.8.33-496-ge99555c35
```

### Operating system

```shell
.
```

### How to reproduce

Brand new commit e99555c35, part of PR #4907, broke the help page's title in all skins.

The dialog's title word "Help" is displayed with the terminal's default colors.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
This broke in recent commit e99555c35.

## Proposed changes

Use named indices rather than just listing the struct items which can easily go out of sync.

For consistency, moved the code to where other similar structs are initialized.

* Resolves: #4915

## Checklist

### DIFF
--- a/lib/widget/dialog.c
+++ b/lib/widget/dialog.c
@@ -54,6 +54,7 @@
 dlg_colors_t dialog_colors;
 dlg_colors_t alarm_colors;
 dlg_colors_t listbox_colors;
+dlg_colors_t help_colors;
 
 /* A hook list for idle events */
 hook_t *idle_hook = NULL;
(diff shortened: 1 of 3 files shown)


mc-worker:
`help_colors` is still used in the `help_interactive_display()` function only. Why did you introduce a global variable instead of fixing the local one?

egmontkob:
`listbox_colors` is also only used at one place, yet is defined globally. Thus moving `help_colors` here makes the code more consistent: fewer different patterns seen at how various `dlg_colors_t` structures are defined.

The old code layout here is obviously pretty fragile; if I simply added a few new members to the current assignment then it could easily break later on, just as it did now. So I definitely wanted to have a syntax where each value is clearly assigned to its proper and explicitly spelled out index. Keeping `help_colors`'s assignment at its current location would only have been possible using runtime (i.e. non-const) assignment which is run over and over again, something that seems to me isn't really liked by mc's current coding practice. `help.c` doesn't have an `init()` method either where I could have convenienently moved this initialization to.

Does this make sense? If you don't like the proposed design then please give clear instructions what would be your preferred solution.

mc-worker (approved):

PR #4932, opened 2025-12-08
4931 skin frame colors

### LINKED ISSUE #4931: Skin: separate colors for frames

It was brought up at https://github.com/MidnightCommander/mc/pull/4813#issuecomment-3558216256 and followup comments that it might make sense to extract the frame colors to possibly make them different from the box's main contents color.

I think this was also brought up many-many years ago, maybe during the design / discussion of some skin, but I can't find it.

Norton / Volkov / Far have a truly interesing vibe by using bright cyan foreground color on their main page, rather than white as Midnight does. But also using bright cyan for the contents (regular file names, the file's contents in the viewer etc.) is personally too much for me. So, switching to bright cyan frames (on top of the blue background), but keeping the main text as white, could be an interesting choice for one of our default-like skins.

I expect that several skin authors would also like to have this possibility and would be happy to play with new frame colors.

sand256.ini uses italic text in the F9 dropdown menu. It's also recently been changed to use rounded corners and dashed lines. In xterm, using some of the fonts it offers by default in its ctrl+right-click menu, these don't show up correctly for me; the italic font doesn't define them. Being able to unset the italic property fixes the look.

PR coming in a sec.

The addition of the new keywords is a backwards compatible change. Even though I add the new keywords to the skin files for convenience, always simply repeating the `_default_` value of its section (except at that one place in sand256 where I remove the italic property), if it is missing (as it would in some 3rd party skins that don't get updated) then they're automatically taken from `_default_` anyways, resulting the same unchanged look.

I'm not yet introducing a default-like skin with a cyan frame, or any other skin where the frame has a different color. This is yet to be done, possibly by someone else.

.

_No response_

### Additional context

_No response_

egmontkob:
## Proposed changes

- Use more consistent internal constant names for colors
- New skin colors for frames
- sand256 skin: Use non-italic frame characters

* Resolves: #4931

## Checklist

### DIFF
--- a/lib/filehighlight/get-color.c
+++ b/lib/filehighlight/get-color.c
@@ -282,7 +282,7 @@ mc_fhl_get_color (const mc_fhl_t *fhl, const file_entry_t *fe)
     int ret;
 
     if (fhl == NULL)
-        return NORMAL_COLOR;
+        return CORE_NORMAL_COLOR;
 
     for (i = 0; i < fhl->filters->len; i++)
     {
@@ -306,7 +306,7 @@ mc_fhl_get_color (const mc_fhl_t *fhl, const file_entry_t *fe)
             break;
         }
     }
-    return NORMAL_COLOR;
+    return CORE_NORMAL_COLOR;
 }
 
 /* --------------------------------------------------------------------------------------------- */
--- a/lib/skin.h
+++ b/lib/skin.h
@@ -23,94 +23,101 @@
 #define CORE_COMMAND_MARK_COLOR    mc_skin_color__cache[7]
 #define CORE_HEADER_COLOR          mc_skin_color__cache[8]
 #define CORE_SHADOW_COLOR          mc_skin_color__cache[9]
+#define CORE_FRAME_COLOR           mc_skin_color__cache[10]
 
 /* Dialog colors */
-#define DIALOG_NORMAL_COLOR          mc_skin_color__cache[10]
-#define DIALOG_FOCUS_COLOR           mc_skin_color__cache[11]
-#define DIALOG_HOT_NORMAL_COLOR      mc_skin_color__cache[12]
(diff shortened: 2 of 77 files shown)


egmontkob:
Here's the makeshift script I used to update the skin files: [mc-4931-update-skins.sh](https://github.com/user-attachments/files/24036503/mc-4931-update-skins.sh)

egmontkob:
Since the PR does not alter the look of any of the skins (except for that tiny italic fix), you'd have to create your own skin to test the change. Something like replacing all `*frame = *` with `*frame = brightcyan;` for testing purposes.

mc-worker (approved):

egmontkob:
[frame.ini.txt](https://github.com/user-attachments/files/24164235/frame.ini.txt)

Here's a slight variation of the default skin, using differently colored frames.

Place in `~/.local/share/mc/skins/frame.ini` (remove the `.txt` extension) and launch `mc -S frame`.

It's for demo purposes only, I don't mean to include this in mc, but hopefully some cool skins using this feature will emerge soon.

egmontkob:
The ad-hoc script didn't take into account the possibility of a section not having a `_default_` line, and thus broke the viewer's frame colors (visible in "Quick view" mode) in the `seasons-*` skins.

PR #4945 will fix this.

zyv:
Happy to see internal naming refactorings! 👍 

What about the skin editor though? Do you think you could update that as well?

egmontkob:
I'd love to do a big revamp of the skin editor! :)

zyv:
> I'd love to do a big revamp of the skin editor! :)

I'd be the last one to dissuade you from this, but I think you've already found (and fixed) some small issues, and until now it wasn't committed because you wanted to do a bigger PR :) So maybe committing several small fixes is better than waiting for a one-big revamp...

Issue #4789, opened 2025-10-15
Slow (10s) startup when shell startup file flushes TTY


krobelus:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Midnight Commander version and build configuration

```shell
$ mc -V
GNU Midnight Commander 4.8.33-365-g05a6cf97e
```

### Operating system

```shell
Linux
```

### Is this issue reproducible using the latest version of Midnight Commander?

- [x] I confirm the issue is still reproducible with the latest version of Midnight Commander

### How to reproduce

Run this shell script

```
gcc -xc - <<'EOF'
#include <assert.h>
#include <termios.h>
#include <fcntl.h>
#include <unistd.h>

int
main (void)
{
    int ttyfd = STDOUT_FILENO;
    assert (isatty (ttyfd));
    int ftty = open ("/dev/tty", O_RDWR | O_NOCTTY | O_CLOEXEC);
    assert (ftty >= 0);
    static struct termios oldTerm;
    int ok;
    ok = tcgetattr (ftty, &oldTerm) != -1;
    assert (ok);
    ok = tcsetattr (ftty, TCSAFLUSH, &oldTerm) != -1;
    assert (ok);
}
EOF
echo '[ -n "$MC_SID" ] && '"$PWD"'/a.out' >> ~/.bashrc
SHELL=$(which bash) mc
```

### Expected behavior

`mc` should start quickly

### Actual behavior

`mc` hangs for 10 seconds, apparently because the `./a.out` process has flushed the TTY, so mc can't read the prompt.

### Additional context

Originally reported in https://github.com/MidnightCommander/mc/discussions/4766
This happens on Garuda Linux which configures its shell (fish) to run the "fastfetch" program on startup (if interactive).
That program flushes the TTY like above in a function called `ffDetectTerminalSize`, called when they want to output an image to the TTY.
I'm not yet sure if this could be fixed in "fastfetch" rather than mc.
cc @egmontkob

egmontkob:
I can't see a clean solution to this problem.

When the shell starts up, mc quickly types some characters that initialize the shell for mc's purposes. These run in the shell's interactive mode, after the shell executes its ini file(s).

We don't know when the shell finishes with its ini file(s) and is about to continue with the letters mc typed; we're just about to install our hacks that would allow us to know.

---

There's one hack I can think of.

[...]

Issue #4253, opened 2021-05-30
Interference of injected subshell commands output with e.g. read in .zshrc causing 10s startup delay


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4253 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `marlonrichert` (marlon.richert@….com) |
> | **Mentions** | `baltazar.bz@….com` |
> | **Keywords** | `subshell`, `common.c`, `init_subshell`, `zsh`, `.zshrc` |

<h2 id="Stepstoreproduce">Steps to reproduce</h2>
<div class="code"><pre>$ <span class="nb">cd</span> <span class="k">$(</span>mktemp -d<span class="k">)</span><span class="p">;</span> <span class="nb">exec</span> env -i <span class="nv">HOME</span><span class="o">=</span>$PWD <span class="nv">TERM</span><span class="o">=</span>$TERM zsh -f 
% print -r <span class="s2">"typeset -F SECONDS
print -nr \$'\e[6n'  # We query the terminal for info.
IFS=\$'\e[;' read -Asd R -- reply  # We read the terminal's reply.
print \$SECONDS
"</span> &gt; .zshrc
% <span class="nb">exec</span> zsh  <span class="c1"># Starting zsh is still instantaneous.
</span>0.0557880000
% <span class="nv">SECONDS</span><span class="o">=</span>0<span class="p">;</span> mc<span class="p">;</span> print $SECONDS  <span class="c1"># Starting mc is now really slow.
</span>
11.5728380000
%
</pre></div><p>
The problem is here: <a class="ext-link" href="https://github.com/MidnightCommander/mc/blob/a88a626e76139259e5b6fc0db39045f051e243dd/src/subshell/common.c#L1378">https://github.com/MidnightCommander/mc/blob/a88a626e76139259e5b6fc0db39045f051e243dd/src/subshell/common.c#L1378</a><br />
</p>
<p>
You print commands to the shell as if it were user input or terminal input. This interferes with the <tt>read</tt> statement in the <tt>.zshrc</tt> file: The terminal's reply is now buried inside the characters you are printing. As a result, the <tt>read</tt> statement hangs until it times out and most of the commands you print don't get executed.<br />
</p>
<h2 id="Expectedbehavior">Expected behavior</h2>
<p>
[...]

Issue #4812, opened 2025-10-23
Persistent command buffer not working on Solaris, 1 sec startup delay


egmontkob:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Midnight Commander version and build configuration

```shell
git
```

### Operating system

```shell
SunOS 5.11 11.4.85.201.2 sun4v sparc sun4v logical-domain
```

### Is this issue reproducible using the latest version of Midnight Commander?

- [x] I confirm the issue is still reproducible with the latest version of Midnight Commander

### How to reproduce

Let's move this to its own issue.

Just start up mc with bash 5.2.37 as subshell. Notice 1 second delay (while testing the persistent command feature – unsuccessfully). The feature then doesn't work.

First round of investigation at https://github.com/MidnightCommander/mc/issues/4480#issuecomment-3399128449 suggests a Solaris bug:

> Getting one step closer to why the persistent command feature doesn't work on Solaris + bash:
> 
> Outside of mc, just in plain bash, execute a sleep 10 and then type ahead foo (do not press Enter).
> 
> When those 10 seconds elapse, on other platforms the shell prompt is printed and is immediately followed by the foo string.
> 
> On Solaris, the shell prompt is printed and it isn't followed by anything. Press one more key and then they suddenly all appear.
> 
> The very same thing happens if it's ESC _ or ESC + instead of foo. bind some action just like in mc's source, preferably some echo with immediate effect. While a command is running, press these keys. They don't get executed as soon as the command finishes, they only get executed after yet another keypress.
> 
> So, somehow when the tty line is switched from cooked to raw mode, the pending bytes in the cooked buffer aren't forwarded to the slave app immediately?
> 
> Interestingly though, it works as expected on zsh. So there's more to this story.

For some reason, sending an Enter after ESC _ or ESC + doesn't fix the issue, I'm not sure why.

### Expected behavior

.

### Actual behavior

.

egmontkob:
A possible workaround:

As per Ossi's suggestion, somehow get to know when the prompt is printed. This could be done by appending some `MC_MAGIC_COOKIE` at the very end of the prompt, and looking for it (and stripping it) when parsing the prompt from the subshell.

Plus, rely on the totally undocumented behavior of bash that it first switches the tty to raw mode, and then prints the prompt. Because if it decides to do it the other way around then there'll still be a race condition.

---

[...]

PR #4934, opened 2025-12-15
4625 slow zsh macos

### LINKED ISSUE #4548: Mc start up is slow on FreeBSD with Bash shell when mc is using a subshell

### LINKED ISSUE #4625: mc 4.8.32 slow start on MacOS

### LINKED ISSUE #4723: mc ignores $PS1 on macOS

```shell
GNU Midnight Commander 4.8.33
Built with GLib 2.82.4
Built with libssh2 1.11.1
With builtin editor
With mouse support on xterm
Virtual File Systems:
Data types:
```

### Operating system

```shell
Darwin cat.andromeda.lan 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:25 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6020 arm64

(MacOS 15.5)
```

### How to reproduce

When I start `mc`, it ignores the `PS1` variable I had exported, and resets the prompt to `bash-3.2$`.

This is on MacOS 15.5, with the default shell set to Bash (v. 3.2.57 on Mac)

In `.profile`, I set PS1 as:

    export PS1='\[\e]2; $PWD \a\e[32;1m\]\h:$PWD/$\[\e[0m\] '

But when I start mc, I have:

```
bash-3.2$ echo $PS1
\s-\v\$
bash-3.2$ 
```
I must re-execute the `export PS1=`... command to get back my prompt (though without the colors, but that is ok).

### Expected behavior

The mc subshell should keep the value of `$PS1`

### Actual behavior

`$PS1` gets replaced.

### Additional context

I don't have this problem on a different machine with MacOS 15.5 but with Midnight Commander version 4.8.28

### LINKED ISSUE #4780: Read from pty master in nonblocking mode (due to unlikely deadlock)

```shell
4.8.33-353-g32b7a14c4
```

### Operating system

```shell
any
```

### How to reproduce

`src/subshell/common.c` contains the following typical pattern 3 times (#4625 is about to add 2 more):

- `select()` on `mc_global.tty.subshell_pty`, possibly along with other fds
- if other fds were present then check if `mc_global.tty.subshell_pty` is said to contain data
- `read()` from this fd

With tty lines, this standard pattern of reading can cause a lockup. That's because the slave side can perform a `tcflush()`, i.e. revoke the data already placed in the tty line. And this could happen between the `select()` and the blocking `read()` steps.

We should set the master fd to `O_NONBLOCK` (a.k.a. `O_NDELAY`) mode. Either for the duration of the `read()`s only, or once and for all but then make sure to adjust the `write()` / `write_all()` calls to expect that attempting to write to a clogged channel behaves differently and returns immediately (`write_all()` would need to be a `select()`+`write()` loop rather than just a `write()` loop).

I guess a helper `read_nonblock()` method flipping the flag back and forth is the easiest solution for now, until proper glib mainloop-based approach is implemented.

Linux has `preadv2()` that allows to pass flags on a per-call basis (for some reason it uses `RWF_NOWAIT` instead of `O_NONBLOCK`) but it's super non-portable.

### Expected behavior

never to block

### Actual behavior

super unlikely to block

### Additional context

_No response_

egmontkob:
## Proposed changes

Fix slow startup on macOS with zsh, and other related problems also affecting other platforms and shells.

* Resolves: #4625
* Resolves: #4780
* Resolves: #4548
* Resolves: #4723

## Checklist

---

This is a reincarnation of closed PR #4784. The patches are essentially unchanged (rebased and a few minor feedback addressed).

I really wanted to address #4253 / #4789 first, i.e. to inject our commands via shell startup files rather than "typing" them to those shells; and rebase this work on top of that. However, as discussed in 4789, I failed to figure out how to do this in `zsh`, it looks like it's not possible. This made me lose motivation and I've abandoned that work, at least for now.

The work in this PR is quite hackish, but so is the problem we're trying to deal with, and the current quite complex codebase. And, unfortunately, I still can't claim I perfectly understand everything, including the suspicious 2 or 3 different ways of handling of sending the Ctrl-C char. It's surely nowhere near the cleanest of my works I've ever made, but honestly I got tired of not seeing a clear solution. It's just not the kind of area I'm comfortable with.

So, IMO, let's land this -- along with remotely related pending PR #4804 --, both have already gotten approval a long time ago but then still the discussion and possible improvements went on. I'm afraid this very PR is not really going to get substantially better, and lack of progress for more than a month clearly shows that I've lost motivation. I suggest that we land these two after a last round of review, and then later (also depending on possible user feedback) iterate further.

### DIFF
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -263,6 +263,31 @@ write_all (int fd, const void *buf, size_t count)
 
 /* --------------------------------------------------------------------------------------------- */
 /**
+ *  Read in nonblocking mode.
+ *
+ *  On a tty master, waiting for data using a select() and then reading it with a blocking read()
+ *  can cause a lockup. That's because between these two steps the slave side can do a tcflush(),
+ *  revoking the data it sent earlier.
+ *
+ *  Reminder for the caller: if no data is available, but data might arrive later, this returns -1
+ *  and errno is set to EAGAIN or EWOULDBLOCK (these two may or may not have the same value).
+ *  Return value 0 means end of stream.
+ */
+
+static ssize_t
+read_nonblock (int fd, void *buf, size_t count)
+{
+    const int old_flags = fcntl (fd, F_GETFL);
+
+    fcntl (fd, F_SETFL, old_flags | O_NONBLOCK);
+    const ssize_t ret = read (fd, buf, count);
+    fcntl (fd, F_SETFL, old_flags);
+
+    return ret;
+}
+
+/* --------------------------------------------------------------------------------------------- */
+/**
  *  Prepare child process to running the shell and run it.
  *
  *  Modifies the global variables (in the child process only):
@@ -896,11 +921,17 @@ feed_subshell (int how, gboolean fail_on_error)
         // for (i=0; i<5; ++i)  * FIXME -- experimental
         {
             const ssize_t bytes =
-                read (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
+                read_nonblock (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
 
-            // The subshell has died
-            if (bytes == -1 && errno == EIO && !subshell_alive)
-                return FALSE;
+            if (bytes == -1)
+            {
+                if (errno == EAGAIN || errno == EWOULDBLOCK)
+                    continue;
+
+                if (errno == EIO && !subshell_alive)
+                    // The subshell has died
+                    return FALSE;
+            }
 
             if (bytes <= 0)
             {
@@ -1896,10 +1927,11 @@ flush_subshell (int max_wait_length, int how)
         timeleft.tv_sec = 0;
         timeleft.tv_usec = 0;
 
-        const ssize_t bytes = read (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
+        const ssize_t bytes =
+            read_nonblock (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
 
         // FIXME: what about bytes <= 0?
-        if (how == VISIBLY)
+        if (bytes > 0 && how == VISIBLY)
             write_all (STDOUT_FILENO, pty_buffer, (size_t) bytes);
     }
 
@@ -1941,10 +1973,13 @@ read_subshell_prompt (void)
             exit (EXIT_FAILURE);
         }
 
-        bytes = read (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
+        bytes = read_nonblock (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
 
-        parse_subshell_prompt_string (pty_buffer, bytes);
-        got_new_prompt = TRUE;
+        if (bytes > 0)
+        {
+            parse_subshell_prompt_string (pty_buffer, bytes);
+            got_new_prompt = TRUE;
+        }
     }
 
     if (got_new_prompt)
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -1236,9 +1236,17 @@ pty_open_slave (const char *pty_name)
 /**
  * Set up `precmd' or equivalent for reading the subshell's CWD.
  *
- * Attention! Never forget that these are *one-liners* even though the concatenated
- * substrings contain line breaks and indentation for better understanding of the
- * shell code. It is vital that each one-liner ends with a line feed character ("\n" ).
+ * Attention!
+ *
+ * Physical lines sent to the shell must not be longer than 256 characters, because above that size
+ * on some platforms the kernel's tty driver in cooked mode begins to lose characters (#4480).
+ *
+ * However, it's preferable to send one logical line to the shell, to prevent the pre-prompt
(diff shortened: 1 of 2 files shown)


src/subshell/common.c
```diff
@@ -1781,10 +1812,11 @@ flush_subshell (int max_wait_length, int how)
         timeleft.tv_sec = 0;
         timeleft.tv_usec = 0;
 
-        const ssize_t bytes = read (mc_global.tty.subshell_pty, pty_buffer, sizeof (pty_buffer));
+        const ssize_t bytes =
```
ossilator, line 1974: nit: unrelated re-wrap.

src/subshell/common.c
```diff
@@ -1427,12 +1431,15 @@ do_subshell_chdir (const vfs_path_t *vpath, gboolean force, gboolean update_prom
      * the command prompt before we send the cd command. */
     if (!use_persistent_buffer)
     {
-        write_all (mc_global.tty.subshell_pty, "\003", 1);
-        subshell_state = RUNNING_COMMAND;
-        if (mc_global.shell->type != SHELL_FISH && !feed_subshell (QUIETLY, TRUE))
+        if (subshell_initialized)
```
ossilator: nit: merging the conditional would avoid nesting and the re-indentation.

ossilator (commented):
good enough for starters.

it's entirely plausible that there is no _right_ solution for this; after all, mc is operating outside anything normally expected for interactive shells.

src/subshell/common.c
egmontkob: `make indent` does these for me.

src/subshell/common.c
egmontkob: This is one of the major points where I don't understand the code's intent: `write_all()` is executed even for fish, but then setting `subshell_active` and returning isn't; whereas at the two other cases where we send Ctrl+C `write_all()` is inside the non-fish branch. This is pretty fishy (pun intended) but unfortunately I just don't feel the motivation to figure it all out (understand, and fix if necessary). I just wanted to touch the code as little as possible.

As far as I can see, merging the two nested conditionals would definitely modify the business logic -- I can't tell if for the better or for the worse.

I'm happy to file a new issue about this fishy situation, to be reviewed and possibly fixed one day.

ossilator: i mean with the outer conditional. that's logically just an AND.

egmontkob: Yeah, sorry, I missed that. Will do in a sec.

zyv (approved):
I'm slowly working through my immense mail backlog and have now made it to this PR... I've looked at all the commits separately and at the code as a whole once again.

To me it looks great now! I can particularly commend @egmontkob for splitting changes into sensible commits with helpful commit messages, and the comments in the code are also really useful.

Not even speaking of all the research and investigation that went into getting to the bottom of it. This is really a remarkable tour de force!

In as far as I'm concerned, I'd like to see it merged now. Okay with you @mc-worker ?

mc-worker (approved):

egmontkob:
Let's see how it goes...

I'm merging it now, and will give it another round of testing on various OSes in a couple of days (please remind me :))

egmontkob:
Just for the record:

A small startup delay on Solaris, #4812, still isn't fixed. There we'd need to somehow work around an obvious Solaris bug.

zyv:
> Let's see how it goes...
> 
> I'm merging it now, and will give it another round of testing on various OSes in a couple of days (please remind me :))

My zsh startup delay on macOS is gone ❤️ 

```
~ % mc -V   
GNU Midnight Commander 4.8.33-535-g045eab71e
~ % zsh --version
zsh 5.9 (arm64-apple-darwin24.0)
~ % uname -a
Darwin Mac 24.6.0 Darwin Kernel Version 24.6.0: Wed Nov  5 21:33:32 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T6030 arm64
```

PR #4945, opened 2026-01-06
4896 skin readme

### LINKED ISSUE #3726: Fix doc about 'default' fg/bg color
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3726 |

<p>
The manual page says:<br />
</p>
<p>
"The possible colors are: [...] And there is a special keyword for transparent background. It is 'default'. The 'default' can only be used for background color."<br />
</p>
<p>
Apparently this restriction is not true, the word 'default' can stand either at the foreground or at the background color position (or even both), meaning the default fg or bg color of the terminal emulator, respectively.<br />
</p>
<p>
We should rephrase the above sentence (which would probably imply omitting the word "transparent" and go for the "terminal emulator's default" instead).<br />
</p>
<p>
(We should double check that it works in ncurses too, I've checked with slang.)<br />
</p>
<p>
This is useful e.g. in order to say this in a skin file:<br />
</p>

```
[core]
    commandlinemark = default;default;reverse
```

### LINKED ISSUE #4885: Wrong comment about 256 colors in truecolor skins

```shell
git
```

### Operating system

```shell
.
```

### How to reproduce

@horkykuba Your commit cd5fc97a86035c49291cfdafa6f5889c02e64f6e changed the comments in all the truecolor skins:

```
-# In addition to these the 256 palette colors are still available, and
-# obviously so are the attributes. See the sand256.ini skin file for details.
+# In addition to these, you can use the basic 8 (non-light) colors and
+# the attributes. However, you can't use the 256 indexed colors.
```

I think this is a leftover for an earlier iteration of your change, correct? You removed this restriction, even for `xterm-direct` and `xterm-direct16` by manually converting to their typical truecolor values, whereas with `xterm-direct256` the 256-indexed colors are used directly.

I've tested, placing a value like `rgb345` there works for me both with ncurses and any of `TERM=xterm-direct`, `...-direct16` and `...-direct256`, as well as with slang + `TERM=xterm-256color` + `COLORTERM=truecolor`.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

### LINKED ISSUE #4896: Revise skin instructions, move to separate README

Currently the colors that can be used in skins, as well as how inheritance works, are described in `sand256.ini`, the first 256-color skin.

Now that we have many 256-color skins, it's hard to find this info.

It should be moved to a separate `README` file which also gets installed to the skin directory. That's how people can easily find it. Or to the manual page?

While at it, do some further cleanups and fixes:
- mention that from `[whatever-section]/_default_` we further fall back to `[core]/_default_`.
- merge truecolor syntax description from the `seasons*` skins.
- #4885
- mention the `256colors` and `truecolors` keywords
- remove sentences referring to ancient versions, like sand256: needs ncurses or slang from 10 years ago or newer; modar*: needs mc 4.8 or newer
- mention that the file needs to be encoded in UTF-8 (all those `*-char` definitions)
- mention that currently only narrow (single charcell) characters are supported for `*-char`
- point to the online skin viewer, to make it easier to discover

@horkykuba as per #4885 you have ongoing work hereabouts, I don't want to conflict with that so I'll work on this once you're done with that. Or, if you feel like taking on any of these, I won't hold you back :)

.

_No response_

### Additional context

_No response_

egmontkob:
## Proposed changes

- Fix frame colors
- Remove some obsolete comments, fix some typos
- Move documentation to a separate file

* Resolves: #4885
* Resolves: #4896
* Resolves: #3726

## Checklist


misc/skins/README.txt
```diff
@@ -0,0 +1,109 @@
+This is an overview of the color capabilities of skin files.
```
zyv: Is it a good idea to call it `README_SKINS.txt` ? In other places, we call those `README.txt` or `README.md`.

misc/skins/README.txt
```diff
@@ -0,0 +1,109 @@
...
+
+See the FAQ for instructions on setting up 256 color or truecolor support in
```
zyv, line 3: There is some description of skins in the manual page:

https://source.midnight-commander.org/man/mc.html#Skins

This might be quite outdated (after tickets like #4932) and/or incorrect.

Would you have the motivation to glance over it and add obviously missing easy stuff, and maybe a reference to this file?

Also, in this file, you could add a reverse reference to check the manual page for the basic documentation.

I think one could argue for this content to go into the manual page, but it feels way too detailed and annoying to typeset, so probably this README is the best place for it.

egmontkob: Indeed, that part of the manual (which I didn't recall existed at all) is super outdated. On the other hand, it talks about the keywords too, which is missing from my README.

We should indeed unify them: keep and update one, and throw away the other.

I would also prefer to further update the README (add the missing keywords), and replace the existing part of the manual with just a brief pointer. Reasons:

- Easier, quicker typesetting (as you said), i.e. less work for us.
- We're more likely to keep it updated as skin capabilities evolve.
- We don't put pressure on translators to translate, should anyone decide to translate the manual to a new language or update one of the outdated translations; in the mean time, keeping the manual short in general might encourage others to translate to new languages. 
- It's borderline between user and developer documentation. We already have similar docs outside of the manpage, e.g. in the FAQ.

I'll do it.

misc/skins/README.txt
ossilator: yeah, the name is redundant with the path. i wouldn't do that, either.
it would be different if the file got installed into the docdir instead of along with the skins, though then an install-time rename would kind of make sense.

egmontkob: Fair enough, I'll rename to `README.txt`.

Rant:

Despite all my hatred towards Markdown, I started with `.md`. Then I got stuck at the two-column table listing the 8*2 colors.

Markdown doesn't support headerless tables, tables need to have a header, even if I don't want to. Okay, I made up some header. Then I converted to HTML using a random command I found on the web. The table didn't work. Then I tried another, completely different command I found. Didn't work either. Third command. Nope. Then I read 3 people recommending 3 different extensions to load to that markdown formatter, none of which worked. (They didn't work with tables copy-pasted from Markdown examples either, it's not that I made a stupid typo.)

The idea that the _source_ is already beautifully formatted, and it is also the source of conversion to another format which is even more beautifully formatted, is utterly braindamaged to begin with.

There's no unified specification, everyone has their own extensions. There's not even a bare minimum, including tables, that would work in all reasonable converters. Everyone does their all random things which barely work. (One of my "favorites", deliberately not linking not to have a backlink, is github jamiemcg Remarkable 207: they got the _very basics_ of line / paragraph wrapping wrong and it took them 2 years to fix.)

Another aspect that I hate is that pretty much all characters have a special meaning, and there's no consistent way of escaping, or often there's no way of escaping at all. E.g. inside backticks you get to print some literal text; but what if you want to print a backtick in that way? You're out of luck. What if you want to have a paragraph that begins with "1.", without converting to a bullet point? You're again out of luck. This list goes on and on and on. It's a giant pile of ad-hoc solutions that work together very badly.

If the source file is tied to a particular formatter (e.g. appearing on github's page, or part of mc's make procedure converting it to an html that we install), _and_ you didn't want to do any "unusual" formatting (e.g. as per my previous paragraph), then Markdown, despite being terribly designed, is at least practical.

If the source just sits there without being converted, or left for the user to be converted (and who's going to test and tell the user which converter to use that works properly on that file?) then Markdown is not only terribly designed but is also terribly impractical. If the file just sits there with no one attempting to convert it to html, there's no reason to attempt to be "valid" Markdown (which concept doesn't exist), just go with whatever contents you want to and call it `.txt`.

My 2 cents on Markdown :)

ossilator: yeah, markdown is crap. ReStructuredText is better.

zyv: > My 2 cents on Markdown :)

Enjoy some roff and the Markdown pain will subside ;-)

egmontkob: My recent roff experientes weren't _that_ bad. Mind you, I didn't stress-test its capabilities :)

Actually, now that I started working on merging the manpage and the README here, I might change my mind. Having access to colors, and auto-formatting to the terminal width, are definitely advantages for the user.

Anyway, let me have the contents first. Formatting will follow.

egmontkob:
How does this look like?

I haven't done a rigorous self-review, so surely there are some things to be cleaned up.

As for formatting: I'm not totally against converting to manual page. Having colors, and auto-formatting to the terminal width could be an advantage to the user. On the other side, required efforts put aside: Currently the txt file is 500+ lines long. The formatted man page is 3000+ at 80 columns, and formatting this document manpage-style would add considerable amount of of left margin, resulting in probably an additional ~800 lines in the manual. I'm tempted to say it's too much to be included there.

zyv:
I just found this issue: #3726 - I guess it's resolved here as well, right?

zyv (approved):
I read through the text, and nothing jumped out to me, although I have to admit that I couldn't concentrate on every sentence. It would be great if someone else could also read through it to make sure nothing is missed. The general structure and formatting also look good to me.

I only hope that we will manage to keep it up to date in the future. My general strategy in these cases is always to try to refer to the code when possible. Here, I don't have great ideas on how to reduce the amount of the documentation this way, if we opt to document it at all.

egmontkob:
> I just found this issue: https://github.com/MidnightCommander/mc/issues/3726 - I guess it's resolved here as well, right?

I've added a fix for that now. It also reminded me that I forgot to document the color value "default", added now.

egmontkob:
Also added documentation about the "base" color, but see #4951.

misc/skins/Makefile.am
```diff
@@ -1,6 +1,7 @@
 skindir = $(pkgdatadir)/skins
 
 skin_DATA = \
+	README.txt \
```
mc-worker: `skin_DATA` is a list of skins. I think `README.txt` should be moved to `EXTRA_DIST`.

egmontkob: If I move it there then a `make install` doesn't install that file. I want it to be installed next to the skins since it's important information for users who don't look at the source (and the man page also refers to it).

How to do that if it's added to EXTRA_DIST? (I'm _really_ not familiar with automake.)

zyv: Well, according to my understanding, autotools **require** you to add installable files to `DATA`.

So if the file is to be installable in the **same** directory, but not mixed with the list of skins, then the whole block has to be duplicated:

```Makefile
skinmiscdir = $(pkgdatadir)/skins
skinmisc_DATA = README.txt
```

To me, this doesn't make sense. An alternative could be as follows:


```Makefile
skin_DATA += README.txt
```

@mc-worker, how do you want it?

egmontkob: As for the build and installation procedure (which automake is about), there's no difference between `README.txt` and `whateverskin.ini`. So I'm wondering: what would be the point in artificially separating them into two groups?

If we really want to separate, could we do the separation the other way around? Roughly:

```
# a variable name with no special meaning to autotools
skinlist = [the list of skins]

skin_DATA = README.txt $skinlist
```

mc-worker: > how do you want it?

```
 EXTRA_DIST = \
-    $(skin_DATA)
+    $(skin_DATA) \
+    README.txt
```

egmontkob: Pushed a new version according to my latest comments. Please take a look, I think it looks nicely.

I'm yet to address #4951 in this PR, too.

egmontkob: > > how do you want it?
> 
> ```
>  EXTRA_DIST = \
> -    $(skin_DATA)
> +    $(skin_DATA) \
> +    README.txt
> ```

That's what I tried first, but it didn't install README.txt.

mc-worker: Well, seems I'm wrong. Files in `EXTRA_DIST` are not installed automatically,

mc-worker (approved):

misc/skins/Makefile.am
zyv: > If we really want to separate, could we do the separation the other way around? Roughly:
> 
> ```
> # a variable name with no special meaning to autotools
> skinlist = [the list of skins]
> 
> skin_DATA = README.txt $skinlist
> ```

Yes, this is how we do it in other places. Sorry, I didn't think about it immediately.

egmontkob:
Removed the "base" color from the new documentation.

Will merge this PR now, and address #4951 in a new one.

PR #4949, opened 2026-01-11
Ticket #1808: replace `hiddenfiles-show-char` in featured skins with middot

### LINKED ISSUE #1808: Featured skin: hiddenfiles-sign-show not showing in Putty
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/1808 |
> | **Reporter** | `dmartina` (dhmartina@….es) |

<p>
The centred dot (Unicode 0x22C5 I think), appears in my Putty under WXP as a square (the default for a missing character). I found this is because of the character missing in the Courier New standard font. I tried some other TTF and most of them didn't support this dot (Deja Vu Sans does). The thick dot for hiddenfiles-sign-hide seems to be a better choice and is always supported.<br />
Can we change in the featured.ini this dot in behalf of the common one, 0x0B7 (in my Spanish keyboard it is in Shift-3)?<br />
BTW, when doing these "experiments" I noticed that it was possible to put there a two character string (I tried both good and bad dots there) and the mouse input was confused by it. Maybe these strings should be checked at runtime and just pick the first unicode character of the string. Anyway, I'm afraid that mouse input in the header line needs some extra fine tuning.<br />
</p>

zyv:
The center dot (U+22C5) seems to be missing from quite some fonts, including Courier New till this day. The much more common middot (U+00B7) looks almost the same and is supported by virtually any font.

## Proposed changes

* Resolves: #1808

### DIFF
--- a/misc/skins/featured-plus.ini
+++ b/misc/skins/featured-plus.ini
@@ -143,7 +143,7 @@
 [widget-panel]
     sort-up-char = ↑
     sort-down-char = ↓
-    hiddenfiles-show-char = ⋅
+    hiddenfiles-show-char = ·
     hiddenfiles-hide-char = •
     history-prev-item-char = «
     history-next-item-char = »
--- a/misc/skins/featured.ini
+++ b/misc/skins/featured.ini
@@ -143,7 +143,7 @@
 [widget-panel]
     sort-up-char = ↑
     sort-down-char = ↓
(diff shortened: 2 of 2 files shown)


egmontkob (approved):
Cool, thanks! :)

mc-worker (approved):

PR #4957, opened 2026-01-12
Ticket #4956: Increase the skin selector's width

### LINKED ISSUE #4956: Skin selector: make it wider

```shell
.
```

### Operating system

```shell
.
```

### How to reproduce

Make it wider so that all the skin names we ship by default fit there without getting truncated.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
## Proposed changes

Make the skin selector window wider. This is just a trivial quick fix; ideally the width should be computed dynamically, but I leave that for another day.

* Resolves: #4956

## Checklist

### DIFF
--- a/src/filemanager/boxes.c
+++ b/src/filemanager/boxes.c
@@ -83,6 +83,8 @@
 #define B_KILL   (B_USER + 3)
 #endif
 
+#define SKIN_NAME_DISPLAY_LEN 25
+
 /*** file scope type declarations ****************************************************************/
 
 /*** forward declarations (file scope functions) *************************************************/
@@ -212,15 +214,15 @@ sel_skin_button (WButton *button, int action)
 
     (void) action;
 
-    skin_dlg = dlg_create (TRUE, 0, 0, 13, 24, WPOS_KEEP_DEFAULT, TRUE, dialog_colors,
-                           skin_dlg_callback, NULL, "[Appearance]", _ ("Skins"));
+    skin_dlg = dlg_create (TRUE, 0, 0, 13, SKIN_NAME_DISPLAY_LEN + 4, WPOS_KEEP_DEFAULT, TRUE,
+                           dialog_colors, skin_dlg_callback, NULL, "[Appearance]", _ ("Skins"));
     // use Appearance dialog for positioning
     skin_dlg->data.p = WIDGET (button)->owner;
 
     // set dialog location before all
     send_message (skin_dlg, NULL, MSG_RESIZE, 0, NULL);
 
-    skin_list = listbox_new (1, 1, 11, 22, FALSE, NULL);
+    skin_list = listbox_new (1, 1, 11, SKIN_NAME_DISPLAY_LEN + 2, FALSE, NULL);
     skin_name = "default";
     listbox_add_item (skin_list, LISTBOX_APPEND_AT_END, 0, skin_name_to_label (skin_name),
                       (void *) skin_name, FALSE);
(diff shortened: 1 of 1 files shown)


zyv (approved):

egmontkob:
This change is so trivial that I'm going ahead and submitting without 2nd approval.

Do we add NEWS entries for such insignificant changes, too?

ossilator:
> This change is so trivial that I'm going ahead and submitting without 2nd approval.

u-ha.
the change is trivial code-wise, but even such changes sometimes have entirely unexpected side effects. so it's better to just stick to the usual procedure. 

> Do we add NEWS entries for such insignificant changes, too?

dunno, but if the question is whether we should: imo, no.

zyv:
> > Do we add NEWS entries for such insignificant changes, too?
> 
> dunno, but if the question is whether we should: imo, no.

We used to add every closed ticket to the NEWS list. I think it still kind of makes sense even for smaller changes like this.

The solution that @mc-worker had for such items was to make a "cleanup" branch for each release and put such small changes there. I can live with that (as long as whomever maintains this branch rebases it). However, I'm not a huge fan of it, because I prefer the items to land in master as soon as possible to get tested, and also long-living branches are very annoying to rebase after sweeping changes.

Conversely, I think that we don't necessarily need an "issue" for every pull request. For bigger items, it might make sense. For small fixes... I don't really see how it helps. So for this one, for example, I have found I have forgone it.

Maybe @mc-worker can say if he agrees or not.

egmontkob:
>  it's better to just stick to the usual procedure.

@mc-worker , @zyv : What's your preferred workflow for tiny, trivial changes?

As for whether to add a NEWS entry, I'd probably vote for "no", but I'm happy to add if that's the project's standard procedure.

Followup question: How about fixes to bugs that were introduced in the same development cycle? (IMO: again "no")

> "cleanup" branch [...] However, I'm not a huge fan of it

I fully agree with you. If a change lands, I want all my other pending and usually irrelevant branches to also receive that fix, and hopefully never get hit by that bug / old behavior again.

> I think that we don't necessarily need an "issue" for every pull request

I did consider going without a separate issue; will do that in the future if there's not much to debug / discuss.

mc-worker:
Right now I have a local branch with a log of small or even tiny changes: refactoring, removing of code duplication, fixing coding style, etc. What should I do? Create an issue for each change? It's a madness.

I still prefer to use the cleanup branch. It's not so long live as you think, It depends on number of commits from one hand and and live time from another one. We had several merges (two or three) between releases.

egmontkob:
If this workflow works for you, you can of course do this. zyv and me probably won't do this as we don't like this workflow.

> What should I do? Create an issue for each change?

As discussed above, PRs without associated issues should be fine. You could just create a short-lived branch / PR and merge it immediately, without carrying it in parallel to master for a while, risking new merge conflicts over time.

(On a remotely related note, I'm really not a fan of having merge commits, I prefer projects with a linear history on master. Then trivial commits don't even need PRs, you can just directly commit to master.)

---

Back to the pending questions. @mc-worker What's your take on these?

-  With trivial changes, shall I wait for both your and zyv's approval, or can I go ahead after one approval? How about with more complex changes?

- Do we add NEWS entries for tiny insignificant changes? Do we add NEWS entries for fixes to bugs that were introduced during the very same development cycle?

mc-worker:
> PRs without associated issues should be fine

You don't like cleanup branch, I don't like PR.

> You could just create a short-lived branch / PR and merge it immediately, without carrying it in parallel to master for a while, risking new merge conflicts over time.

Short-lived branch is a branch with one-two-three commits. I'd prefer to make a branch with more commits (at least a dozen). But some time is required to collect such number of changes. The short-lived branch is actually not short-lived.

> With trivial changes, shall I wait for both your and zyv's approval, or can I go ahead after one approv

I have a lot of work and some problems with Internet, therefore I can be inactive a lot of time. Don't wait for me.

> Do we add NEWS entries for tiny insignificant changes?

I think we should have all **issues** in NEWS as visible for MC users. Small insignificant changes could be PRs without issues (like commits in the cleanup branch).

> Do we add NEWS entries for fixes to bugs that were introduced during the very same development cycle?

I add such bugfix to the parent issue. For example:

```
The command line history is deleted when file panel is switched to file list mode from any other one (#4919, #4944)
```

PR #4962, opened 2026-01-13
Ticket #4961: Revert the look of the "seasons" skins

### LINKED ISSUE #4961: Revert recent color changes to the "seasons-*16M" skins

```shell
current git
```

### Operating system

```shell
.
```

### How to reproduce

I would like to revert @horkykuba's commit b524cc367094580bfc80e493d6dd553f0f23b371 "misc/skins: Differentiate directories and files in true color skins".

My reasons:

1\. I don't think it addressed an actual issue.

Surely, most of our skins use a different color for directories vs. regular files. But I don't think that sharing the same color makes it hard to see whether an entry is a directory or a file. All those leading `/` characters underneath each other, suddenly switching to spaces, makes it extremely prominent. For symlinks, executables, and other rare file types (e.g. devices) that have some other symbol in that column, a prominently different color is used anyway.

2\. It doesn't address the claimed issue well.

The new dark gray vs. black (winter, spring, summer), or whiteish vs. yellowish (autumn) are quite close to each other. It's very hard to spot where the listing changes from one color to the other.

3\. Overall dullness

winter, spring, summer: The change does not only affect the color of regular files. It affects the frames, the ink color of the viewer, etc. Especially when viewing a file with F3 and the entire screen is painted with this dark gray (rather than pitch black), the look is duller, feels less crispy.

autumn: At directories, the less yellowish, more whiteish color give me a much lesser autumn vibe.

I'm using the winter skin by default for my daily work, and I've had that unconscious, very slight "it's not as crips, not as easily readable... are my eyes tired?... is the brightness of the screen too low?..." feeling for a while now; until – to check something utterly unrelated – I loaded up git vs. 4.8.33 in two different tabs, and noticed a slight change in colors as I switched tabs. That's how I realized that the skin actually changed and no longer used pitch black ink.

Finally, although it might not be a good reason to revert per se (as elsewhere I'm also criticizing and want to modify others' skins, and I'm not a good artist / graphical designer at all), but I'd like to also invoke the "I designed these skins, putting a lot of meticulous work into, and this is not the design I wanted to see" card.

(On an unrelated note, I'm surprised to see that over the course of 9 years no new truecolor skins appeared at all. These skins were meant to be examples only, although I believe they turned out pretty well – definitely magnitudes better than my sand256, LOL.)

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
Revert "misc/skins: Differentiate directories and files in true color skins"

This reverts commit b524cc367094580bfc80e493d6dd553f0f23b371.

## Proposed changes

* Resolves: #4961

## Checklist

### DIFF
--- a/misc/skins/seasons-autumn16M.ini
+++ b/misc/skins/seasons-autumn16M.ini
@@ -34,8 +34,6 @@
     MainFg = #ffa
     MarkedFg = #ff5
     HeaderFg = MarkedFg
-    DirectoryFg = #ffc
-    SelectedFg = DirectoryFg
     Selected = #681400
     Dialog = #85a918
     DialogFocus = #69880c
@@ -73,7 +71,7 @@
 
 [core]
     _default_ = MainFg;Main
-    selected = SelectedFg;Selected
+    selected = ;Selected
     marked = MarkedFg;;bold
     markselect = MarkedFg;Selected;bold
     gauge = ;DialogFocus
@@ -106,7 +104,7 @@
     errdframe = #fff;Error
 
 [filehighlight]
-    directory = DirectoryFg
+    directory =
     executable = #84ea84
     symlink = #ff58fd
     hardlink =
--- a/misc/skins/seasons-spring16M.ini
+++ b/misc/skins/seasons-spring16M.ini
@@ -31,11 +31,9 @@
 
 [aliases]
     Main = #f3ffe7
-    MainFg = #303030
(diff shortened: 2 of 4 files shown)


zyv (approved):
I don't have a stake in these skins because I don't use them myself, and I'm fine with you pulling the "original author" card.

mc-worker (approved):

PR #4964, opened 2026-01-15
Ticket #4963: Use OSC 8 hyperlinks in manual pages

### LINKED ISSUE #4963: Clicking on long, wrapped links in mc manuals takes to the wrong place

```shell
4.8.33, git
```

### Operating system

```shell
.
```

### How to reproduce

Many terminals automatically detect URLs on their display, and allow a convenient way (often Ctrl+click) to open them.

mc's man pages end with a few long(ish) URLs.

If the terminal isn't wide enough and they get wrapped, trying to open them opens the fragment which happened to be rendered in the first line.

It's also a bit cumbersome to copy-paste such wrapped URLs to the browser.

---

Luckily, there's an extension to solve this problem. Quite a few terminals support the OSC 8 escape sequence to create [web-like hyperlinks in the terminal](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda). This escape sequence is supported by [`groff` 1.23](https://lists.gnu.org/archive/html/groff/2021-10/msg00000.html) and [`less` 580](https://greenwoodsoftware.com/less/news.580.html) (although not without a bug, see below).

As per the groff announcement, the feature is disabled by default, you need to enable it in `/etc/groff/man.local`. At least Debian (and thus its derivatives too) patch it to enable by default, which is a pretty large user base. If I manually disable there, the feature still works for me, so there's more to this enabling-disabling story than this, which I haven't investigated.

When using this feature (making the URL text explicitly a hyperlink to the same URL), the first line of a wrapped URL becomes a hyperlink to the page. That is, Ctrl+clicking (or whatever way the terminal offers) on the first line opens the desired page, not a truncated URL.

Unfortunately, due to the interaction between `groff` (which emits a single run of URL spanning across multiple lines) vs. `less` (which expects all the SGR and this OSC attributes to start from the default state at the beginning of every line), subsequent lines of the URL can't be clicked. But it's up for them to fix one day, and even despite this bug, having a portion of the URL clickable and taking you to the right place is already a strong improvement.

---

Technically, just like on web pages, the displayed text could differ from the URL, as seen e.g. in the `dig` and `nsupdate` manuals pointing to some RFCs. I do not recommend that here though, because either groff or the user's terminal might be lacking this feature, and in that case we want the URL to be shown.

For the same reason, I don't want to eliminate the space in front of the sentence's full stop. When every component supports this feature, it would be safe to get rid of that. When not, and we revert to the terminal's URL auto-recognition, which isn't smart enough to exclude that in some terminals, it's useful to keep that space there.

---

What if such a manpage with explicit hyperlinks is viewed within `mcview`? Short answer: no improvement there, but luckily no regression either.

`mcview` invokes `groff`/`grotty` with the legacy compatibility `-c` option which emits bold and underlined text using the backspace + overstrike technique, rather than the modern output format using SGR and OSC. That's because mcview is only able to parse that legacy format, not the modern one.

As documented in `groff`'s announcement, in this mode such hyperlinks as meta-info are simply omitted. That is, mcview sees the very same formatted file as prior to this suggested improvement.

If one day we implement SGR support in `mcview` (which would probably need an underlying screen library that supports direct colors without the need for a pre-set palette, which neither ncurses nor slang do) then we'll surely be able to easily add OSC 8 hyperlink support as well.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
## Proposed changes

Use OSC 8 hyperlinks in manual pages

* Resolves: #4963

### DIFF
--- a/doc/man/es/mc.1.in
+++ b/doc/man/es/mc.1.in
@@ -4141,7 +4141,7 @@ carencia de garantía.
 .\"NODE "AVAILABILITY"
 .SH "DISPONIBILIDAD"
 La última versión de este programa se puede encontrar en
-https://ftp.osuosl.org/pub/midnightcommander/ .
+\X'tty: link https://ftp.osuosl.org/pub/midnightcommander/'https://ftp.osuosl.org/pub/midnightcommander/\X'tty: link' .
 .\"NODE "SEE ALSO"
 .SH "VÉASE TAMBIÉN"
 mcedit(1), sh(1), bash(1), tcsh(1), zsh(1), ed(1), view(1),
@@ -4149,7 +4149,7 @@ terminfo(1), gpm(1).
 .PP
 .nf
 La página web de Midnight Commander está en:
-	https://midnight\-commander.org
+	\X'tty: link https://midnight\-commander.org'https://midnight\-commander.org\X'tty: link'
 .fi
 .PP
 La presente documentación recoge información relativa a la versión 4.8
@@ -4169,8 +4169,8 @@ distribución.
 .SH "ERRORES"
 Véase el archivo "TODO" en la distribución para saber qué falta por hacer.
 .PP
-Para informar de problemas con el programa, introducir una nueva incidencia
-en https://github.com/MidnightCommander/mc/issues .
+Para informar de problemas con el programa, introducir una nueva incidencia en
+\X'tty: link https://github.com/MidnightCommander/mc/issues'https://github.com/MidnightCommander/mc/issues\X'tty: link' .
 .PP
 Se debe proporcionar una descripción detallada del problema, la
 versión del programa (que se obtiene con
--- a/doc/man/hu/mc.1.in
+++ b/doc/man/hu/mc.1.in
@@ -3156,7 +3156,7 @@ zsh(1).
 .nf
 A Midnight Commander World Wide Web oldalának címe a
 következő:
-	https://midnight\-commander.org
+	\X'tty: link https://midnight\-commander.org'https://midnight\-commander.org\X'tty: link'
 .fi
 .\"NODE "AUTHORS"
 .SH "Szerzők"
--- a/doc/man/it/mc.1.in
+++ b/doc/man/it/mc.1.in
@@ -3144,7 +3144,7 @@ per i dettagli sulla licenza e sulla mancanza di garanzie.
 .\"NODE "AVAILABILITY"
 .SH "REPERIBILITA'"
 L'ultima versione di questo programma si trova su
-https://ftp.osuosl.org/pub/midnightcommander/ .
+\X'tty: link https://ftp.osuosl.org/pub/midnightcommander/'https://ftp.osuosl.org/pub/midnightcommander/\X'tty: link' .
 .\"NODE "SEE ALSO"
 .SH "VEDERE ANCHE"
 ed(1), gpm(1), terminfo(1), view(1), sh(1), bash(1),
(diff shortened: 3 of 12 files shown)


egmontkob:
Currently I've only updated the English manual pages.

If you're conceptually okay with this change, I'll be happy to update all the translations as well.

zyv:
> If you're conceptually okay with this change, I'll be happy to update all the translations as well.

I have no objections on principle, but I'm wondering how it regresses in older Groff releases and other *roff renderers. For instance, what will happen to our rendered manual pages?

https://source.midnight-commander.org/man/mc.html

If it regresses gracefully, I see no reason to block this change.

egmontkob:
How is that manpage generated? If using the html backend of groff 1.23 then chances are you'll get clickable links there, too :)

---

Unfortunately I have no information on how older groff, or other roff implementations behave here. My assumption is that `\X'...'` is a generic placeholder for extensions, but I'm not sure.

I'm sort of hoping that groff folks were not foolish to introduce a backwards incompatible extension; and that `dig` and friends who already added this feature to their manpages also kinda verified compatibility, or would have already received bugreports and acted upon.

---

Tested now on:

Solaris: groff 1.23, the manpage renders as before (no OSC 8 hyperlink, no garbage)

macos: no groff, but mandoc (which seems to me is an alternate solution with the same overall functionality??? I"m really not sure); again, identical rendering as before, no OSC 8 hyperlink, no garbage

zyv:
> How is that manpage generated? If using the html backend of groff 1.23 then chances are you'll get clickable links there, too :)

https://github.com/MidnightCommander/mc/blob/676c9f497ec35b397c3ca77fcacca7fd4087c35d/.github/workflows/ci-fedora.yml#L71-L85

Unfortunately, every other generator that I've tried back then was truly garbage.

zyv:
> Tested now on:
> 
> Solaris: groff 1.23, the manpage renders as before (no OSC 8 hyperlink, no garbage)
> 
> macos: no groff, but mandoc (which seems to me is an alternate solution with the same overall functionality??? I"m really not sure); again, identical rendering as before, no OSC 8 hyperlink, no garbage

Well, in that case, I see no reason why we wouldn't want to merge this. Thank you for testing!

egmontkob:
> Unfortunately, every other generator that I've tried back then was truly garbage.

So, this is `man.mandoc` on Fedora.

On Ubuntu it wasn't installed for me; I installed `mandoc` (I expect it's the same) and got essentially the same output. No hyperlinks, no garbage either with the pending change.

Also tried `cat mc.1 | groff -Thtml > mc.html` now. No hyperlinks for these URLs (although I expected they'd appear), and, as you say, the entire output is "truly garbage"; hardly any formatting, pretty much every text fills the entire width of the browser.

---

> Well, in that case, I see no reason why we wouldn't want to merge this. Thank you for testing!

I was just thinking: If we have the slightest doubt or worry, I'd be more than happy to have this PR postponed until after the next release. It's really not urgent, and there it could receive more testing.

Let's hear Andrew...

mc-worker (approved):

egmontkob:
Updated the translations as well, plus `mcdiff.1.in` which I accidentally left out.

In one of the translations I removed italic typesetting (`.IR`) because the hyperlink doesn't work within that (with groff 1.23).

The first version was created manually, I re-did that part (to be sure I didn't make a mistake) and created the new bits by piping each URL through this script:

```
#!/bin/bash
IFS='' read -r link
echo -n "\\X'tty: link $link'$link\\X'tty: link'"

```

zyv (approved):

Issue #4975, opened 2026-01-28
Skin and color cleanups for 4.9 (meta-issue)


egmontkob:
### Is there an existing request for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

It's been suggested that the next release could be called 4.9. Have we agreed on this?

If so, I'd like to do some more skin/color-related changes.

---

#3160 already brought a slightly backwards incompatible change to the colors in this dev cycle.

#4974 (which I'd like to implement anyway even if stay on 4.8 series) will bring another slight one, unless explicit action is taken to stay backwards compatible.

If we go with 4.9, I'd like to address #3159 which would rename many-many keywords.

---

We can just update our skins and let users update their own modifications to them. Post a script somewhere (e.g. in this ticket) that does the search-replace. Document the old values along with the new ones in `misc/skins/README.txt`.

Or we can easily add a backwards compatibility fallback mechanism to mc's code. `mc_skin_get()` could have a new argument `fallback_key` to be looked up within the same `group` if `key` is not found. That way old skins would keep running until we drop this compatibility (e.g. in 4.10 / 5.0). This mechanism could also help us fork colors later on without problems. (We could also add the fallback keys for some widget chars whose names were changed a few releases ago.)

---

A 4.9 version bump would also be a perfect opportunity to remove some legacy options: mostly #4908, along with the easier ones #4950 / #4951.

As for #4908, if the functionality is needed then we'd still need to rework the code to accept the exact same identifiers as used in skin files, without having to manually maintain the list.

---

_Update:_

#5010: Drop `16` / `256` / `16M` from skin file names??

---

Are you guys conceptionally okay with this direction?

### Describe the solution you'd like

.

Issue #3159, opened 2014-01-15
Skin files: fix badly chosen keywords


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3159 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `egmont` (@egmontkob) |
> | **Mentions** | `ales.janda@….cz` (@kybl) |

<p>
Inspired by #2920 I started adding comments to default.ini. But it would be better if the names were more self-explanatory so we needed fewer comments.<br />
</p>
<p>
I propose to change at least these names:<br />
</p>
<p>
gauge -&gt; progressbar (seems people have no clue what gauge could stand for)<br />
reverse -&gt; currentdir (keyword should describe the role, not the look)<br />
[core] disabled -&gt; [dialog] disabled (or "ddisabled" with double d for consistency?), it's only used in dialogs<br />
helpslink -&gt; helpselectedlink<br />
</p>

egmontkob:
To follow the example of `[widget-*].*-char`, it would probably improve readability to separate words by dashes. E.g. `inputunchaged` -> `input-unchaged`, `commandhistory` -> `command-history`, `editrightmargin` -> `right-margin` etc.

---

Avoid unnecessary abbreviations. E.g. `[core]` has `selected`, `marked` and `markselect`, the latter should be `marked-selected`.

`dselnormal` -> `selected-normal` etc.

---

Under `[core]`, the meaning of `selected` vs. `marked` is totally confusing, I never know which one is which. Even the Skin Editor, up to very recently, demonstrated two files named `selected-file-1` and `selected-file-2` which were not selected but marked, according to the skin's file terminology.

I find both words ambiguous. I'd rather use `focus(ed)` (consistently with dialog buttons and such) and `tag(ged)`, I think both of these words are more obvious, and the Ctrl+T shortcut and the %t placeholder in the cmdline also refer to the terminology of "tagging".

egmontkob:
`lefttop` etc. are not proper English, they should be `topleft` etc..

For consistency with the rest, probably spelled as `top-left`.

The leading `d` for perhaps-double-perhaps-not should also be some properly spelled out `prefix-`. At this moment I can't think of a generic word that emphasizes the role rather than the look (e.g. `major-*`, `main-*`?? -- look silly to me). So maybe stick with the typical look and call them `double-top-left` etc.?

[...]

PR #4976, opened 2026-01-28
Ticket #4974: Skinnable hint bar, prompt and cmdline; plus other cleanups

### LINKED ISSUE #4950: Revise "terminal_default_color"

There's an undocumented skin keyword: Under the section `[skin]` you can have a `terminal_default_color`.

`lib/skin/colors.c` goes like

```
    CORE_DEFAULT_COLOR = mc_skin_color_get ("skin", "terminal_default_color");
    CORE_NORMAL_COLOR = mc_skin_color_get ("core", "_default_");
```

which is about as error prone as it could be.

Across the source, `CORE_DEFAULT_COLOR` is used at a couple of places, at some of which presumably it was meant to be `CORE_NORMAL_COLOR` (i.e. `[core]` -> `_default_`) instead.

What is this all about? Can we remove it? Or at least clean it up, not to have a "default" vs. "normal" confusion.

.

_No response_

### Additional context

_No response_

### LINKED ISSUE #4951: Revise "base" color

Skin files can have a special color or attrubite value called `base`. This refers to the corresponding color (or attributes) of `[core]` -> `_default_`.

None of the skins use it. Its documentation and its use is IMO counterintuitive.

The only place it would be useful is if there's a `[foo]` -> `_default_` defined to some other color, yet for `[foo]` -> `bar` you'd want to fall back by two steps to the default `[core]` -> `_default_`.

I don't see too much need for it. If there is, you can use aliases, or just repeat the same color (which you have to do many times anyways in skin definitions, unless you go with aliases).

Do we need this feature? Or okay to remove?

.

_No response_

### Additional context

_No response_

### LINKED ISSUE #4974: Skinnable hint bar & cmdline

It's hardcoded that the hint bar, the prompt, and the command line are "transparent", i.e. use the terminal's default colors.

(Except for marked text within the command line: `[core].commandlinemark`.)

There's no reason to keep this limitation.

For the command line, the rationale is probably that we want to provide the impression that mc is transparent here, although technically that's not true, we do paint the prompt ourselves, and e.g. colors in the user's prompt are lost. The hint bar being "transparent" can't even be justified by us trying to give this impression of transparency, I don't have an idea what was the design thought here.

It might be handy to separate the prompt and the command line using a different color, I expect a few skin authors wanting to do so. A third color for the hint bar (maybe some skins want to repeat the panels' colors, or so).

```
[core]
    hintbar = ...
    shellprompt = ...   (or prompt?)
    commandline = ...
```

Skins can already refer to the terminal's default color using the "default" keyword, as used intensively in the `modar*debfg*` themes. That is, the current look can be preserved by adding these keywords to our existing skins, with `default;default` values.

Now, this change would break the look of 3rd party skins until they're updated (since if a keyword is missing then we fall back to `[core]._default_`); #4975's fallback key might be a remedy.

Nitpicking: shall we call it "prompt" or "shellprompt"? The latter is more specific and is IMO more intuitive (mc's dialogs are also prompts in some sense), although technically not correct if subshell support is disabled and thus there's no shell.

---

This change would not only add a feature, but allow an internal cleanup.

The internal `transparent` flag would no longer be needed, it could be removed (replaced by a custom color). (Also remove it from the `hline` widget where it's unused now.)

The badly named `CORE_DEFAULT_COLOR`, referring to the terminal's default, would no longer be needed. ~Similarly, we'd no longer have to have color pair 0 always referring to the terminal's default color. The terminal's default colors, if used at all, would be allocated a random color pair number.~ The badly named `CORE_NORMAL_COLOR` could be renamed to its proper name `CORE_DEFAULT_COLOR`. See also #4950.

.

_No response_

### Additional context

_No response_

egmontkob:
## Proposed changes

- Make the hint bar, shell prompt and command line skinnable
- Use buttonbar hotkey color for unused areas of buttonbar (if the winow is narrower than 70 columns), so that we don't have any non-skinnable characters
- Remove "base" color value
- Remove "[skin].terminal_default_color" key
- Internal renames, cleanups and other minor fixes

* Resolves: #4974
* Resolves: #4950
* Resolves: #4951

## Checklist


egmontkob:
Please review along with pending questions in the referred issues, including:

- will the next release be called 4.9?
- if not, are we okay with landing this entire PR, or only the first 2 commits out of the 5?
- are we okay with the slight visual change at the bottom-right corner if the window is narrow?
- shall the keyword be "prompt" or "shellprompt"?

lib/widget/label.c
```diff
@@ -48,6 +48,11 @@
 
 /*** global variables ****************************************************************************/
 
+/* Color styles for label widgets */
+label_colors_t label_colors;
+label_colors_t hintbar_colors;
+label_colors_t prompt_colors;
```
mc-worker: Hintbar and prompt are parts of file manager. `hintbar_colors` and `prompt_colors` should be defined and declared in filemanager.[ch].

lib/widget/label.c
```diff
@@ -145,13 +141,35 @@ label_new (int y, int x, const char *text)
...
+{
+    label_colors[LABEL_COLOR_MAIN] = DIALOG_NORMAL_COLOR;
+    label_colors[LABEL_COLOR_DISABLED] = CORE_DISABLED_COLOR;
+
+    hintbar_colors[LABEL_COLOR_MAIN] = CORE_HINTBAR_COLOR;
+    hintbar_colors[LABEL_COLOR_DISABLED] = CORE_DISABLED_COLOR;  // unused
+
+    prompt_colors[LABEL_COLOR_MAIN] = CORE_PROMPT_COLOR;
+    prompt_colors[LABEL_COLOR_DISABLED] = CORE_DISABLED_COLOR;  // unused
+}
```
mc-worker: `hintbar_colors` and `prompt_colors` assignments should be moved to a separate function in the file manager area.

lib/widget/label.c
egmontkob: We've had pretty much the same discussion in PR #4916 :)

I continued the existing pattern to keep the code consistent.

I agree with you that it would be a cleaner design pattern to declare and initialize these colors at the caller, rather than the widget knowing about them. In order to keep our code consistent, I guess the existing code (e.g. alarm a.k.a. error, listbox, and help colors) should also be ported to this pattern.

Practically, it would mean the introduction of several tiny new methods, e.g. adding an `init()` to help as per the discussion of that PR. Also, when initializing mc or when applying a skin, we'd need to call many more such small methods.

I have doubts that it would be an overall win. But let me know if you still want me to do this, I will do then. Or maybe meet halfway: if you'd want to do that refactoring around alarm_colors / listbox_colors / help_colors in the exisiting code and then I'll rebase my changes and keep following your pattern?

egmontkob:
Labels in error dialogs have the wrong color.

I need to figure out how to denote in a label whether to inherit the colors from the parent widget (the wrongly removed `colors = widget_get_colors (w);` bits) or to store the colors in itself.

In a way that mostly fits the current design.

The "disabled" color is now hardwired to always use `[core].disabled`. The feature is only used in regular dialogs, not in error boxes -- and not in the newly added fields either. So it can probably stay hardcoded.

Which means that only a single color needs to be configurable. I might eliminate `label_colors_t` (and via that make the previous review comment irrelevant), just store either -1 for inheriting the color, or >=0 for an explicit value.

zyv:
> Please review along with pending questions in the referred issues, including:

I'm not sure what's the best place to have this discussion such that the posts are not mixed up with posts related to specific issues/PRs. Probably we could open a GitHub discussion for that. I would answer here, and if it becomes too annoying, then I would suggest we split it out.

> * will the next release be called 4.9?

My understanding is that Andrew didn't veto it, so in as far as I'm concerned, I'd say yes.

> * if not, are we okay with landing this entire PR, or only the first 2 commits out of the 5?

Irrelevant considering the above.

> * are we okay with the slight visual change at the bottom-right corner if the window is narrow?

In as far as I'm concerned, yes, I'm okay with that.

> * shall the keyword be "prompt" or "shellprompt"?

I think maybe this should be discussed in the ticket. I understand and agree that ambiguity is not good. However, I would tend towards "shell-prompt" or something like this if we decide to change it, because I find "separatorless" words annoying and very hard to read.

And, for the record, I'm still struggling to get through my mail.

egmontkob:
> However, I would tend towards "shell-prompt" or something like this if we decide to change it, because I find "separatorless" words annoying and very hard to read.

I will go for "shellprompt" in this PR, to be consistent (separatorless) with all the other color words.

I agree with your preference of using a separator, see #3159. I'll probably address that, along with a backwards compatible fallback option as per #4975, for all existing keywords. Then it'll turn into "shell-prompt". If I get to do this in the same dev cycle (which I'm truly hoping and aiming for) then there'll be no need for compat fallback for this one.

zyv:
> > However, I would tend towards "shell-prompt" or something like this if we decide to change it, because I find "separatorless" words annoying and very hard to read.
> 
> I will go for "shellprompt" in this PR, to be consistent (separatorless) with all the other color words.

👍

egmontkob:
Rebased. Renamed "prompt" to "shellprompt". Removed the last commit for now (there are more variables named similarly, I'll have to rename all or none).

Still haven't addressed Andrew's comments, I'm thinking about a bit of underlying refactoring around color handling, let's see how that turns out...

egmontkob:
Rebased and adjusted to go on top of the current state of PR #4995. This addresses @mc-worker's two review comments.

Let's review / finish #4995 first and then we can return to this one.

egmontkob:
> Labels in error dialogs have the wrong color.

Fixed now.

mc-worker (approved):

lib/widget/label.c
egmontkob: The refactoring in #4995 make it possible to address this comment cleanly.

lib/widget/label.c
egmontkob: The refactoring in #4995 make it possible to address this comment cleanly.

egmontkob:
@zyv please review, thanks!

lib/widget/hline.c
```diff
@@ -100,15 +100,11 @@ hline_callback (Widget *w, Widget *sender, widget_msg_t msg, int parm, void *dat
...
-        else
-        {
-            const int *colors;
+    {
+        const int *colors;
 
-            colors = widget_get_colors (w);
-            tty_setcolor (colors[DLG_COLOR_FRAME]);
-        }
+        colors = widget_get_colors (w);
```
zyv, line 106: I think even Andrew's guidelines allow joining declaration and assignment here.

zyv (approved):
Looks good to me (see minor comment).

lib/widget/hline.c
egmontkob: I'm leaving that code unchanged, just indenting by 1 that's why it appears here (but not in a `git diff -w`), so I'd prefer not to touch it.

egmontkob:
Just for the record: existing skin definitions were updated using this tiny makeshift script:

```
#!/bin/bash

while IFS='' read -r line; do
    if [[ $line =~ commandlinemark ]]; then
        echo "    hintbar = default;default"
        echo "    shellprompt = default;default"
        echo "    commandline = default;default"
    fi
    echo "$line"
done
```

PR #4978, opened 2026-02-01
Ticket #4978: remove `torben_fj_mode` panel navigation strategy

zyv:
## Proposed changes

As discussed in #3159:

I think it's time to remove this obscure navigation mode, which is disabled by default.


egmontkob (commented):
"Ticket #XXX" in the commit log. Other than that, looks good, thanks!

mc-worker (approved):

ossilator:
it would be prudent to include a few words arguing why this mode doesn't really add anything worthwhile.

i'd squash the commits.

zyv:
> "Ticket #XXX" in the commit log. Other than that, looks good, thanks!

This was intended :) I can't get a ticket number until I've pushed the commits and opened a PR. I didn't feel like opening an issue on top of that, and I thought that I could update the number during the rebase after I get any comments on the PR.

> it would be prudent to include a few words arguing why this mode doesn't really add anything worthwhile.

I added my justification with the caveat that "worthwhile" is really in the eyes of the beholder.

> i'd squash the commits.

I'd rather not do it.

PR #4980, opened 2026-02-01
Ticket #4908: Remove support for --colors and MC_COLOR_TABLE

### LINKED ISSUE #4908: Remove or update the `--colors` option?

```shell
4.8.33, git
```

### Operating system

```shell
.
```

### How to reproduce

Skin support appeared 16 years ago.

The `--colors` option is heavily lagging behind, both in the name of the recognized keywords, as well as documentation. I am pretty sure none of us feels like updating it.

Over time, conflicting keywords have also appeared in the skin file, under different sections. We have `[menu]/menusel` and `[popupmenu]/menusel` in the skins, they can't simply be flattened and both called `menusel` in the flat structure of the commandline option.

The ever-increasing number of colors makes it less and less useful to be able to override a few individual ones, or specifying an entire color scheme via this option. It's easier to copy a skin file to `~/.local/share/mc/skins` and edit it.

Can we just simply remove the `--colors` command line feature?

Or if not then somehow we'd need modernize to _automatically_ (without us ever having to maintain it manually) recognize all the skin file keywords, with their section names included, such as `--colors core/selected=blue,green:menu/menusel=red,black` etc.? Which would be a backwards incompatible change anyway. Is it worth it?

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
## Proposed changes

Remove support for the --colors command line option and the MC_COLOR_TABLE environment variable for specifying individual colors. The list of accepted keywords has long been unmaintained and outdated, making many of the colors impossible to override.

Use a skin file instead.

* Resolves: #4908

## Checklist


zyv (approved):
I like this, thank you very much for your work! 👍

I agree that this option has by now become more problematic than helpful. Either it has to be updated or removed. I'm also of the opinion that updating isn't worth it.

One thing that I was concerned about was custom skin loading. My understanding is that this option was often used to override colors for the root session and stuff like that.

However, it seems that `mc_skin_ini_file_load` can load a skin from the user directory, so this usage can be replaced by deriving a proper root skin.

There are even sample root skins on the basis of `modarin`...

egmontkob:
If there's public demand then we can revive this option; in a more generic way that allows anything in the skin file to be overridden.

Possibly using a slight variation on the syntax, like `--color 'core._default_ = red;green <some separator> core.selected = blue;yellow <separator> lines.horiz = -'`.

Possibly using the very same .ini syntax (having to pass literal newlines as part of cmdline args) and mc basically just merging the two .ini files: `--color $'[core]\n_default_ = red;green\nselected = blue;yellow\n[lines]\nhoriz = -'`

Maybe something like a `--skin-override` similarly to `--skin`, pointing to a partial skin `.ini` file whose contents would be merged. Or `--skin a+b` specifying two skins to be merged. The problem here becomes how to filter out partial skins from the graphical selector.

We'll figure it out if there's demand for this.

Some of my skins are heavily parameterized via the `[aliases]` section, basically you'd only need to override a few values there to have a consistent color change carried across the entire skin.

mc-worker (approved):

PR #4983, opened 2026-02-01
Ticket #4983: remove SLang-only softkeys resetting for extinct HP terminals

zyv:


ossilator:
seems reasonable.

i'd mention when the last relevant terminal was marketed.

zyv:
> seems reasonable.
> 
> i'd mention when the last relevant terminal was marketed.

Do you have a date? I have no idea, but the latest I remember seeing some was like 30 years ago.

ossilator:
no. but as you're making the claim that it's extinct, the onus is on you to make it appear plausible. it doesn't need to be super-precise ...

mc-worker (approved):

ossilator (approved):

PR #4986, opened 2026-02-02
Ticket #3023, #4985: Black and white skin cleanups

### LINKED ISSUE #3023: -b, --nocolor options not working usefully
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3023 |
> | **Reporter** | `dnied` (dnied@….it) |

<p>
The '-b' and '--nocolor' mc options don't produce usable output because the selection bar is not visible. Mc should probably use inverse-screen for the selection bar when asked to go colorless.<br />
</p>
<p>
Experimenting with $TERM, I found that mc actually draws an inverse-screen selection bar, though only if TERM=xtermm (but then the arrow keys don't work). Mc should draw it like that for any value of TERM, when '-b' or '--nocolor' are specified.<br />
</p>
<p>
Verified on 'xterm -cm', with and without GNUscreen.<br />
</p>

```
~> mc -V
GNU Midnight Commander 4.8.4
Built with GLib 2.32.4
Using the S-Lang library with terminfo database
With builtin Editor
With support for X11 events
Virtual File Systems: cpiofs, tarfs, sfs, extfs, ftpfs, fish, smbfs
Data types: char: 8; int: 32; long: 32; void *: 32; size_t: 32; off_t: 64;
~> xterm -version
X.Org 7.6.0(281)
~>
```

### LINKED ISSUE #4985: Remove "SPEC_A_*" colors

These constants

```
typedef enum
{
    SPEC_A_REVERSE = -100,
    SPEC_A_BOLD = -101,
    SPEC_A_BOLD_REVERSE = -102,
    SPEC_A_UNDERLINE = -103
} tty_special_color_t;
```

are only used if, as per this table

```
static mc_tty_color_table_t const color_table[] = {
    [...]
    // special colors
    { "A_REVERSE", SPEC_A_REVERSE },
    { "A_BOLD", SPEC_A_BOLD },
    { "A_BOLD_REVERSE", SPEC_A_BOLD_REVERSE },
    { "A_UNDERLINE", SPEC_A_UNDERLINE },
    // End of list
    { NULL, 0 },
};
```

one of the "A_BOLD" etc. keywords appear in color definition files. This is an undocumented feature and is not used in any of the skins. It is used internally for the hardcoded black-and-white skin, as per #3023.

The way it is used is: the keywords can only appear in "foreground color" position (or rather: as the entire value, with no following bg color or attrs), in which case it overrides the foreground, background, and attributes. It's unclear what happens if used in "background color" position (probably the same as any other unrecognized keyword).

The chosen colors use hardcoded lightgray/white (white/brightwhite) values for foreground and hardcoded black for background, i.e. were only designed to work on terminals with dark color scheme, and override the terminal's default choice for the exact shade of these colors.

The values (which shade of white) are inconsistent across ncurses vs. slang.

"A_BOLD" is not bold but bright in slang. "A_UNDERLINE" is not underlined in slang. "A_REVERSE" does not reverse the colors in ncurses. "A_BOLD_REVERSE" is unusable in both: same fg and bg.

Skin support supersedes this, as one can easily specify any these colors or the terminal's default color, combined with any combination of bold/reverse/etc.

Looks like an unused leftover from the pre-skin days which should be removed. In the hardcoded skin definition, their occurrances should be replaced by skin-compatible values such as `;;underline` or `white;black;underline` or such.

This ticket is very closely related to #3023, I wasn't sure whether just to comment there, but then I chose to open this new ticket for the removal of the engine, and keep that one for properly fixing the black-and-white skin.

.

_No response_

### Additional context

_No response_

egmontkob:
## Proposed changes

Fix & update the black and white skin. Followup internal cleanup.

* Resolves: #3023
* Resolves: #4985

## Checklist


egmontkob:
The combination of Slang with `TERM=xtermm` (and no `COLORTERM`) doesn't work as expected. The error dialog is reversed (although, according to the definitions, it shouldn't), and there's no highlight on the help pages.

As if there was another hardwired skin...?

Weird. Investigating...

egmontkob:
(Side note: `TERM=xterm-mono` might be a better choice for testing, since the arrow keys work there.)

No, there isn't another hardwired skin.

With mono TERM values, combined with Slang's color-aware methods, the "bold" flag becomes "reverse" instead. So if either of the properties (or both) are present, you get reverse video, and never bold.

(The same color-aware methods work properly if you pass "default" for both colors, i.e. restrict yourself not to use any colors, however TERM describes a color-aware terminal.)

I suspect this might be a Slang bug, or a pretty unfortunate design decision. Anyway, I've worked it around.

zyv:
Just tested on AIX (this is built with S-Lang, before all drawing characters were broken):

Screenshot 2026-02-03 at 09 23 32

It's looking better and better, but != signs still remain, no idea why.

zyv (approved):
Oh, that bugged me last time I worked on AIX. I agree with simplifying stuff. Thank you!

egmontkob:
> Just tested on AIX (this is built with S-Lang, before all drawing characters were broken):

Sorry, I'm really not following this comment.

The screenshot was taken without my patch, as per the commit ID shown. Does my patch change the colors in a notable way (for better or worse)?

By breaking all drawing characters do you mean that at some point real lines were replaced by `-` `+` `≠`, or did they become even more broken with my patch?

Could you please file a bugreport for the broken lines with the details?

The ACS variant of `|` is `≠`, while `-` and `+` stay intact. `printf $'\e)0\x0E - + | \e)0\x0F\n'` prints ` - + ≠ `. Surely this is what happens somewhere, the question is where, why, whose fault is that.

zyv:
> The screenshot was taken without my patch, as per the commit ID shown. Does my patch change the colors in a notable way (for better or worse)?

It is with your patch, and it makes it better. Your changes are in a different repository, and it's hard for me to check them out there. So I rolled the patch up manually. Black and white now makes sense. Before it was still black and white, somehow not consistent.

> By breaking all drawing characters do you mean that at some point real lines were replaced by `-` `+` `≠`, or did they become even more broken with my patch?

Before (not long after 4.8.33), all characters were completely randomly broken. Now, as you can see, the only weird stuff I can notice now is ≠. So the work you did before must have led to these improvements. I'm not sure why ≠ is not replaced.

> Could you please file a bugreport for the broken lines with the details?

I'm not sure what I should put in the report. If you log into cfarm119 and run mc, you should see what I see. The source is in `~src/mc`.

egmontkob:
> If you log into cfarm119 and run mc

Yup, I see this. The printed escape sequence is somewhat different, equivalent to `printf '\e(0 - + | \e(B'`, but essentially the same story.

> I'm not sure what I should put in the report.

Whatever we already know about the story, plus the screenshot :)

By the way, did you tamper with mc's config in some way? In my opinion, the first question is not why there are `≠` symbols instead of `|`. the first question is why doesn't mc fully shine in colors and proper line drawing chars; do you happen to know this?

`xterm`, `xterm-256color` and even `xterm-direct256` terminfos seem to be properly installed. `TERM=xterm infocmp -a` produces essentially the same output as on Linux, the only functional difference is `kbs=^H` vs. `kbs=^?` which is irrelevant.

Maybe slang wasn't built with color support or cannot locate the terminfo database or something like this?

At this point I'd bet it's a slang misconfiguration + a slang bug triggered by this misconfiguration.

egmontkob:
> With mono TERM values, combined with Slang's color-aware methods, the "bold" flag becomes "reverse" instead. [...]
> 
> 
> I suspect this might be a Slang bug, or a pretty unfortunate design decision. Anyway, I've worked it around.

I've found this in Slang's documentation, so it's not a bug but a feature, and what I happened to do as workaround is the proper solution.

Updated the comment in the source explaining the situation.

@mc-worker Could you please review?

mc-worker (approved):

PR #4990, opened 2026-02-03
Ticket #4834: suppress printing last `pwd` with `-P` only for `S-F10`

### LINKED ISSUE #4834: PWD Is not set on exit

```shell
GNU Midnight Commander 4.8.33-393-gcae5c32a8
Built with GLib 2.86.1 (using GLib 2.86.1)
Built with libssh2 1.11.1
With builtin editor
With support for X11 events
Virtual File Systems:
Data types:
```

### Operating system

```shell
Linux 6.17.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 23 Oct 2025 18:49:03 +0000 x86_64 GNU/Linux
```

### How to reproduce

When you quit the program with **exit** command, `current working directory` is not put in the `-P` flag file.

https://github.com/MidnightCommander/mc/blob/master/src/filemanager/command.c#L106
calls **quiet_quit_cmd()** which set `print_last_revert = TRUE;`

```cpp
else if (strcmp (cmd, "exit") == 0)
    {
        input_assign_text (lc_cmdline, "");
        if (!quiet_quit_cmd ())
            return MSG_NOT_HANDLED;
    }
```

```cpp
gboolean
quiet_quit_cmd (void)
{
    print_last_revert = TRUE;
    return quit_cmd_internal (1);
}
```

### Expected behavior

PWD is put to the -P flag

### Actual behavior

PWD is not put to the -P flag

### Additional context

_No response_

zyv:
## Proposed changes

According to my research, this behavior existed since the beginning of days, but users seem to find it confusing.

* Resolves: #4834

/cc @sandman7920 @alexhk please test

### DIFF
--- a/src/execute.c
+++ b/src/execute.c
@@ -527,7 +527,7 @@ toggle_subshell (void)
     if ((quit & SUBSHELL_EXIT) != 0)
     {
         // User did 'exit' or 'logout': quit MC
-        if (quiet_quit_cmd ())
+        if (quiet_quit_cmd (FALSE))
             return;
 
         quit = 0;
--- a/src/filemanager/command.c
+++ b/src/filemanager/command.c
@@ -106,7 +106,7 @@ enter (WInput *lc_cmdline)
     else if (strcmp (cmd, "exit") == 0)
     {
         input_assign_text (lc_cmdline, "");
-        if (!quiet_quit_cmd ())
+        if (!quiet_quit_cmd (FALSE))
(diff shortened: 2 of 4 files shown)


sandman7920:
Tested, everything works as it should.

alexhk:
Works, thanks for the quick fix!

mc-worker (approved):

PR #4995, opened 2026-02-05
Ticket #4995: color handling refactoring

egmontkob:
What would you guys think of a color handling refactoring along these lines?

The current code does the resolution from the _role_ (e.g. "dialog's title color") to the ncurses/slang color pair id as soon as possible.

A significant drawback of this design is that initializing some color constant (e.g. the ones used for normal, error, listbox and help dialogs), or later adjusting these values upon a runtime skin change are pretty cumbersome. Either some locally used colors need to be defined in a much more global context (pointed out as unfortunate design twice during code review: https://github.com/MidnightCommander/mc/pull/4916#issuecomment-3591709937 and https://github.com/MidnightCommander/mc/pull/4976#discussion_r2736867667), or the code that switches the skin needs to dig deep down into the territory of individual widgets to change the color they use.

My idea is to instead carry the color role identifier for as long as possible, and only resolve to the actual color when it comes to painting. This way only the color cache needs to be updated from the new skin; widget's idea on which color (i.e. color role) to use remains the same.

We have those 80-ish uppercase variables like `CORE_NORMAL_COLOR` etc. Currently it's pretty counterintuitive that despite the uppercase naming, these are variables and do change at a skin change. Yet, for some reason, they aren't declared as individual variables but as members of an array, so we have to maintain their indexing and manually renumber every time a new item is added.

With my suggestion, these become constants (actually enum values) as their name suggests, and we no longer have to manually index them.

Now, due to having two places in mc that bypass these variables, namely file highligthing and editor syntax highlighting, the story becomes a bit more complex. Sometimes we know the color role id, yet to be resolved to an ncurses/slang color pair id, but sometimes the latter one is the only thing we have.

Might not be the cleanest design per se, but I just distinguish these two based on how large the value is: if the value is low then it refers directly to the ncurses/slang color pair id; if the value is large (above the arbitrarily picked 4096) then it's a color role id which still needs a hop to resolve. This sort of pattern already exists in mc's file highlighting code where negative vs. positive values have different roles, although it looks to me it's legacy leftover because both are just color pair id, the negative ones will simply be flipped to positive (and, by the way, due to two faulty `ret > 0` checks rather than `ret >= 0`, color pair id 0 (i.e. the terminal's default colors) can't be used for file highlighting; something that probably nobody cares about). I've ported this negative vs. positive split to the new ranges.

No more conceptual `tty_setcolor()` vs. `tty_lowlevel_setcolor()` separation (although it was probably a legacy anyway, they were identical, except for a `& 0x7F` mask in slang that I don't understand), now by design there's a single method only which decides based on the value (small vs. large) how to handle it.

As you can see, the initialization of `command_colors` etc. becomes compile-time rather than runtime, no need to reinitialize at skin change, and easy to move all of them where they really belong.

I've moved `CORE_NORMAL_COLOR` etc. from the upper level skin layer to the lower level tty layer because `tty_setcolor()` now needs to know about them, and I didn't want tty to know about skin. This is debatable where it should really go, what the new variables / types / etc. should be named, and so on, suggestions welcome.

Let me know please how you like the overall direction.

### DIFF
--- a/lib/skin.h
+++ b/lib/skin.h
@@ -9,120 +9,123 @@
 
 /*** typedefs(not structures) and defined constants **********************************************/
 
-/* Beware! When using Slang with color, not all the indexes are free.
-   See color-slang.h (A_*) */
-
-/* cache often used colors */
-#define CORE_DEFAULT_COLOR         mc_skin_color__cache[0]
-#define CORE_NORMAL_COLOR          mc_skin_color__cache[1]
-#define CORE_MARKED_COLOR          mc_skin_color__cache[2]
-#define CORE_SELECTED_COLOR        mc_skin_color__cache[3]
-#define CORE_MARKED_SELECTED_COLOR mc_skin_color__cache[4]
-#define CORE_DISABLED_COLOR        mc_skin_color__cache[5]
-#define CORE_REVERSE_COLOR         mc_skin_color__cache[6]
-#define CORE_COMMAND_MARK_COLOR    mc_skin_color__cache[7]
-#define CORE_HEADER_COLOR          mc_skin_color__cache[8]
-#define CORE_SHADOW_COLOR          mc_skin_color__cache[9]
-#define CORE_FRAME_COLOR           mc_skin_color__cache[10]
-
-/* Dialog colors */
-#define DIALOG_NORMAL_COLOR          mc_skin_color__cache[11]
-#define DIALOG_FOCUS_COLOR           mc_skin_color__cache[12]
-#define DIALOG_HOT_NORMAL_COLOR      mc_skin_color__cache[13]
-#define DIALOG_HOT_FOCUS_COLOR       mc_skin_color__cache[14]
-#define DIALOG_SELECTED_NORMAL_COLOR mc_skin_color__cache[15]
-#define DIALOG_SELECTED_FOCUS_COLOR  mc_skin_color__cache[16]
-#define DIALOG_TITLE_COLOR           mc_skin_color__cache[17]
-#define DIALOG_FRAME_COLOR           mc_skin_color__cache[18]
-
-/* Error dialog colors */
-#define ERROR_NORMAL_COLOR     mc_skin_color__cache[19]
-#define ERROR_FOCUS_COLOR      mc_skin_color__cache[20]
-#define ERROR_HOT_NORMAL_COLOR mc_skin_color__cache[21]
-#define ERROR_HOT_FOCUS_COLOR  mc_skin_color__cache[22]
-#define ERROR_TITLE_COLOR      mc_skin_color__cache[23]
-#define ERROR_FRAME_COLOR      mc_skin_color__cache[24]
-
-/* Menu colors */
-#define MENU_ENTRY_COLOR    mc_skin_color__cache[25]
-#define MENU_SELECTED_COLOR mc_skin_color__cache[26]
-#define MENU_HOT_COLOR      mc_skin_color__cache[27]
-#define MENU_HOTSEL_COLOR   mc_skin_color__cache[28]
-#define MENU_INACTIVE_COLOR mc_skin_color__cache[29]
-#define MENU_FRAME_COLOR    mc_skin_color__cache[30]
-
-/* Popup menu colors */
-#define PMENU_ENTRY_COLOR      mc_skin_color__cache[31]
-#define PMENU_SELECTED_COLOR   mc_skin_color__cache[32]
-#define PMENU_HOT_COLOR        mc_skin_color__cache[33]  // unused: not implemented yet
-#define PMENU_HOTSEL_COLOR     mc_skin_color__cache[34]  // unused: not implemented yet
-#define PMENU_TITLE_COLOR      mc_skin_color__cache[35]
-#define PMENU_FRAME_COLOR      mc_skin_color__cache[36]
-
-#define BUTTONBAR_HOTKEY_COLOR mc_skin_color__cache[37]
-#define BUTTONBAR_BUTTON_COLOR mc_skin_color__cache[38]
-
-#define STATUSBAR_COLOR        mc_skin_color__cache[39]
-
-/*
- * This should be selectable independently. Default has to be black background
- * foreground does not matter at all.
- */
-#define CORE_GAUGE_COLOR             mc_skin_color__cache[40]
-#define CORE_INPUT_COLOR             mc_skin_color__cache[41]
-#define CORE_INPUT_UNCHANGED_COLOR   mc_skin_color__cache[42]
-#define CORE_INPUT_MARK_COLOR        mc_skin_color__cache[43]
-#define CORE_INPUT_HISTORY_COLOR     mc_skin_color__cache[44]
-#define CORE_COMMAND_HISTORY_COLOR   mc_skin_color__cache[45]
-
-#define HELP_NORMAL_COLOR            mc_skin_color__cache[46]
-#define HELP_ITALIC_COLOR            mc_skin_color__cache[47]
-#define HELP_BOLD_COLOR              mc_skin_color__cache[48]
-#define HELP_LINK_COLOR              mc_skin_color__cache[49]
-#define HELP_SLINK_COLOR             mc_skin_color__cache[50]
-#define HELP_TITLE_COLOR             mc_skin_color__cache[51]
-#define HELP_FRAME_COLOR             mc_skin_color__cache[52]
-
-#define VIEWER_NORMAL_COLOR          mc_skin_color__cache[53]
-#define VIEWER_BOLD_COLOR            mc_skin_color__cache[54]
-#define VIEWER_UNDERLINED_COLOR      mc_skin_color__cache[55]
-#define VIEWER_BOLD_UNDERLINED_COLOR mc_skin_color__cache[56]
-#define VIEWER_SELECTED_COLOR        mc_skin_color__cache[57]
-#define VIEWER_FRAME_COLOR           mc_skin_color__cache[58]
-
-/*
- * editor colors - only 4 for normal, search->found, select, and whitespace
- * respectively
- * Last is defined to view color.
- */
-#define EDITOR_NORMAL_COLOR       mc_skin_color__cache[59]
-#define EDITOR_NONPRINTABLE_COLOR mc_skin_color__cache[60]
-#define EDITOR_BOLD_COLOR         mc_skin_color__cache[61]
-#define EDITOR_MARKED_COLOR       mc_skin_color__cache[62]
-#define EDITOR_WHITESPACE_COLOR   mc_skin_color__cache[63]
-#define EDITOR_RIGHT_MARGIN_COLOR mc_skin_color__cache[64]
-#define EDITOR_BACKGROUND_COLOR   mc_skin_color__cache[65]
-#define EDITOR_FRAME_COLOR        mc_skin_color__cache[66]
-#define EDITOR_FRAME_ACTIVE_COLOR mc_skin_color__cache[67]
-#define EDITOR_FRAME_DRAG_COLOR   mc_skin_color__cache[68]
-/* color of left 8 char status per line */
-#define EDITOR_LINE_STATE_COLOR     mc_skin_color__cache[69]
-#define EDITOR_BOOKMARK_COLOR       mc_skin_color__cache[70]
-#define EDITOR_BOOKMARK_FOUND_COLOR mc_skin_color__cache[71]
-
-/* Diff colors */
-#define DIFFVIEWER_ADDED_COLOR       mc_skin_color__cache[72]
-#define DIFFVIEWER_CHANGEDLINE_COLOR mc_skin_color__cache[73]
-#define DIFFVIEWER_CHANGEDNEW_COLOR  mc_skin_color__cache[74]
-#define DIFFVIEWER_CHANGED_COLOR     mc_skin_color__cache[75]
-#define DIFFVIEWER_REMOVED_COLOR     mc_skin_color__cache[76]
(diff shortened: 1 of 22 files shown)


zyv (commented):
I've had a look at it, and mechanically your rework definitively seems like a much better way to achieve the same goal. My understanding of the whole system, however, is so limited that I can't judge if this approach introduces some new problems or not, and if there is an even better way to do it. I'm sorry, but I would like to defer to Andrew on this...

lib/skin/colors.c
```diff
@@ -241,97 +239,108 @@ mc_skin_color_set_default_for_terminal (mc_skin_t *mc_skin)
 
 /* --------------------------------------------------------------------------------------------- */
 
+/*
+ * Helper for mc_skin_color_cache_init ()
+ */
+static void
```
mc-worker: Let's make it inline.

lib/tty/tty-internal.c
```diff
@@ -108,3 +109,14 @@ tty_destroy_winch_pipe (void)
 }
 
 /* --------------------------------------------------------------------------------------------- */
+
+int
+tty_maybe_map_color (int color)
```
mc-worker: Let's move it to `color-internal.[ch]`.

lib/tty/tty-internal.c
```diff
@@ -108,3 +109,14 @@ tty_destroy_winch_pipe (void)
...
+{
+    if (color >= COLOR_MAP_OFFSET)
+        return tty_color_role_to_pair[color - COLOR_MAP_OFFSET];
+    else
```
mc-worker: `else` after `return` is unnecessary.

mc-worker (commented):
Cool! Thank you very much!

lib/skin/colors.c
egmontkob: Done.

lib/tty/tty-internal.c
egmontkob: Obviously, that's a much better place. I didn't realize we had such a file :) Done.

lib/tty/tty-internal.c
egmontkob: Done.

lib/tty/color.h
```diff
@@ -31,10 +31,146 @@ typedef struct
...
+ * even if they use the same colors in the skin. The problem with this is that for 8-color terminal
+ * descriptors (like TERM=linux and TERM=xterm) ncurses only allows 64 color pairs, so we can't
+ * afford to allocate new ids for duplicates.)
+ *
+ * In src/editor/editdraw.c these values are shifted by 16 bits to the left and stored in an int.
+ * Keep this number safely below 65536 to avoid overflow.
+ */
+#define COLOR_MAP_OFFSET 4096
+
+enum
```
mc-worker: I think the lib/tty layer should not know which basic widgets and "applications" (editor, viewer, ...) are implemented. Let's move color definitions and related macros back to lib/skin.

egmontkob:
(just a rebase; the review comment is yet to be addressed)

lib/tty/color.h
egmontkob: Sorry for the long delay. I was thinking of a truly clean solution, but I didn't find one.

I did the change you requested, but now we need to have two `lib/tty/*` -> `#include "lib/skin.h"` dependencies. Not sure if it's better. What do you think?

lib/skin/colors.c
```diff
@@ -241,99 +239,110 @@ mc_skin_color_set_default_for_terminal (mc_skin_t *mc_skin)
...
-
-    DIFFVIEWER_ADDED_COLOR = mc_skin_color_get ("diffviewer", "added");
-    DIFFVIEWER_CHANGEDLINE_COLOR = mc_skin_color_get ("diffviewer", "changedline");
-    DIFFVIEWER_CHANGEDNEW_COLOR = mc_skin_color_get ("diffviewer", "changednew");
-    DIFFVIEWER_CHANGED_COLOR = mc_skin_color_get ("diffviewer", "changed");
-    DIFFVIEWER_REMOVED_COLOR = mc_skin_color_get ("diffviewer", "removed");
-    DIFFVIEWER_ERROR_COLOR = mc_skin_color_get ("diffviewer", "error");
-
-    FILEHIGHLIGHT_DEFAULT_COLOR = mc_skin_color_get ("filehighlight", "_default_");
+    mc_skin_one_color_init (CORE_DEFAULT_COLOR, "skin", "terminal_default_color");
```
ossilator: given how uniform this code is, i would recommend a data-driven approach here.

lib/skin.h
```diff
@@ -9,120 +9,141 @@
...
-#define DIFFVIEWER_CHANGEDLINE_COLOR mc_skin_color__cache[73]
-#define DIFFVIEWER_CHANGEDNEW_COLOR  mc_skin_color__cache[74]
-#define DIFFVIEWER_CHANGED_COLOR     mc_skin_color__cache[75]
-#define DIFFVIEWER_REMOVED_COLOR     mc_skin_color__cache[76]
-#define DIFFVIEWER_ERROR_COLOR       mc_skin_color__cache[77]
-
-#define FILEHIGHLIGHT_DEFAULT_COLOR  mc_skin_color__cache[78]
-
-#define MC_SKIN_COLOR_CACHE_COUNT    79
+#define COLOR_MAP_OFFSET 4096
```
ossilator: you could just put that inside the/an enum as well.

lib/tty/color.h
ossilator: "there is no problem in cs that cannot be solved with another layer of indirection" ... :rofl: 

but more seriously, if you reframe a skin as the color palette, then it becomes obvious that it is intrinsically linked with color resolution. one could avoid that only by making the system way more dynamic, but that would also make it much heavier. so i'm not convinced that andrew's assertion holds up, but i haven't looked into it very deeply.

lib/skin/colors.c
egmontkob: If I understand correctly what you mean here then I agree, done.

lib/skin.h
egmontkob: Could you please be more specific? I can easily do this by "wasting" the first number in the enum, and correspondingly in the array of ints, but I'd prefer not to. Or the code would be fragile because elsewhere we'd need to know the symbolic name of the first entry and it'd break if someone changes the order.

egmontkob:
> but more seriously, if you reframe a skin as the color palette

Can't figure out how to reply to a comment on a file which is now outdated. (Yes, I know, we should switch to gerrit :))

Your comment makes sense, it might make sense to reposition this part of the source code as "palette" rather than "skin", but at the same time it is also a quite vague direction and I can't see either how the exact details should go, and for that reason preferably I wouldn't spend time on this right now.

lib/skin.h
ossilator:
```
enum
{
   COLOR_MAP_OFFSET = 4096,

   /* Basic colors */
   CORE_DEFAULT_COLOR = COLOR_MAP_OFFSET,
...

```

ossilator:
> Can't figure out how to reply to a comment on a file which is now outdated.

following the link in the notification mail takes you directly to it, where it's possible to reply as usual. but yeah, gerrit is _so_ much better at this ...

zyv:
> > but more seriously, if you reframe a skin as the color palette
> 
> Can't figure out how to reply to a comment on a file which is now outdated. (Yes, I know, we should switch to gerrit :))

Click on the "Files changed" tab and then on the grey "Comments" button on the right next to the green "Submit review".

lib/tty/color.h
mc-worker: > I did the change you requested, but now we need to have two `lib/tty/*` -> `#include "lib/skin.h"` dependencies. Not sure if it's better. What do you think?

I think I've fixed this: edf652eabd10db1dedec6d82ca8be9bc29c3a58e. What do you think?

lib/skin.h
egmontkob: I didn't know enums could have repeated values :) but Andrew's change is incompatible with your idea, so I guess we'll just leave his version.

lib/tty/color.h
egmontkob: > I think I've fixed this: [edf652e](https://github.com/MidnightCommander/mc/commit/edf652eabd10db1dedec6d82ca8be9bc29c3a58e). What do you think?

I generally like your approach, thanks!

Two nitpicks:

- You've added a `TTY_` prefix to `COLOR_MAP_OFFSET`, don't you want to carry it consistently to `COLOR_MAP_NEXT` and `COLOR_MAP_SIZE`, too?

- With the new g_array business, it's essential that we've called `tty_init_colors()` but not `tty_colors_done()`. Otherwise a color lookup accesses random uncontrolled parts of the memory. What do you think of setting `tty_color_role_to_pair` to `NULL` at its definition/initialization as well as in `tty_colors_done()`? It'd make broken code more obviously broken, accessing a `NULL` pointer rather than some random address.

And one more question:

I'm not familiar with this "fixup" workflow... is it that you simply manually begin the commiit message with `fixup!`, and then hitting the big green Merge button in GitHub will automatically squash it? Or what's the exact workflow with them?

Thanks!

mc-worker: > I think I've fixed this

Probably, usage of `GArray` is overdesign. Simple `int *` is enough.

egmontkob: > Probably, usage of `GArray` is overdesign. Simple `int *` is enough.

Indeed, good ol' malloc/free is just fine.

zyv: > I'm not familiar with this "fixup" workflow... is it that you simply manually begin the commiit message with `fixup!`, and then hitting the big green Merge button in GitHub will automatically squash it? Or what's the exact workflow with them?

You can edit things and then `git add -p` and `git commit --fixup REF` to mark them as belonging to a particular commit. After that, if you do `git rebase -i --autosquash`, it will automatically reorder marked commits and fuse them together. This is what my rebase bot does among other things. It's useful for self-review when you check your own patch series and make many small adjustments. Or to show you what I want to change more exactly without having you do a range diff on an already rebased patches. Check the manual for more details.

mc-worker: > * You've added a `TTY_` prefix to `COLOR_MAP_OFFSET`, don't you want to carry it consistently to `COLOR_MAP_NEXT` and `COLOR_MAP_SIZE`, too?

No. `COLOR_MAP_OFFSET` is defined in lib/tty, others aren't.

> With the new g_array business...

Done.

> I'm not familiar with this "fixup" workflow... is it that you simply manually begin the commiit message with fixup!

Such message is generated by `git commit --fixup COMMIT_TO_FIX`.

> and then hitting the big green Merge button in GitHub will automatically squash it? 

Can't say. I do everything on git-client side. All "fixup!" commits are squashed with target ones using `git rebase -i --autosquash ...`.

zyv: > > and then hitting the big green Merge button in GitHub will automatically squash it?
> 
> Can't say. I do everything on git-client side. All "fixup!" commits are squashed with target ones using `git rebase -i --autosquash ...`.

I have added merge checks that should ideally prevent exactly that. You might be able to bypass it as an administrator, but generally the idea is to make sure fixups don't get merged. To get rid of them, one would use the rebase bot or do it manually like Andrew.

egmontkob:
Thx guys!

Just pushed the autosquashed version.

I think it's fine to merge; right?

Once merged, let's move on to #4976 which was waiting for this one.

lib/skin/colors.c
```diff
@@ -244,96 +342,11 @@ mc_skin_color_set_default_for_terminal (mc_skin_t *mc_skin)
...
-
-    DIFFVIEWER_ADDED_COLOR = mc_skin_color_get ("diffviewer", "added");
-    DIFFVIEWER_CHANGEDLINE_COLOR = mc_skin_color_get ("diffviewer", "changedline");
-    DIFFVIEWER_CHANGEDNEW_COLOR = mc_skin_color_get ("diffviewer", "changednew");
-    DIFFVIEWER_CHANGED_COLOR = mc_skin_color_get ("diffviewer", "changed");
-    DIFFVIEWER_REMOVED_COLOR = mc_skin_color_get ("diffviewer", "removed");
-    DIFFVIEWER_ERROR_COLOR = mc_skin_color_get ("diffviewer", "error");
-
-    FILEHIGHLIGHT_DEFAULT_COLOR = mc_skin_color_get ("filehighlight", "_default_");
+    for (unsigned int i = 0; i < G_N_ELEMENTS (color_keywords); i++)
```
zyv: Can `unsigned int` here be replaced with `size_t`? I'd stay away from old-style types if possible at all; they always only bring uncertainty, confusion, and portability problems.

zyv (approved):
Fine with me, with one small comment...

lib/skin/colors.c
egmontkob: >  Can unsigned int here be replaced with size_t?

Done.

ossilator: "int" is optimized for the machine word size, and unless there is a good reason to actually consider values >4G (mc is 32+ bit only) or require a fixed word size for protocol reasons, using anything else than (unsigned) int is kinda counter-productive.

mc-worker (approved):

lib/skin/colors.c
egmontkob: We're not perf-critical here. The compiler already warns if the signedness of the two sides differ, who knows when it'll start to warn if their sizes differ. I prefer code cleanness over shaving off a few nanoseconds.

ossilator: it won't warn, because that would be utterly pointless, given that it's comparing with a constant. but it can be induced to emit truncation warnings when assigning variables.

sure it doesn't matter here, but it doesn't seem wise to train yourself to apply unnecessary pessimizations.

egmontkob:
Oops: #5023

PR #4998, opened 2026-02-05
Ticket #4996: skins: Add [filehighlight]._default_ color

### LINKED ISSUE #4996: In skins allow styling default file differently from UI labels

Currently it's not (afaik) possible to style differently the default file and then the labels in the UI (like the disk size). Both of these seem to be using `[core]._default_`. I'd like to style these differently - see the picture. 

Image

For solution, you could either add `[filehighlight]._default_` which would be consistent with other skin blocks or you could provide a separate param for styling the UI labels (like the inactive path, current file size and disk size). 

Note that my hacky solution to this problem was to add "any" category in filehighlight.ini and then style it separately, see https://github.com/notnout/mashdark so technically providing a catch-all category for files would also be a solution.

_No response_

### Additional context

_No response_

egmontkob:
## Proposed changes

Add [filehighlight]._default_ color

* Resolves: #4996

## Checklist


zyv (approved):

mc-worker (approved):

ossilator (commented):
i'd have appreciated a "slightly" more verbose commit message on the 2nd commit ...

PR #5001, opened 2026-02-06
Ticket #3159: Update color names

egmontkob:
## Proposed changes

Change the color names as proposed in https://github.com/MidnightCommander/mc/issues/3159#issuecomment-3812235479.

## Checklist

---

While working on this change, I noticed that mc used the same color names as slang (it does take strings like "brown", "yellow" etc. in its API). I assume that's where mc's terminology originated from, it just simply decided to pass the strings unchanged to slang and document them.

Luckily, slang has also accepted since the ancient days (199x) keywords like "color0", "color1" etc. instead. So I'm switching to using them, rather than having two confusingly slightly different namings, like mapping mc's "brightyellow" to slang's "yellow" and such.

---

Most of the big change was prepared using this trivial script

```
#!/bin/bash

sed -e 's/\bwhite\b/brightwhite/g' \
    -e 's/\blightgray\b/white/g' \
    -e 's/\bgray\b/brightblack/g' \
    -e 's/\byellow\b/brightyellow/g' \
    -e 's/\bbrown\b/yellow/g' \
    -i "$@"
```

but then I did a thorough review and had to revert here and there; most notably in css.syntax so that we don't change the color-related css keywords.

---

Although I've prepared this change now and I'd appreciate a review, I think I'll hold it back until we come up with more backwards incompatible changes, like the revamp of the skin file's keywords. I don't want this one to be the first (and if for whatever reason the work stop then the only) backwards incompatible skin change.


zyv (approved):
I simply scrolled through the changes trying to see if anything stands out that shouldn't have been replaced, but nothing jumped out.

egmontkob:
I forgot the obvious last step: a recursive search for the old names. I forgot to update one file, addressed now.

Issue #3733, opened 2016-11-29
Add the 'ENVIRONMENT VARIABLES' section to the manual page


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3733 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `andrew_b` (@aborodin) |

<p>
Add the 'ENVIRONMENT VARIABLES' section to the manual page and describe there all MC's environment variables: MC_KEYMAP, MC_SKIN, MC_TMPDIR, etc.<br />
</p>

PR #5004, opened 2026-02-06
Disable colors if NO_COLOR env var is present and nonempty

egmontkob:
## Proposed changes

Disable colors when environment variable `NO_COLOR` is present and nonempty, according to the "standard" described in http://no-color.org/

## Checklist

---

This was already submitted a long time ago, at https://github.com/MidnightCommander/mc-old/pull/133. Then it was closed without mc devs expressing their opinions.

Since then, a couple of things have changed:
- The standard fixed the handling of the env var's empty value.
- They clarified that app-specific options should be able to override it.
- They sort of clarified a few things on their homepage (it's not super-duper obvious, but I have the strong impression that they're fine with attributes like reverse video, but ANSI colors (even grayscale) are not okay.
- mc's black and white mode was fixed recently, and it happens to match this previous requirement.
- mc's development has moved here to github.
- The number of supporting projects listed on the no-color homepage has grown from 3 to 257.

NO_COLOR is still a somewhat controversial initiative, several people have expressed that it should have been done differently, but it's definitely gaining traction. With about 4 lines of source code, why not join them?

@RauliL I've prepared the change in your name as the original author. I had to modify two things: the handling of the empty value, and moving the code elsewhere so that `--color` can override it. Neither were your fault: the standard changed since then.

It's signed-off-by me. Not sure how it should be done ideally, but I don't think I should add someone else's name _there_. Please let me know if you have any comment or preference; maybe the cleanest would be if you could take back full ownership and re-submit the PR fully under your name?

### DIFF
--- a/doc/man/mc.1.in
+++ b/doc/man/mc.1.in
@@ -20,7 +20,9 @@ Unix\-like operating systems.
 Disable usage of graphic characters for line drawing.
 .TP
 .I \-b, \-\-nocolor
-Force black and white display.
+Force black and white display. Setting the
+.B NO_COLOR
+environment variable to a nonempty value has the same effect.
 .TP
 .I \-c, \-\-color
 Force color mode, please check the section
--- a/doc/man/mcdiff.1.in
+++ b/doc/man/mcdiff.1.in
@@ -18,7 +18,9 @@ specified on the command line.
 .SH OPTIONS
 .TP
 .I "\-b"
-Force black and white display.
+Force black and white display. Setting the
+.B NO_COLOR
+environment variable to a nonempty value has the same effect.
 .TP
 .I "\-c"
 Force color mode on terminals where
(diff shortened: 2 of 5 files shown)


zyv (approved):
Given the evolution of the `NO_COLOR` project, I wouldn't object on principle.

egmontkob:
Friendly pings

@mc-worker Review please

@RauliL Any feedback? I'll submit on your behalf in about a week if I don't hear back from you.

RauliL:
@egmontkob I'm OK with this.

egmontkob:
I realized I did not add documentation, so added now.

We don't have a dedicated "env vars" section and I didn't want to introduce one, so I've added it to the description of the `-b` option, which is a bit unusual choice. Let me know if you guys would prefer a different solution.

zyv:
> Let me know if you guys would prefer a different solution.

I'm still fine with that.

mc-worker:
> We don't have a dedicated "env vars" section and I didn't want to introduce one

#3733

mc-worker (approved):

Issue #4553, opened 2024-06-22
Allow styling of borders separately from default files


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4553 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `nout` (mc.lie8a@….me) |

<p>
One big issue that's preventing creation of better looking skins is that borders are always styled the same color as files (and other default foreground text). <br />
</p>
<p>
One partial trick is to introduce a new filehighlight that captures all files and style that separately in the skin. <br />
This is what I did in <a class="ext-link" href="https://github.com/notnout/mashdark">https://github.com/notnout/mashdark</a> , but that's hacky and users have to download my filehighlight.ini definition. Note that this trick doesn't work in dialogs and menus. <br />
</p>
<p>
So the request here is to add a new param in the style definition that would allow styling of the borders (ideally including the "buttons" for hidden files). <br />
</p>

> [!NOTE]
> Original attachments:
>
> * [mashdark-screenshot.jpg](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/4553/mashdark-screenshot.jpg) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/4553/mashdark-screenshot.jpg)) by `nout` (mc.lie8a@….me) on <span title="2024-06-22T17:30:19+00:00">`Jun 22, 2024 at 17:30 UTC`</span>

notnout:
In general - it would be great if the various labels, texts and borders are style-able separately. That's the main missing piece to allow really cool, smooth and consistent skins/themes.

zyv:
Related discussion in #2737.

egmontkob:
I've totally missed this issue when I filed and addressed duplicate issue #4931, my bad.

Closing this one as "reverse duplicate".

Could you please update your skin, and test with latest git snapshot of mc, or alternatively on https://skins.midnight-commander.org/ ? Note that the fix hasn't been released in a stable tarball yet.

There were other generic skin changes too which you might want to or need to incorporate, see my comments containing the word "rebase" at PR #4775 to see what they are.

notnout:
[...]

Issue #3531, opened 2015-10-01
roff underline+bold not recognized


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3531 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `egmont` (@egmontkob) |

<p>
I've just found that the \f4 macro in chacl.1 turns on underlined and bold at the same time, and is displayed by "less" accordingly.<br />
</p>
<p>
The *roff pipeline translates an underlined+bold character into:<br />
</p>
<p>
underscore backspace letter backspace letter<br />
</p>
<p>
Both the viewer and the search engine should be updated to recognize this.<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [chacl.1](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/3531/chacl.1) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/3531/chacl.1)) by `egmont` (@egmontkob) on <span title="2015-10-01T20:00:36+00:00">`Oct 1, 2015 at 20:00 UTC`</span>
> * [mc-3531-nroff-bold-underlined.patch](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/3531/mc-3531-nroff-bold-underlined.patch) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/3531/mc-3531-nroff-bold-underlined.patch)) by `egmont` (@egmontkob) on <span title="2015-10-01T21:51:10+00:00">`Oct 1, 2015 at 21:51 UTC`</span>

egmontkob:
Just checked the installed manual pages on my system.

Out of them, only `chacl(1)` and `rpcgen(1)` use the `\f4` macro, i.e. bold+underlined text.

I thought this change would have a bigger impact. Nevermind.

PR #5006, opened 2026-02-07
Add MashDark skin

notnout:
## Proposed changes

Adding new dark MashDark skin. This skin is using multiple new features, like styling borders separately and styling default files.

* Refs: #4553

## Checklist

Screenshot_2026-02-06_15-58-48

### DIFF
--- a/misc/skins/Makefile.am
+++ b/misc/skins/Makefile.am
@@ -12,6 +12,7 @@ SKINS = \
 	gray-orange-blue256.ini \
 	julia256.ini \
 	julia256root.ini \
+	mashdark256.ini \
 	mc46.ini \
 	modarcon16-defbg.ini \
 	modarcon16-defbg-thin.ini \
--- a/misc/skins/mashdark256.ini
+++ b/misc/skins/mashdark256.ini
@@ -1,0 +1,199 @@
+#
+# MashDark Midnight Commander skin
+#
+# See https://github.com/notnout/mashdark
+#
+
+[skin]
+    description = MashDark Skin
+    256colors = true
+
+[aliases]
+    # --- Backgrounds (Surface colors) ---
+    bg-main = color233
+    bg-menu = color235
+    bg-dialog = color238
+    bg-error = color52
+
+    # --- Foregrounds (Text colors) ---
+    fg-main = color231
+    fg-inactive = color246
+
+    # --- Accents (The "Pop" colors) ---
+    fg-accent = color214
+    bg-selection = color60
+    bg-secondary = color94
+
+    # --- Frames & Borders ---
+    fg-frame-main = color238
+    fg-frame-dialog = color243
+
+[lines]
+    horiz = ─
+    vert = │
+    lefttop = ╭
+    righttop = ╮
+    leftbottom = ╰
+    rightbottom = ╯
+    topmiddle = ┬
+    bottommiddle = ┴
+    leftmiddle = ├
+    rightmiddle = ┤
+    cross = ┼
+    dhoriz = ─
+    dvert = │
+    dlefttop = ╭
+    drighttop = ╮
+    dleftbottom = ╰
+    drightbottom = ╯
+    dtopmiddle = ─
+    dbottommiddle = ─
+    dleftmiddle = ├
+    drightmiddle = ┤
+
+[core]
+    _default_ = color246;bg-main
+    selected = fg-main;bg-selection
+    marked = fg-main;bg-secondary;bold
+    markselect = fg-main;color172;bold
+    gauge = color9;bg-secondary
+    input = color187;color235;bold
+    inputmark = color228;color88;bold
+    inputunchanged = color144;color235;bold
+    hintbar = fg-inactive;bg-main
+    shellprompt = fg-main;bg-main
+    commandline = fg-main;bg-main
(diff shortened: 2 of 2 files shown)


zyv:
Obsoletes #5005.

zyv (approved):
Looks nice to me.

egmontkob (commented):
Generally I quite like it. First few nitpicks:

- Please follow the naming convention of having `256` in the filename. (I don't really like this convention anymore, maybe we can discuss in another ticket to get rid of it for all our skins.)
- Please add new keyword `viewboldunderline`, can be tested with `chacl.1` (#3531).
- There are a few trailing spaces in the file.

Up until last night I thought it was my idea to use rounded corners (1b438db2a1ea82c8658337f91e7db8a5b904f6fd), then realized your skins has had them for years, cool! :)

egmontkob:
#4974 / #4976 is expected to land soon (a bit of underlying technical refactoring is on the way, not affecting the feature as it'll end up).

If, by any chance, you want to be the first skin specifying any of those three new colors as other than transparent, I think you can go ahead now and add those colors (or wait for a couple of days until that lands).

egmontkob:
The shadow color is almost invisible (I thought it was actually invisible until I looked up the definition). When there's a popup dialog, let's say Copy file, it essentially chops off parts of some filenames, which can be quite misleading.

Would you please consider making text in the shadow a bit less dark?

Those dark shades of gray are so close to each other. Even after magnifying the screen 8x with gimp and beefing up my screen's brightness to the max, I still literally can't see any difference with my naked eyes, although the color picker indeed confirms a slightly brighter black.

Generally I find the frame, headers etc. quite dark. Still readable, but close to borderline. I think it's an "artistic choice" that you want to de-emphasize the chrome more than other skins do so I'm fine with it. Maybe reusing that same color for the shadow could work better? Or just one step darker? It doesn't really have to be easily readable, but I think it should be hinted that there _is_ something there.

(I know it also depends on the monitor, surrounding light conditions, one's eyes etc... and I'm also blaming standard RGB namespace, and in turn the 256-color cube's grayscale ramp too, for really not being linear to how humans perceive their brightness and contrast. Or is it just me?)

mc-worker:
> #4974 / #4976 is expected to land soon (a bit of underlying technical refactoring is on the way, not affecting the feature as it'll end up).

There are a lot of conflicts with #4995. I think, #4995 should be merged first.

egmontkob:
> There are a lot of conflicts with #4995. I think, #4995 should be merged first.

Sorry, I wasn't explicit on this: yes, this is my plan (this is the underlying technical refactoring I was referring to). First complete #4995, then rebase #4974/#4976 on top of that.

For notnout to possibly test this new feature this is irrelevant, he can test the look with #4976's current state, if he's interested in this feature at all.

notnout:
Thank you for the feedback
- Indeed the borders and other elements are intentionally dark - the specific choice here is to make the files&folders really pop up (with #fff) and have the orange/yellow accent on some main action elements, while other control elements are intentionally dark. 
- updated the shadow fg to be more visible 
- renamed to add 256 (my suggestion would be to drop this requirement, unless it's meaningfully differentiating the same skin, e.g. if there's both xxx256.ini and xxx16M.ini)
- I introduced some aliases for common colors
- I added viewboldunderline, although I'm not styling it differently at this point
- I would like to style the commandline and hint, but I'd prefer to wait for the final keyword? Or is that clear? 

What is the submission process? Do I need to do anything else here other than fixing based on feedback? (this is my first time submitting like this on github, so I'm likely doing something wrong - please let me know)

egmontkob:
Thanks for the adjustments!

> I would like to style the commandline and hint, but I'd prefer to wait for the final keyword? Or is that clear?

I think we've settled on "hintbar", "shellprompt" and "commandline", the ones used by the current state of that PR. (If they'll change, I'll take care of adjusting this skin.)

If you want to skin them (some or all of them), I think the easiest is if we wait for that in this PR.

> What is the submission process? Do I need to do anything else here other than fixing based on feedback?

The commits should be squashed, either you can do that, or we can (we do have permissions to push to this branch of yours). After that we'll click the "Merge" button and add a NEWS entry, you don't have to do anything else!

egmontkob:
> renamed to add 256 (my suggestion would be to drop this requirement, unless it's meaningfully differentiating the same skin, e.g. if there's both xxx256.ini and xxx16M.ini)

Filed #5010.

notnout:
Thank you. I added "hintbar", "shellprompt" and "commandline", just changing the background for now.

egmontkob:
> just changing the background for now

That's not a good idea, it's unreadable if the terminal's basic colors are black on white (as e.g. I prefer it).

notnout:
Ok, added! 
(and I finally set my defaults for git to actually include signoff, sorry for any commit noise)

egmontkob:
Pushed an update on top of skinnable hintbar + cmdline which has landed now. Squashed the two commits and removed the unnecessary changelog bits.

egmontkob (approved):

egmontkob:
Thanks for this skin!

It has now landed in master and also appears on [skins.midnight-commander.org](https://skins.midnight-commander.org/), with the underlying engine changes included.

PR #4788, opened 2025-10-15
syntax: basic conf-file coloring and support for Dockerfile

KuzinAndrey:
## Proposed changes

Add basic syntax coloring support for most config files (*.cfg/*.conf)

PR #5009, opened 2026-02-07
syntax: fix TOML / add Kconfig syntax highlighting

### LINKED ISSUE #4933: syntax: multiple issues in syntax highlighting of TOML files

```shell
GNU Midnight Commander 4.8.30
Built with GLib 2.80.0
Built with libssh2 1.11.0
With builtin Editor and Aspell support
With support for X11 events
Virtual File Systems:
 cpiofs, tarfs, sfs, extfs, ext2undelfs, ftpfs, sftpfs, fish
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64;
```

### Operating system

```shell
Linux sknaumov-ThinkPad-T14-Gen-3 6.8.0-90-generic #91-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 18 14:14:30 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

### How to reproduce

Just open attached TOML [file](https://github.com/user-attachments/files/24141314/pyproject.toml.txt) in mcedit.

### Expected behavior

I've prepared alternative [toml.syntax](https://github.com/user-attachments/files/24141358/toml.syntax.txt) file.

Exact colors are discussable, but I've tried to unify them with syntax files of similar formats.

Image
Image

### Actual behavior

1. Broken comments.
2. Escapes in basic strings are neither highlighted nor properly handled.
3. Literal strings are not highlighted.
4. Multi-line strings are not properly handled.
5. Inconsistent highlighting of inline tables.
6. Sections are hardly distinguished from comments.
7. Colors are not unified with other similar files (YAML, INI). 
8. Acid-like look when there are a lot of strings.

Image
Image

### Additional context

Is there a chance that this fix will be available in Ubuntu 26.04?

### LINKED ISSUE #5003: syntax: add highlighting for Kconfig files

Currently there is no syntax highlighting for Kconfig files in mcedit.

I've prepared [syntax](https://github.com/user-attachments/files/25130621/kconfig.syntax.txt) file based on [Kconfig](https://docs.kernel.org/kbuild/kconfig-language.html) and [Kconfiglib](https://github.com/ulfalizer/Kconfiglib?tab=readme-ov-file#kconfig-extensions) specifications. Not sure about escapes, though - just copied them from TOML syntax file.

Image

As to files to highlight, I've come up with the following rule:
```
file (.\*[Kk][Cc]onfig.\*|Config.in)$ KConfig
include kconfig.syntax
```

`Kconfig`, `Kconfig.suffix` - for Linux kernel, `Config.in` - for Buildroot, and it should also be applicable for `file.kconfig` files.

I've also tried to make context starting at `help` not to highlight keywords like `if` in help text, but failed, as Kconfig help context is based on indentation, and, AFAIK, mcedit syntax can't account for indentation levels.

> The end of the help text is determined by the indentation level, this means it ends at the first line which has a smaller indentation than the first line of the help text.

Because of this I had to give up on trying to highlight single-quote strings, as  words like `don't` in help text would incorrectly start string context. Anyway, AFAIK, nobody uses single quote strings in Kconfig files.

sknaumov:
## Proposed changes

Improve syntax highlighting:

- Resolves: #4933
- Resolves: #5003

Highlight Kconfig output files.

## Checklist

I've not tried to build / install mc - I tested my changes on existing installation by modifying `/usr/share/mc/syntax/*` files.

### DIFF
--- a/misc/syntax/Makefile.am
+++ b/misc/syntax/Makefile.am
@@ -50,6 +50,7 @@ SYNTAXFILES =			\
 	java.syntax		\
 	js.syntax		\
 	json.syntax		\
+	kconfig.syntax		\
 	kotlin.syntax		\
 	latex.syntax		\
 	lisp.syntax		\
--- a/misc/syntax/Makefile.am
+++ b/misc/syntax/Makefile.am
@@ -51,6 +51,7 @@ SYNTAXFILES =			\
 	java.syntax		\
 	js.syntax		\
 	json.syntax		\
+	kconfig.syntax		\
 	kotlin.syntax		\
 	latex.syntax		\
 	lisp.syntax		\
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -369,5 +369,8 @@ include dockerfile.syntax
 file (/Kconfig(\\.[A-z0-9-]\+)\*|/Config\\.(in|src)(\\.[A-z0-9-]\+)\*|.\*\\.kconf(ig)?)$ Kconfig
 include kconfig.syntax
 
+file ..\*(\\.config(\\.[A-z0-9-]\+)\*|defconfig)$ Kconfig\soutput
+include properties.syntax
+
 file .\* unknown
 include unknown.syntax
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -366,5 +366,8 @@ include caddyfile.syntax
 file Dockerfile.\*$ Dockerfile
 include dockerfile.syntax
 
+file (/Kconfig(\\.[A-z0-9-]\+)\*|/Config\\.(in|src)(\\.[A-z0-9-]\+)\*|.\*\\.kconf(ig)?)$ Kconfig
+include kconfig.syntax
+
 file .\* unknown
 include unknown.syntax
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -378,5 +378,8 @@ include tree-meta.syntax
 file (/Kconfig(\\.[A-z0-9-]\+)?|/Config\\.(in|src)(\\.[A-z0-9-]\+)?|.\*\\.kconf(ig)?)$ Kconfig
 include kconfig.syntax
 
+file ..\*(\\.config(\\.[A-z0-9-]\+)?|defconfig)$ Kconfig\soutput
+include properties.syntax
+
 file .\* unknown
 include unknown.syntax
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -375,5 +375,8 @@ include meta-ii.syntax
 file ..\*\\.(?i:tmo?)$ TREE-META\sMetasyntax
 include tree-meta.syntax
 
+file (/Kconfig(\\.[A-z0-9-]\+)?|/Config\\.(in|src)(\\.[A-z0-9-]\+)?|.\*\\.kconf(ig)?)$ Kconfig
+include kconfig.syntax
+
 file .\* unknown
 include unknown.syntax
--- a/misc/syntax/kconfig.syntax
+++ b/misc/syntax/kconfig.syntax
@@ -1,0 +1,83 @@
+# Kconfig syntax
+#
+# Specifications:
+# https://docs.kernel.org/kbuild/kconfig-language.html
+# https://github.com/ulfalizer/Kconfiglib?tab=readme-ov-file#kconfig-extensions
+
+define kword     yellow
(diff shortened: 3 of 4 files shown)


sknaumov:
Please note that I used `/` for Kconfig.* files not to treat all files in `Kconfig` directory as Kconfig files.

It looks like Dockerfile file matching rules are incorrect, as all files in Dockerfile* directory will be treated as docker files.

sknaumov:
BTW, I decided to highlight kconfig output files as Java properties, just because `.config` - default Kconfig output file, is already highlighted by mc this way, but I'm not sure that it is a right way to go.

`.config` is quite generic name to treat it exclusively as some Java-related file, and in fact it is an env file in Kconfig case - that could be sourced into any shell script. So probably the right way is to write syntax file for `.env` files and use it here too instead of relying on some Java-related syntax. Or adjust this syntax to suit both Java properties and env files.

zyv:
> Please note that I used `/` for Kconfig.* files not to treat all files in `Kconfig` directory as Kconfig files.
> 
> It looks like Dockerfile file matching rules are incorrect, as all files in Dockerfile* directory will be treated as docker files.

Actually, there are quite a lot of rules (starting from the very first ones) that assume that we match on the file**name** and not the full path to the file.

@mc-worker, do you know if there is a reason why we actually match on the full path to the file and does it make sense? Maybe we should change this to really only matching on the file name?

```diff
% git diff
diff --git a/src/editor/syntax.c b/src/editor/syntax.c
index 9ccef28ae..8da7173d8 100644
--- a/src/editor/syntax.c
+++ b/src/editor/syntax.c
@@ -1555,7 +1555,7 @@ edit_load_syntax (WEdit *edit, GPtrArray *pnames, const char *type)
 
     f = mc_config_get_full_path (EDIT_SYNTAX_FILE);
     if (edit != NULL)
-        r = edit_read_syntax_file (edit, pnames, f, vfs_path_as_str (edit->filename_vpath),
+        r = edit_read_syntax_file (edit, pnames, f, vfs_path_get_last_path_str (edit->filename_vpath),
                                    get_first_editor_line (edit),
                                    auto_syntax ? NULL : edit->syntax_type);
     else
```

zyv:
> > Please note that I used `/` for Kconfig.* files not to treat all files in `Kconfig` directory as Kconfig files.
> > It looks like Dockerfile file matching rules are incorrect, as all files in Dockerfile* directory will be treated as docker files.
> 
> Actually, there are quite a lot of rules (starting from the very first ones) that assume that we match on the file**name** and not the full path to the file.

@KuzinAndrey, since you submitted the Dockerfile rules in #4788, could you please have a look at that and possibly other instances? Actually, I would love to see someone replace `..\*` with `.\+` for clarity and leave `.\*` wherever it is really needed. I'm afraid many cases are due to submitters simply not understanding the regex system.

zyv (approved):
Okay, I'm sorry for the lack of reaction. I have rebased the PR and updated the patterns to make them more understandable to me. @sknaumov, could you please check this?

I think this should be neither held back by #5067, nor the Dockerfile comment and/or other rule mistakes, of which it seems that we have plenty.

Hopefully, @KuzinAndrey would be interested enough to maybe clean the mistakes up, add case-insensitiveness where it makes sense, and make the patterns more understandable. He's done some valuable contributions in this direction in the past.

In my opinion, we should merge this.

KuzinAndrey:
@zyv i don't undestand issue with Dockefile rule.
In my case i have different filenames, for example:
- Dockerfile
- Dockerfile.alpine
- Dockerfile.debian
- Dockerfile-test1
- Dockerfile.bak1

And regexp with chars `./*$` at end satisfies all my cases.
If filename treated as full path to file, then path `Dockerfile/filename` of course be treated as Dockerfile rule by this regexp.
You suggest to change `vfs_path_as_str` -> `vfs_path_get_last_path_str` in `syntax.c:1558` and i think this really fix the problem.
I don't know where the path could be used, maybe only in the case of creating a syntax file for `.git/config`, and if you want i can make PR with new regexp ignoring separator char `/`:
```
file Dockerfile[^/]\*$ Dockerfile
include dockerfile.syntax
```
Or you can make these changes yourself.

sknaumov:
@zyv I used `\*` instead of `?` in patterns for 2 reasons:
1. There are Kconfig files with multiple extensions in Zephyr, like `Kconfig.board.v2`, `Kconfig.template.log_config.net` or `Kconfig.defconfig.series`. They are no longer highlighted with your changes.
2. It is always better to still highlight a file if possible if it is renamed with `.bak` suffix.

Issue #2142, opened 2010-04-15
mcedit highlighting stumbles upon multiple nested items (e.g. quotes)


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/2142 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `zaytsev` (@zyv) |
> | **Mentions** | `ksquirrel.iv@….com`, `gotar@….pl`, `Curtis@….net` (@dotysan), `muzzol`, `vitalif@….ru`, `polynomial-c@….de`, `teresaejunior@….com`, `xmlich02@….fit.vutbr.cz` |

<p>
Forwarded from Debian: <a class="ext-link" href="http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=577929">http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=577929</a><br />
</p>

```
From: Dmitry Baryshev <ksquirrel.iv@gmail.com>
Subject: mcedit colorer problem
Date: Thu, 15 Apr 2010 13:38:59 +0300

If I type the following string in a shell script:

valid=$(echo "$device" | awk '{ match($0, /^\/dev\/sd[a-z][0-9]$/); print
RSTART }')

mcedit will color all the following text as a character string (screenshot
is attached)

-- 
Regards, Krasu

```

> [!NOTE]
> Original attachments:
>
> * [снимок1.png](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/2142/%D1%81%D0%BD%D0%B8%D0%BC%D0%BE%D0%BA1.png) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/2142/%D1%81%D0%BD%D0%B8%D0%BC%D0%BE%D0%BA1.png)) by `zaytsev` (@zyv) on <span title="2010-04-15T11:53:31+00:00">`Apr 15, 2010 at 11:53 UTC`</span>
> * [highlighting.png](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/2142/highlighting.png) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/2142/highlighting.png)) by `asy` (a_s_y@….ru) on <span title="2014-04-02T05:35:44+00:00">`Apr 2, 2014 at 5:35 UTC`</span>

zyv:
[...]

PR #5019, opened 2026-02-19
Ticket #5018: Color pairs beyond 256

### LINKED ISSUE #5018: mcedit: some foreground / background color combinations look incorrect

```shell
GNU Midnight Commander 4.8.33-612-g3befffc92
Built with GLib 2.84.4 (using GLib 2.84.4)
Built with S-Lang pre2.3.4-16 and terminfo database (using S-Lang pre2.3.4-16)
With builtin editor
With mouse support on xterm
Virtual File Systems:
Data types:
```

### Operating system

```shell
Linux myhost 6.12.0-160000.9-default #1 SMP PREEMPT_DYNAMIC Fri Jan 16 09:29:05 UTC 2026 (9badd3c) x86_64 x86_64 x86_64 GNU/Linux
```

### How to reproduce

1. Configure `mcedit` to display color table: add the following two lines to `~/.local/share/mc/syntax/Syntax` file, before the final catch-all rule:
```
file ..\*\\.colors$ mcedit\scolors
include colors.syntax
```
2. Place the [colors.syntax](https://github.com/user-attachments/files/25406060/colors.syntax.txt) file attached to `~/.local/share/mc/syntax` directory.
3. Open the [catalog.colors](https://github.com/user-attachments/files/25406062/catalog.colors.txt) file attached in `mcedit` .

### Expected behavior

All foreground / background color combinations would look correct.

### Actual behavior

Some combinations seems to be incorrect e.g. 
- magenta / black
- lightgray / gray
- and many others...

And last two rows are black in whole.

(sorry, upload error when I'm trying to attach screenshots, I'll retry later)

### Additional context

Terminals:
either in KDE Konsole, "Black on White" color scheme:
```
COLORTERM=truecolor
TERM=xterm-256color
```
, or in the text virtual console:
```
TERM=linux
```
Possibly related issues:
- https://github.com/MidnightCommander/mc/issues/2169
- https://github.com/MidnightCommander/mc/issues/4968

egmontkob:
## Proposed changes

editor: Allow more contexts per syntax highlighting rule
tty: Allow more than 128 / 256 color pairs

* Resolves: #5018

## Checklist

### DIFF
--- a/lib/tty/color-ncurses.c
+++ b/lib/tty/color-ncurses.c
@@ -192,7 +192,7 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
 void
 tty_setcolor (int color)
 {
-    attrset (COLOR_PAIR (color) | color_get_attr (color));
+    attr_set (color_get_attr (color), color, NULL);
 }
 
 /* --------------------------------------------------------------------------------------------- */
--- a/lib/tty/color-slang.c
+++ b/lib/tty/color-slang.c
@@ -168,7 +168,7 @@ tty_setcolor (int color)
 void
 tty_lowlevel_setcolor (int color)
 {
-    SLsmg_set_color (color & 0x7F);
+    SLsmg_set_color (color);
 }
 
 /* --------------------------------------------------------------------------------------------- */
--- a/src/editor/editwidget.h
+++ b/src/editor/editwidget.h
@@ -35,8 +35,8 @@ struct edit_syntax_rule_t
 {
     unsigned short keyword;
     off_t end;
-    unsigned char context;
-    unsigned char _context;
+    unsigned short context;
+    unsigned short _context;
     unsigned char border;
 };
 


egmontkob:
I've tested with slang, ncurses 6.5 with widechar, ncurses 5.9 without widechar; combined with TERM=xterm, TERM=linux and more modern ones (256+ colors). Not all combinations of these, but a lot of them.

In 8-color terminals the situation still isn't perfect (isn't expected to be either), but there's no regression.

In 256+-color terminals the issues are now fixed, the test file of the original bugreport appeared as expected.

IMO let's land this first, I'll rebase #4995 on top of this.

mc-worker (approved):

zyv (approved):
Great work, thank you! 🚀 

I don't think there was a deep thinking behind limiting the number of contexts. My guess was that one had to pick something in the first place, and 256 was picked out of memory considerations along the lines of "640K should be enough for everybody".

I can't claim I understand (or took the time to understand) the system in-depth, but the idea was to allow for an extremely fast recovering streaming parser, which would be flexible enough to cover popular formats in a halfway sane way.

For this, you have "contexts" and within those "contexts" you have "keywords". The latter would roughly correspond, e.g. to standard library function names in a programming language. The former is more like areas within these function names that should be processed in different ways (like comment blocks). That's about it (and, by the way, the way the context idea is implemented is the genetic reason for #2142).

We definitively have syntax files with more "keywords" per "context" than 256, but I wasn’t sure that we have any with more than 256 contexts. A quick grep shows that we only have files with up to 60 contexts or so:

```
for i in *.syntax ; do grep -E "^context" $i | wc -l ; done
```

I don't think that relaxing this limitation would cause any serious issues though, other than giving users yet another opportunity to shoot themselves in the foot, of which they already have plenty anyways...

Issue #2931, opened 2012-11-18
Replace mcedit syntax highlighting engine with an external (?) library


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/2931 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `birdie` (aros@….com) |
> | **Mentions** | `gotar@….pl`, `mooffie@….com` (@mooffie), `teresaejunior@….com` |

<p>
MC's internal syntax highlighting engine has many serious limitations like it's unable to highlight embedded quotations, it breaks on new lines, etc. etc. etc.<br />
</p>
<p>
Colorer's can be licensed under GPL 2.0, LGPL 2.1 so it's fully compatible with MC's own license.<br />
</p>
<p>
Colorer can be used as an external library, it also has a C language API.<br />
</p>
<p>
<a class="ext-link" href="http://colorer.sourceforge.net/mc.html">http://colorer.sourceforge.net/mc.html</a><br />
</p>

PR #5022, opened 2026-02-20
Add QML syntax to editor

eugenesan:
## Proposed changes

* Adds support for QML  (Qt declarative meta lang)

## Checklist

- [N/A] I have referenced the issue(s) resolved by this PR (if any)
- [V] I have signed-off my contribution with `git commit --amend -s`
- [N/A] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [N/A] I have added tests that prove my fix is effective or that my feature works
- [N/A ] I have added the necessary documentation (if appropriate)

## Discussion
Currently QML was added as a standalone syntax file.
I'd like to hear opinions if instead it should be added to existing JavaScript syntax (`js.syntax`) as they share ~70% of syntax and QML can contains JavaScript code.

### DIFF
--- a/misc/syntax/Makefile.am
+++ b/misc/syntax/Makefile.am
@@ -82,6 +82,7 @@ SYNTAXFILES =			\
 	protobuf.syntax		\
 	puppet.syntax		\
 	python.syntax		\
+	qml.syntax		\
 	r.syntax		\
 	ruby.syntax		\
 	rust.syntax		\
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -366,5 +366,8 @@ include caddyfile.syntax
 file Dockerfile.\*$ Dockerfile
 include dockerfile.syntax
 
+file .\*\\.qml$ QML\sFile
+include qml.syntax
+
 file .\* unknown
 include unknown.syntax
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -375,5 +375,8 @@ include meta-ii.syntax
 file ..\*\\.(?i:tmo?)$ TREE-META\sMetasyntax
 include tree-meta.syntax
 
+file ..\*\\.qml$ QML\sFile
+include qml.syntax
+
 file .\* unknown
 include unknown.syntax
--- a/misc/syntax/qml.syntax
+++ b/misc/syntax/qml.syntax
@@ -1,0 +1,380 @@
+#
+# QML/JavaScript syntax highlighting
+# for MC Editor/CoolEdit
+#
+# Authors:
+#    * Vlad Romanenko <VladRomanenko@ukr.net>
+#    * Timur Shemsedinov <timur.shemsedinov@gmail.com>
+#    * Pavel Roskin <proski@gnu.org>
+#    * Milian Wolff <mail@milianw.de>
+#    * Eugene Sanivsky <eugenesan@gmail.com>
+#
+# Based on c.syntax; includes js.syntax and borrows keywords from Kate's qml.xml
+#
+# 2026-02-20 - Eugene Sanivsky <eugenesan@gmail.com>
+#    * Initial public release
+#
+
+context default
+#=========================
+    # Elements (http://doc.trolltech.com/4.7-snapshot/qmlelements.html)
+    keyword whole State yellow
+    keyword whole PropertyChanges yellow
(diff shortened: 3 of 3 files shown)


zyv:
> I'd like to hear opinions if instead it should be added to existing JavaScript syntax (`js.syntax`) as they share ~70% of syntax and QML can contains JavaScript code.

I don't think that QML should be added to the JavaScript syntax. As far as I understand, QML files can contain JavaScript, but the opposite is not true: QML is not valid JavaScript.

However, I think that you can include other syntax files in your syntax. So if the overlap is extremely high, you might have luck organizing the files in such a way that a common section is included in both, or one is included in another.

eugenesan:
> I don't think that QML should be added to the JavaScript syntax. As far as I understand, QML files can contain JavaScript, but the opposite is not true: QML is not valid JavaScript.

I agree.


> However, I think that you can include other syntax files in your syntax. So if the overlap is extremely high, you might have luck organizing the files in such a way that a common section is included in both, or one is included in another.

I tried "concatenation" using double include in Syntax and including from inside .syntax and this approach didn't work (`Load syntax file: Error in file .../qml.syntax on line 130`). Seems like currently the parser can handle default context only once and only at the beginning of the file.

Seems like unless someone reworks syntax loading in editor we'll have to deal with duplication across syntax files.

zyv:
> > However, I think that you can include other syntax files in your syntax. So if the overlap is extremely high, you might have luck organizing the files in such a way that a common section is included in both, or one is included in another.
> 
> I tried "concatenation" using double include in Syntax and including from inside .syntax and this approach didn't work (`Load syntax file: Error in file .../qml.syntax on line 130`). Seems like currently the parser can handle default context only once and only at the beginning of the file.
> 
> Seems like unless someone reworks syntax loading in editor we'll have to deal with duplication across syntax files.

Yes, but can't you split it into 3 files like `js.syntax`, `qml.syntax`, and `common-js-qml.syntax` or something like this even with the current parser?

ossilator:
preferably, js-common.syntax. there shouldn't be anything qml-specific in it.

more broadly, #2931 seems relevant once again. if only because dedicated highlighter libraries usually come with a sizable collection of syntax definitions.

eugenesan:
Unfortunately (unless I am missing something), it won't be possible to use common syntax file unless we want to cram "default context" of JS related languages into common and have them all use it.
Also, I've stopped investigating at "default context" level but other contexts might have the same limitations...

zyv:
> Unfortunately (unless I am missing something), it won't be possible to use common syntax file unless we want to cram "default context" of JS related languages into common and have them all use it. Also, I've stopped investigating at "default context" level but other contexts might have the same limitations...

Actually, I don't quite understand your problem. I think that our `include` statement is working just like text replacement. So, of course, if you include a file with a default context after a default context, it will not work. But my understanding is that something like this should work:

**qml.syntax**
```
context default
    include javascript-common-default.syntax

    keyword more...
```

**js.syntax**
```
context default
    include javascript-common-default.syntax
```

**javascript-common-default.syntax**
```
    keyword one
    keyword two
    keyword three
```

If you need to share more, you can make more files. Or does it not work?

It would be good to have this sorted out so that I can merge this PR.

eugenesan:
After some more testing, I was unable to use include lines in .syntax files.

It seems like the code in https://github.com/MidnightCommander/mc/blob/master/src/editor/syntax.c was designed to support `include` lines **only** in the main Syntax file.
It just errors when it sees `include` in .syntax file and I couldn't figure why in the time I had.

The code is undocumented and has parts that doesn't make sense to me.
For example the following part opens include file but doesn't seem to use the new file handle - https://github.com/MidnightCommander/mc/blob/9927d0710638f41ac27ffb1e8249d7d17814bdbf/src/editor/syntax.c#L1016

Unless someone figures out what is wrong with the code or shows what I am wrong, feel free to either merge the PR as is or close it.

misc/syntax/Makefile.am
```diff
@@ -109,7 +109,8 @@ SYNTAXFILES =			\
 	yabasic.syntax		\
 	yaml.syntax		\
 	yum-repo.syntax		\
-	yxx.syntax
+	yxx.syntax		\
+	qml.syntax
```
mc-worker: Please make this list sorted alphabetically again.

eugenesan: Sure. Done.

PR #5029, opened 2026-02-27
Learn keys: support Ctrl/Meta/Shift modifiers

### LINKED ISSUE #5027: Learn keys should support Ctrl/Meta/Shift modifiers and save modifier-aware terminal entries

Yes. Learn keys can currently teach base keys, but it does not provide an explicit way to teach/save modified
  variants as separate terminal mappings (for example shift-up, ctrl-f9, meta-right).
  As a result, users often have to manually edit ~/.config/mc/ini under [terminal:TERM], and it is hard to reliably
  persist modifier-specific escape sequences.

Add explicit modifier-aware key learning to the Learn keys dialog.

  Expected behavior:

  - Provide three modifier toggles: [ ] Ctrl, [ ] Meta (Alt), [ ] Shift.
  - When learning a key, selected modifiers are applied to that key definition.
  - Save learned sequences in [terminal:TERM] with modifier-prefixed key names, for example:
      - shift-up=...
      - ctrl-f9=...
      - meta-right=...
      - combined forms for multiple selected modifiers.
  - Keep modifier controls in a separate, non-overlapping row in the dialog so layout remains readable.
  - Preserve existing behavior when no modifiers are selected (base key names remain unchanged).

• 1. Manual editing of ~/.config/mc/ini ([terminal:TERM]) after using Learn keys.
     This works, but it is error-prone and not user-friendly for common modifier mappings.
  2. Using only keymap-level remapping (mc.keymap) without terminal mapping changes.
     This does not help when the terminal sequence itself is unknown or not learned correctly in the terminal
     section.
  3. Relying on terminal/terminfo fixes only.
     Helpful in some environments, but it does not solve the usability gap in MC’s own key-learning workflow for
     modifier variants.

### Additional context

In many terminals, modified arrows/function keys emit different escape sequences (for example CSI 1;2A, CSI 1;5C,
  etc.).
  cat -v may show distinct sequences, but Learn keys currently stores only base key entries, so users still need
  manual ini edits for modifier variants.

  A practical target is to allow learning and persisting entries such as:

  [terminal:xterm]
  up=\e[A
  down=\e[B
  left=\e[D
  right=\e[C

  shift-up=\e[1\;2A
  shift-down=\e[1\;2B
  shift-right=\e[1\;2C
  shift-left=\e[1\;2D
  ctrl-f9=...
  meta-right=...

  This would make terminal-specific modifier setup reproducible and avoid post-editing config files.

ilia-maslakov:
Fixes #5027

## Summary
- Add Ctrl/Meta(Alt)/Shift modifier support to Learn Keys dialog.
- Save learned terminal key entries with modifier-aware key names.
- Keep existing learning flow, extending it with explicit modifier selection.

## Details
- Added modifier checkboxes in Learn Keys UI.
- Applied selected modifiers when building learned key definitions.
- Saved key names with prefixes like `ctrl-`, `meta-`, `shift-`.
- Updated key validation logic to account for selected modifiers.

## Testing
- Built in `../build` with `--enable-vfs-sftp=no` due local libssh2 link issue unrelated to this change.
- Ran `make -C tests/src learn_modifiers check-TESTS TESTS=learn_modifiers -j$(nproc)` (PASS).

### DIFF
--- a/src/learn.c
+++ b/src/learn.c
@@ -63,6 +63,7 @@ typedef struct
     Widget *label;
     gboolean ok;
     char *sequence;
+    int modifiers;
 } learnkey_t;
 
 /*** forward declarations (file scope functions) *************************************************/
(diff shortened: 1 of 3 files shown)


ilia-maslakov:
dialog
learn_dlg

result
learn_res

ossilator:
i don't think that this will work ui-wise. iiuc, you basically implemented a tabbed interface, just in a very obfuscated way.

GfEW:
@ilia-maslakov 
Thank you - that's a very promising prospect!

@ossilator 
> i don't think that this will work ui-wise. iiuc, you basically implemented a tabbed interface, just in a very obfuscated way.

As an ordinary `mc` user, I don't understand the problem I'm "supposed to have" with that UI. The dialog is consistent with `mc`'s overall design principles, the separate modifier checkboxes can be toggled as needed and always indicate the current modifier combination with utmost clarity. What else should users be hoping for?

Moreover, I can't reasonably conceive a tabbed interface doing this job - it would need eight tabs to cover all 2^3 modifier combinations.



_UPDATE_: I do see other problems with this proposal version by now, some discussed by Egmont in #5027, some new:
- NumLock missing - in my view, this should be added as fourth mod toggle (some keys do send unique codes with NumLock on)
- numpad keys `KP0`..`KP9`,`KP_DEC` - missing
  - should be added
- `F14`/`S-F4` duplicates - `F13`..`F20`
  - should be removed
- `S-Tab`/`Shift`+`Tab` duplicate
  - `S-Tab` should be removed, use the mod toggle, only
- `M-Tab`/`Meta`+`Tab` duplicate
  - `M-Tab` should be removed, use the mod toggle, only.

I'll greatly appreciate if Ilia resumes, so modifier toggles can get into production.

PR #5041, opened 2026-03-03
Ticket #5040: Add a left padding to the help dialog

### LINKED ISSUE #5040: Help window has no left padding

```shell
4.8.33, git
```

### Operating system

```shell
.
```

### How to reproduce

The Help window has no inner padding on the left side. This results in a less pleasant look than with a bit of gap there.

It does have such a padding on the right side (not just in the screenshot but with any text), so the text flow isn't even centered.

Pretty much all the widgets do have an inner padding on the left and right sides (with the notable exception of the panels themselves).

Image

### Expected behavior

.

### Actual behavior

.

### Additional context

Unfortunately, at the de facto standard window width of 80 columns, it just so happens that adding that padding decreases the width available for text by 1 column, and it results on the trailing dot of "contents." to be wrapped to the next line.

It's another story that the rendering engine treats a change of rendering properties as a possible breaking point, which would be nontrivial to fix, and isn't my goal here.

I think at least the help's opening page shouldn't suffer from such a rendering issue. Without spending much time on fixing the rendering engine, there are two possible solutions:

- Eliminate "french spacing", i.e. the double spaces between sentences. This line happens to have one, it'd make up for the lost column. Many people claim that it's an outdated typesetting practice, and in the computer era it's preferable just to have a single space. It's used inconsistently, the other occurrence in this very screenshot having one space only.
 
- Increase the help window's width by 2 characters (1 on each side).

Any preference? :)

---

Also note that while I wish to push the left edge of the text 1 cell to the right, I don't want to push the big logo alongside. The logo alignes so nicely with the dialog's corner. This big logo is now defined with a leading space in each of its lines, removing that leading space would compensate for the new left padding.

egmontkob:
Also: Slightly increase the dialog's width to preserve the same wrapping on the opening page.

## Proposed changes

* Resolves: #5040

## Checklist

### DIFF
--- a/doc/hlp/es/xnc.hlp
+++ b/doc/hlp/es/xnc.hlp
@@ -1,11 +1,11 @@
 
 [main]
- lqwqk     k           k     
- x x x .   x     .     x     
- x x x k lqu wqk k lqw tqk n 
- x x x x x x x x x x x x x x 
- v   v v mqv v v v mqu v v mj
-     qqqqqqCommanderqj 
+lqwqk     k           k     
+x x x .   x     .     x     
+x x x k lqu wqk k lqw tqk n 
+x x x x x x x x x x x x x x 
+v   v v mqv v v v mqu v v mj
+    qqqqqqCommanderqj 
 
 Esta es la pantalla principal de ayuda de GNU Midnight Commander.
 
--- a/doc/hlp/hu/xnc.hlp
+++ b/doc/hlp/hu/xnc.hlp
@@ -1,10 +1,10 @@
 [main]
- lqwqk     k           k     
- x x x .   x     .     x     
- x x x k lqu wqk k lqw tqk n 
- x x x x x x x x x x x x x x 
- v   v v mqv v v v mqu v v mj
-     qqqqqqCommanderqj 
+lqwqk     k           k     
+x x x .   x     .     x     
+x x x k lqu wqk k lqw tqk n 
+x x x x x x x x x x x x x x 
+v   v v mqv v v v mqu v v mj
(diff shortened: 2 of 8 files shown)


egmontkob:
This corresponds to the second screenshot in the bug.

(I've had this PR lying around for a few weeks now... finally gotten to submit it :))

Let me know if you support removing the double spaces, I'd be happy to rework accordingly (and keep the original dialog width).

egmontkob:
dd0dd00 – what a nice commit id :)

mc-worker (approved):

zyv:
> This corresponds to the second screenshot in the bug.
> 
> (I've had this PR lying around for a few weeks now... finally gotten to submit it :))
> 
> Let me know if you support removing the double spaces, I'd be happy to rework accordingly (and keep the original dialog width).

I would very much favor removing the double spaces! Unfortunately, they are used in a lot of places, and quite inconsistently at that :(

egmontkob:
Went with removing double spaces; at least from the English bits of the help pages that aren't generated from the manuals, and also excluding some preformatted bits in the GPL text.

It's funny how, at the default terminal width of 80, the first 5-ish paragraphs of the GPL text were manually stretched using double spaces so that each line ends at the right margin. Except for the first line of the paragraphs which all align differently at the right edge. And it all completely falls apart if the window width is different from 80. Whoever did this gave up after about 5 paragraphs.

Reverted the widening of the help dialog, it's no longer needed.

egmontkob:
@zyv Friendly ping :)

zyv (approved):
Thank you!

PR #5042, opened 2026-03-04
Fix GString API migration regressions: lack of NUL after strncpy and use after free

peripherium:
strncpy() was called with text->len, which does not include the terminating NUL byte. As a result, when a shorter filename was copied after a longer one, the global buffer retained trailing characters from the previous value, producing an invalid path.

The fix copies text->len + 1 bytes to include the guaranteed NUL terminator provided by GString.

### DIFF
--- a/src/filemanager/ext.c
+++ b/src/filemanager/ext.c
@@ -80,12 +80,12 @@
 
 /*** forward declarations (file scope functions) *************************************************/
 
-// To prevent compiler warnings from being generated with --enable-tests
+#if defined(HAVE_TESTS)
 MC_TESTABLE
 GString *exec_make_shell_string (const char *lc_data, const vfs_path_t *filename_vpath);
 
-// To prevent compiler warnings from being generated without --enable-tests
-#if defined(HAVE_TESTS)
+MC_TESTABLE GString *exec_get_export_variables (const vfs_path_t *filename_vpath);
+
 extern char buffer[BUF_1K];
 #endif
 
@@ -185,7 +185,7 @@ exec_expand_format (char symbol, gboolean is_result_quoted)
 
 /* --------------------------------------------------------------------------------------------- */
 
-static GString *
+MC_TESTABLE GString *
 exec_get_export_variables (const vfs_path_t *filename_vpath)
 {
     GString *text;
--- a/src/filemanager/ext.c
+++ b/src/filemanager/ext.c
@@ -80,6 +80,15 @@
 
 /*** forward declarations (file scope functions) *************************************************/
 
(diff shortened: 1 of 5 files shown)


mc-worker (approved):

mc-worker:
Fixes 260126a89654dc83b2e2efa1c41e7539957a96aa.

zyv:
I wish we had a test for this function, especially when a bug is getting fixed.

@peripherium, would you be up for the task? I have added a test under similar circumstances; you can check this as an example: https://github.com/MidnightCommander/mc/commit/7488fc5acb1a1dac69e06d108f4c784292e0a676 .

KuzinAndrey:
I build mc from last commit https://github.com/MidnightCommander/mc/commit/9f0be66037ed17082406dc5fd77496ce908e7fcc and now can't open any tgz archive (after open one). Always get an error: 

But this fix is help to return to work.

peripherium:
Hi,

while adding the requested unit test for `exec_make_shell_string()`, I ran into a small question regarding the intended testing pattern.

The function was declared static, so I exposed it to the tests using `MC_TESTABLE`. However, it also relies on the global variable `buffer`, which is likewise declared static in the same C file.

To make the test build cleanly with and without `--enable-tests`, I added the following in the forward declarations section in `ext.c`:

```
// To prevent compiler warnings from being generated with --enable-tests
MC_TESTABLE
GString *exec_make_shell_string (const char *lc_data, const vfs_path_t *filename_vpath);

// To prevent compiler warnings from being generated without --enable-tests
#if defined(HAVE_TESTS)
extern char buffer[BUF_1K];
#endif
```

This works for the test, but I wanted to ask if this is the preferred approach in MC, or if there is an established pattern for handling static globals used by testable functions.

If there is a better way to structure this, I’d be happy to adjust the patch.

Thanks.

mc-worker:
@zyv let's merge this PR. Currently, in master, we have broken entering to archives.

mc-worker:
I found another regression in 260126a89654dc83b2e2efa1c41e7539957a96aa. The second commit in this PR fixes it.

@zyv let's merge this PR to fix the broken extfs in master.

zyv:
> @zyv let's merge this PR to fix the broken extfs in master.

Sorry, I'll try to have a look today.

zyv (approved):
Okay, guys, I'm sorry that it has taken me so long.

I have fixed the build on macOS due to duplicate symbols. Otherwise, I find the code good. And I appreciate the test very much. Thank you @peripherium for an excellent contribution!

I will not be adding this to the NEWS, since it's a regression in master after an internal migration.

PR #5050, opened 2026-03-06
man2hlp: remove escaping from dashes in section titles

zyv:
## Proposed changes

Some 15 years ago, I did a mass-fixing of unescaped dashes in the manual pages, including section titles (8a692f96). Unfortunately, our man2hlp converter wasn't able to remove escaped dashes form the section titles. Now it can.

Screenshot 2026-03-05 at 19 56 55

```diff
hlp % diff -Naur mc.hlp.orig mc.hlp
--- mc.hlp.orig	2026-03-06 12:50:43.620959892 +0100
+++ mc.hlp	2026-03-06 12:52:50.133107091 +0100
@@ -55,7 +55,7 @@
   Internal Diff ViewerDiff Viewer
   Internal File ViewerInternal File Viewer
   Internal File EditorInternal File Editor
-  Options of editor in ini\-fileInternal File Editor / options
+  Options of editor in ini-fileInternal File Editor / options
   Screen selectorScreen selector
   CompletionCompletion
   Virtual File SystemVirtual File System
```

### DIFF
--- a/src/man2hlp/man2hlp.in
+++ b/src/man2hlp/man2hlp.in
@@ -355,8 +355,8 @@ sub handle_node($$)
     }
     else
     {
-        # Remove quotes
-        $buffer =~ s/^"// and $buffer =~ s/"$//;
+        # Remove quotes and escaping from dashes
+        $buffer =~ s/^"// and $buffer =~ s/"$// and $buffer =~ s/\\-/-/;
         # Calculate heading level
         $heading_level = 0;
         $heading_level++ while substr($buffer, $heading_level, 1) eq ' ';


egmontkob (approved):
Just out of curiosity: what's the difference between `-` and `\-` in manpages?

zyv:
> Just out of curiosity: what's the difference between `-` and `\-` in manpages?

https://man.openbsd.org/mandoc_char.7#Dashes_and_Hyphens

In particular, Debian sticks to `\-` and regularly submits patches upstream for "offending" manual pages...

egmontkob:
So I guess it's only imporant to use `\-` where it really needs to stand for the ASCII `-` char, e.g. `\-\-version`, `ctrl\-alt\-right=...` etc.

The `\` could be omitted in plain English text, such as `Options of editor in ini-file`, `mc - Visual shell for Unix-like systems` etc.?

But it's probably easier just to be consistent and use `\-` throughout, rather than have to stop and think each time?

zyv:
> But it's probably easier just to be consistent and use `\-` throughout, rather than have to stop and think each time?

Yes, exactly, especially as distro linting tools do not differentiate between the two usages.

mc-worker (approved):

PR #5054, opened 2026-03-08
Tickets #4872, #4873 and #4874: fix 'insert control code' function

### LINKED ISSUE #4872: mcedit: Display corruption (missing header) after inserting literal `^J`

```shell
4.8.33, git
```

### Operating system

```shell
Linux
```

### How to reproduce

Start `mcedit`. Type a few random characters into the first line.

Press Ctrl-Q to invoke the "Insert literal" dialog. Press Ctrl-J.

The previously typed characters are gone.

Press the Left arrow. mc goes back to those previously typed characters, but now they occupy the header bar's room.

Interestingly the same doesn't happen if you simply press `J` without Ctrl. Then a newline is inserted and the display is updated as expected.

I guess the vertical positioning isn't adjusted??

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

### LINKED ISSUE #4873: mcedit: Insert literal, followed by special key acts weirdly

```shell
4.8.33, git
```

### Operating system

```shell
Linux
```

### How to reproduce

In `mcedit` press Ctrl-Q to open the "Insert literal" dialog, then press one of the special keys (like F1..F10, arrows, home-end-pgup-pgdn etc.).

In slang build, a raw byte of value roughly in the 128-160-ish range is inserted.

In ncurses build, often a raw control byte (0-31-ish) is inserted.

### Expected behavior

Do nothing, maybe close the dialog but don't insert anything.

### Actual behavior

Inserts something random, depending on the underlying library's representation of that key.

### Additional context

_No response_

### LINKED ISSUE #4874: mcedit: Insert literal should accept plain non-letters

```shell
4.8.33, git
```

### Operating system

```shell
Linux
```

### How to reproduce

`mcedit`, Ctrl-Q to open "Insert literal".

Pressing `a` ... `z`, or `^A` ... `^Z` equally inserts the corresponding control character. It doesn't matter if you hold down Ctrl or not.

The few non-letter symbols that conventially correspond to control characters, namely `@` `[` `\` `]` `^` `_`, have to be pressed with Ctrl. If pressed without, the plain character is inserted rather than the control code.

### Expected behavior

Should equally accept `@` `[` `\` `]` `^` `_` keypresses without Ctrl being held down.

### Actual behavior

They are inserted literally.

### Additional context

_No response_

peripherium:
The previous implementation used `ascii_alpha_to_cntrl()`, which assumed
alphabetic input and did not properly validate key codes returned by the
terminal libraries. The function has been renamed to `keycode_to_cntrl()`
and extended to correctly validate and convert key codes.

Changes:

- correctly handle `KEY_M_CTRL` combinations (e.g. `Ctrl+J`).
  - Fixes #4872.

- ignore special keys returned by `get_key_code()` (e.g. function or cursor keys).
  - Fixes #4873.

- extend control mappings for `@ [ \ ] ^ _`.
  - Fixes #4874.

- return `-1` for invalid input and ignore it at the call sites

The function is used both in the *Insert literal* dialog in `mcedit` and in the `mc` command line.

## Checklist
- [X] I have referenced the issue(s) resolved by this PR (if any)
- [X] I have signed-off my contribution with `git commit --amend -s`
- [X] Lint and unit tests pass locally with my changes (`make indent && make check`)

### DIFF
--- a/lib/util.c
+++ b/lib/util.c
@@ -51,6 +51,7 @@
 #include "lib/vfs/vfs.h"
 #include "lib/strutil.h"
 #include "lib/util.h"
+#include "lib/tty/key.h"  // KEY_M_CTRL
 
 /*** global variables ****************************************************************************/
 
@@ -64,11 +65,6 @@
 #endif
 
 #define TMP_SUFFIX ".tmp"
-
-#define ASCII_A    (0x40 + 1)
-#define ASCII_Z    (0x40 + 26)
-#define ASCII_a    (0x60 + 1)
-#define ASCII_z    (0x60 + 26)
 
 /*** file scope type declarations ****************************************************************/
 
@@ -1158,13 +1154,27 @@ early_error:
 
 /* --------------------------------------------------------------------------------------------- */
 
-extern int
-ascii_alpha_to_cntrl (int ch)
+int
+keycode_to_cntrl (int ch)
 {
-    if ((ch >= ASCII_A && ch <= ASCII_Z) || (ch >= ASCII_a && ch <= ASCII_z))
-        ch &= 0x1f;
+    // Extended keys such as F_KEY(n), Page Up and Page Down are not allowed.
+    if ((ch & ~(KEY_M_CTRL | 0x7f)) != 0)
+        return -1;
 
-    return ch;
+    // User pressed Ctrl+(ch) or '\t', '\'n
+    if ((ch & KEY_M_CTRL) != 0 || ch < 0x1b)
+        return ch & 0x1f;
+
+    if (ch >= '@' && ch <= '_')
+        return ch - '@';
+
+    if (ch >= 'a' && ch <= 'z')
+        return ch - '`';
+
+    if (ch == '?')
+        return 0x7f;
+
+    return -1;
 }
 
 /* --------------------------------------------------------------------------------------------- */
--- a/lib/util.h
+++ b/lib/util.h
@@ -262,9 +262,9 @@ void load_file_position (const vfs_path_t *filename_vpath, long *line, long *col
 void save_file_position (const vfs_path_t *filename_vpath, long line, long column, off_t offset,
                          GArray *bookmarks);
 
-/* if ch is in [A-Za-z], returns the corresponding control character,
- * else returns the argument. */
-extern int ascii_alpha_to_cntrl (int ch);
+/* Return the control character corresponding to key code ch,
+   or -1 if ch cannot be represented as a control character. */
+int keycode_to_cntrl (int ch);
 
 #undef Q_
 const char *Q_ (const char *s);
--- a/lib/widget/input.c
+++ b/lib/widget/input.c
@@ -1030,7 +1030,13 @@ input_callback (Widget *w, Widget *sender, widget_msg_t msg, int parm, void *dat
         if (parm == XCTRL ('q'))
         {
             quote = TRUE;
-            v = input_handle_char (in, ascii_alpha_to_cntrl (tty_getch ()));
+
+            const int control_code = keycode_to_cntrl (tty_getch ());
+
+            if (control_code != -1)
+                v = input_handle_char (in, control_code);
+            else
+                v = MSG_NOT_HANDLED;
             quote = FALSE;
             return v;
         }
--- a/tests/lib/Makefile.am
+++ b/tests/lib/Makefile.am
@@ -21,6 +21,7 @@ TESTS = \
 	serialize \
 	terminal \
 	tty \
+	util__keycode_to_cntrl \
 	utilunix__mc_pstream_get_string \
 	utilunix__my_system_fork_fail \
 	utilunix__my_system_fork_child_shell \
@@ -52,6 +53,9 @@ terminal_SOURCES = \
 tty_SOURCES = \
 	tty.c
 
+util__keycode_to_cntrl_SOURCES = \
+	util__keycode_to_cntrl.c
+
 utilunix__mc_pstream_get_string_SOURCES = \
 	utilunix__mc_pstream_get_string.c
 
@@ -66,4 +70,3 @@ utilunix__my_system_fork_child_SOURCES = \
 
 x_basename_SOURCES = \
 	x_basename.c
-
--- a/tests/lib/util__keycode_to_cntrl.c
+++ b/tests/lib/util__keycode_to_cntrl.c
@@ -1,0 +1,167 @@
+/*
+   lib - keycode_to_cntrl() function testing
+
+   Copyright (C) 2026
+   Free Software Foundation, Inc.
+
+   Written by:
+   Manuel Einfalt <einfalt1@proton.me>, 2026
+
+   This file is part of the Midnight Commander.
+
+   The Midnight Commander is free software: you can redistribute it
+   and/or modify it under the terms of the GNU General Public License as
+   published by the Free Software Foundation, either version 3 of the License,
+   or (at your option) any later version.
+
+   The Midnight Commander is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#define TEST_SUITE_NAME "/lib"
+
+#include "tests/mctest.h"
+#include "lib/util.h"
+#include "lib/tty/key.h"
+
+/* --------------------------------------------------------------------------------------------- */
+
+static const struct test_keycodes
+{
+    int keycode;
+    int expected_return;
+} test_keycodes[] = { { '@', 0 },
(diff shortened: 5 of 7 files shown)


zyv:
Since you are fixing `keycode_to_cntrl`, could you please add a test for it? Thank you!

zyv:
@egmontkob adding you, since this PR fixes 3 issues that you have reported...

peripherium:
I initially didn't add a test for this function because the previous implementation didn't have one, so I assumed that adding a test was not required.

In the meantime, I have taken a closer look at the test suite and added the requested test here.

If this test is not sufficient or needs adjustments, please let me know.

lib/widget/input.c
```diff
@@ -1029,8 +1029,13 @@ input_callback (Widget *w, Widget *sender, widget_msg_t msg, int parm, void *dat
     case MSG_KEY:
         if (parm == XCTRL ('q'))
         {
+            int control_code;
             quote = TRUE;
```
mc-worker, line 1032: Please add an empty line after variable declaration.

tests/lib/Makefile.am
```diff
@@ -25,7 +25,8 @@ TESTS = \
 	utilunix__my_system_fork_fail \
 	utilunix__my_system_fork_child_shell \
 	utilunix__my_system_fork_child \
-	x_basename
+	x_basename \
+	util__keycode_to_cntrl
```
mc-worker: Please move this line upper to make this list sorted alphabetically.

tests/lib/util__keycode_to_cntrl.c
```diff
@@ -0,0 +1,103 @@
+/*
+   lib - keycode_to_cntrl() function testing
+
+   Copyright (C) 2011-2026
+   Free Software Foundation, Inc.
```
mc-worker, line 5: Really 2011?

tests/lib/util__keycode_to_cntrl.c
```diff
@@ -0,0 +1,103 @@
...
+    /*
+     * Invalid inputs are non-ASCII or other special keys.
+     * These have bits set outside of 0x7f and KEY_M_CTRL.
+     */
+    {0x5000, -1}, {128, -1}, {0x0507, -1}
+};
+
+START_PARAMETRIZED_TEST (keycode_test, test_keycodes)
+{
+    int actual_result;
```
mc-worker: Please add an empty line after variable declaration.

mc-worker (commented):
Formatting issues: 
https://github.com/MidnightCommander/mc/actions/runs/22849421605/job/66290234418

lib/util.c
```diff
@@ -1159,12 +1155,23 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
 /* --------------------------------------------------------------------------------------------- */
 
 extern int
-ascii_alpha_to_cntrl (int ch)
+keycode_to_cntrl (int ch)
 {
-    if ((ch >= ASCII_A && ch <= ASCII_Z) || (ch >= ASCII_a && ch <= ASCII_z))
-        ch &= 0x1f;
+    // Only KEY_M_CTRL and lowest 7 bits of ch may be set
+    if ((ch & KEY_M_CTRL) != 0 && (ch & ~(KEY_M_CTRL | 0x7f)) != 0)
```
egmontkob: The condition doesn't match the comment above.

lib/util.c
```diff
@@ -1159,12 +1155,23 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
...
+        return ch & 0x1f;
+
+    if (ch >= '@' && ch <= '_')
+        return ch - '@';
+
+    if (ch >= '`' && ch <= '~')
+        return ch - '`';
 
-    return ch;
+    return -1;
```
egmontkob, line 1177: Shouldn't you also handle the common practice of Ctrl+`?` as 0x7F ?

lib/util.h
```diff
@@ -264,7 +264,7 @@ void save_file_position (const vfs_path_t *filename_vpath, long line, long colum
 
 /* if ch is in [A-Za-z], returns the corresponding control character,
```
egmontkob: You should update this comment with the additional characters you handle now; also the modified "else" return value.

lib/util.c
```diff
@@ -1159,12 +1155,23 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
...
 extern int
-ascii_alpha_to_cntrl (int ch)
+keycode_to_cntrl (int ch)
 {
-    if ((ch >= ASCII_A && ch <= ASCII_Z) || (ch >= ASCII_a && ch <= ASCII_z))
-        ch &= 0x1f;
+    // Only KEY_M_CTRL and lowest 7 bits of ch may be set
+    if ((ch & KEY_M_CTRL) != 0 && (ch & ~(KEY_M_CTRL | 0x7f)) != 0)
+        return -1;
+
```
egmontkob: Trailing spaces here, please run `make indent`.

lib/util.c
```diff
@@ -1159,12 +1155,23 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
...
+        return -1;
+    
+    // User pressed Ctrl+(ch)
+    if ((ch & KEY_M_CTRL) != 0)
+        return ch & 0x1f;
+
+    if (ch >= '@' && ch <= '_')
+        return ch - '@';
+
+    if (ch >= '`' && ch <= '~')
```
egmontkob: Do we need to extend beyond the lowercase ASCII chars?

Ctrl+backtick, Ctrl+`{` etc. are absolutely unconventional ways to denote the control chars, I don't think we should handle them.

egmontkob (commented):
The commit message is duplicated.

egmontkob (commented):
Tested now, and it indeed fixes those three bugs, thanks!

peripherium:
I included the lowercase characters because I thought it would be user-friendly: even if shift wasn't pressed during input, the corresponding control code can be entered via the dialog. However, this is a bit inconsistent with the last inserted condition, according to which `ch == '?'` returns 0x7f. Should I remove the conditions with the lowercase characters from the code?

egmontkob:
Lowercase letters are absolutely fine, in fact, expected; the user probably won't bother pressing Shift.

What's unnecessary IMHO is support the characters "surrounding" the lowercase letters in the ASCII table: ``` ` ``` `{` `|` `}` `~`. Here it's absolutely reasonable that the user will instead press `@` `[` `\` `]` `^` (or `_` which didn't have a counterpart).

peripherium:
Originally I allowed characters from `´` to `~` because I assumed someone looking at the ASCII table might wonder why those printable characters were excluded.

However, if this is not intuitive and could instead confuse future readers, then it doesn't belong in the code. So I removed it and restricted the range to `a-z`.

peripherium:
> The commit message is duplicated.

Sorry about the duplicated commit message - that must have slipped in during the amend commits while I was updating the patch. I've fixed the commit message now.

Please let me know if there is anything else in the patch that should be improved.

egmontkob (approved):
Just very quickly skimming through it this time, it looks good, thank you!

(Super minor nitpicking: you add an extra newline to the end of tests/lib/Makefile.am which already has enough (some will say that already has more than enough.))

Let's wait for Andrew's review as well.

lib/util.c
```diff
@@ -1159,12 +1155,26 @@ save_file_position (const vfs_path_t *filename_vpath, long line, long column, of
 /* --------------------------------------------------------------------------------------------- */
 
 extern int
```
mc-worker: I'd remove an `extern` word here.

lib/util.h
```diff
@@ -262,9 +262,9 @@ void load_file_position (const vfs_path_t *filename_vpath, long *line, long *col
 void save_file_position (const vfs_path_t *filename_vpath, long line, long column, off_t offset,
                          GArray *bookmarks);
 
-/* if ch is in [A-Za-z], returns the corresponding control character,
- * else returns the argument. */
-extern int ascii_alpha_to_cntrl (int ch);
+/* Return the control character corresponding to key code ch,
+   or -1 if ch cannot be represented as a control character. */
+extern int keycode_to_cntrl (int ch);
```
mc-worker: I'd remove an `extern` word here too.

mc-worker (commented):
Please begin the commit message with following line:

`Ticket #4872, #4873, #4874: fix 'insert control code' function`.

mc-worker:
1. Not all test are passed. `util__keycode_to_cntrl.log` is attached.
[util__keycode_to_cntrl.log](https://github.com/user-attachments/files/25976768/util__keycode_to_cntrl.log)

2. In editor, I use a paste of `\t` via "Insert literal" dialog (`ctrl-q`) very often. It doesn't work anymore.

peripherium:
I completely forgot to adjust the test after changing the `keycode_to_cntrl()` function. Thanks for pointing that out. I've already implemented all the other changes (especially in Makefile.am).

zyv (approved):
I'm happy with the code now, and I’m very happy about the test. Thank you very much! How about the others?

mc-worker (approved):

egmontkob:
Looked good to me the last time, won't check it again :P Thanks!

Issue #4741, opened 2025-08-07
mcview does not open for empty archives (xz, zstd, etc.)


kybl:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Midnight Commander version and build configuration

```shell
GNU Midnight Commander 4.8.33
Built with GLib 2.83.3
Built with S-Lang 2.3.3 with terminfo database
Built with libssh2 1.11.1
With builtin editor and aspell support
With subshell support as default
With support for background operations
With mouse support on xterm and Linux console
With support for X11 events
With internationalization support
With multiple codepages support
With ext2fs attributes support
Virtual File Systems:
 cpiofs, tarfs, sfs, extfs, ext2undelfs, ftpfs, sftpfs, shell
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64; uintmax_t: 64;
```

### Operating system

```shell
Linux p16v 6.14.0-27-generic #27-Ubuntu SMP PREEMPT_DYNAMIC Tue Jul 22 17:01:58 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
```

### Is this issue reproducible using the latest version of Midnight Commander?

- [ ] I confirm the issue is still reproducible with the latest version of Midnight Commander

### How to reproduce

```
$ echo hello | zstd > hello.zst
$ echo hello | pzstd > hello-parallel.zst
$ file *.zst
hello.zst:          Zstandard compressed data (v0.8+), Dictionary ID: None
hello-parallel.zst: Zstandard compressed data (v0.8+), Dictionary ID: None
```

Then open the folder in Midnight Commander. Press F3 on file:

- hello.zst: It is decompressed successfully, I see "hello"
- hello-parallel.zst: It shows the compressed bytes (it does not recognizes as zstd) <-- that is wrong

Both `zstd -d` and `zstdcat` (and of course `pzstd -d`) can decompress both files.
I wonder why mc does not recognize it.

### Expected behavior

When I press F3, it should show the decompressed content.

### Actual behavior

It does not recognize the file produced by pzstd as a zstd file. It shows the compressed binary form.

zyv:
I think the problem is the difference in `file` output. It should be the same, but it isn't.

```
mc % file -z -b -L -S hello.zst
ASCII text (Zstandard compressed data (v0.8+), Dictionary ID: None)

mc % file -z -b -L -S hello-parallel.zst
[...]

PR #5058, opened 2026-03-09
viewer: fix crash when viewing files with PK/ZIP magic bytes

### LINKED ISSUE #4760: Cannot open <OpenDocument file path> in parse mode. Operation not supported (95)

```shell
GNU Midnight Commander 4.8.33
Built with GLib 2.82.4
Built with libssh2 1.11.1
With builtin editor and aspell support
With support for X11 events
Virtual File Systems:
 cpiofs, tarfs, sfs, extfs, ext2undelfs, ftpfs, sftpfs, shell
Data types:
```

### Operating system

```shell
Linux admin 6.12.43+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.43-1 (2025-08-27) x86_64 GNU/Linux
```

### How to reproduce

Instead of a single office suites, I am using different softwares. ODS default is Gnumeric, ODT default is AbiWord, etc. Both are not opening when pressed Enter on the respective files. With F3, I am getting the (95) error.

### Expected behavior

It should simply open the file in a separate default application window.

### Actual behavior

Cannot open \<file path\> in parse mode
Operation not supported (95)

When pressed [OK], it just opens the file in the terminal itself, which just displays gibberish.

### Additional context

mc.ext.ini:

```
# StarOffice 6 and OpenOffice.org formats
[OpenOffice.org]
Regex=\\.(odt|fodt|ott|sxw|stw|ods|fods|ots|sxc|stc|odp|fodp|otp|sxi|sti|odg|fodg|otg|sxd|std|odb|odf|sxm|odm|sxg)$
RegexIgnoreCase=true
Open=/usr/lib/mc/ext.d/doc.sh open ooffice
View=%view{ascii} /usr/lib/mc/ext.d/doc.sh view odt

[AbiWord]
Shell=.abw
ShellIgnoreCase=true
Open=/usr/lib/mc/ext.d/doc.sh open abw

[Gnumeric]
Shell=.gnumeric
ShellIgnoreCase=true
Open=/usr/lib/mc/ext.d/doc.sh open gnumeric
```

ilia-maslakov:
## Summary

- Fix segfault when pressing F3 on files with PK/ZIP magic bytes (PPTX, DOCX, ODS, ODT, JAR, etc.)
- When VFS decompression fails, silently display raw file content instead of corrupting viewer state

* Resolves: #4760

## Root cause

In `mcview_load()`, when `get_compression_type()` detects a ZIP header and `mc_open()` on the VFS decompression path fails (`fd1 == -1`):

1. `mcview_close_datasource(view)` is called but datasource is `DS_NONE` (no-op)
2. `mcview_show_error()` for in-panel viewers calls `mcview_set_datasource_string()`, setting datasource to `DS_STRING`
3. No `goto`/`return`, so execution falls through to `mcview_set_datasource_file()` which overwrites datasource to `DS_FILE` without cleaning up `DS_STRING` state

## Sample content of the test file

The following **hex content** produces a file that causes `mcview` to segfault when opened:

`50 4B 03 04 14 00 00 00 08 00`

This corrupts viewer state and causes segfault on subsequent F3.

## Test plan
- [ ] Create a file without extention containing the hex content above.
- [ ] Open this file in mcview
- [ ] F3 - should show raw file content without error dialog
- [ ] Close viewer, press F3 again - no crash

### DIFF
--- a/tests/src/viewer/mcview__load_zip_magic.c
+++ b/tests/src/viewer/mcview__load_zip_magic.c
@@ -1,0 +1,331 @@
+/*
+   src/viewer - tests for mcview_load() with PK/ZIP magic bytes
+
+   Copyright (C) 2025
+   Free Software Foundation, Inc.
+
+   This file is part of the Midnight Commander.
+
+   The Midnight Commander is free software: you can redistribute it
+   and/or modify it under the terms of the GNU General Public License as
+   published by the Free Software Foundation, either version 3 of the License,
+   or (at your option) any later version.
+
+   The Midnight Commander is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#define TEST_SUITE_NAME "/src/viewer"
+
+#include "tests/mctest.h"
+
+#include <fcntl.h>
+
+#include "lib/strutil.h"
+#include "lib/vfs/vfs.h"
+#include "src/vfs/local/local.c"
+#include "src/util.h"
+
+#include "src/viewer/internal.h"
+
+/* --------------------------------------------------------------------------------------------- */
+/* Stubs for symbols pulled in transitively from editor/diffviewer/tty.
+ * These are never actually called in our tests, but the linker needs them
+ * because libinternal.a bundles all of mc's subsystems together. */
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-prototypes"
+
+int mc_skin_color__cache[64];
+void tty_lowlevel_setcolor (MC_UNUSED int color) {}
+void *mc_editor_plugin_list = NULL;
+void mc_editor_plugin_add (MC_UNUSED void *p) {}
+void mc_editor_plugins_load (void) {}
+int button_get_width (MC_UNUSED void *b) { return 0; }
+int we_are_strstrstrbackground = 0;
+
+#pragma GCC diagnostic pop
+
+/* --------------------------------------------------------------------------------------------- */
+
+/* @CapturedValue */
+static int mcview_show_error__call_count;
+
+/* mock: mcview_compute_areas -- no-op in test (depends on widget geometry) */
+void
+mcview_compute_areas (MC_UNUSED WView *view)
+{
+}
+
+/* mock: mcview_update_bytes_per_line -- no-op in test (depends on display areas) */
+void
+mcview_update_bytes_per_line (MC_UNUSED WView *view)
+{
+}
+
+/* mock: mcview_set_codeset -- no-op in test (depends on charset subsystem) */
+void
+mcview_set_codeset (MC_UNUSED WView *view)
+{
+}
+
+/* mock: mcview_display -- no-op in test (depends on tty subsystem) */
+void
+mcview_display (MC_UNUSED WView *view)
+{
+}
+
+/* mock: mcview_show_error -- capture call count instead of showing dialog */
+void
+mcview_show_error (MC_UNUSED WView *view, MC_UNUSED const char *format, MC_UNUSED const char *filename)
+{
+    mcview_show_error__call_count++;
+}
+
+/* mock: file_error_message -- no-op in test */
+void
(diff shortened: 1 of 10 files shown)


mc-worker:
@ilia-maslakov I've cleaned up your test and created a `fix-viewer-zip-crash` branch in this repo. If you are agree with my changes, please take them.

zyv:
> I've cleaned up your test and created a `fix-viewer-zip-crash` branch in this repo. If you agree with my changes, please take them.

I vaguely remember we had a related problem some time ago:

* #4741

What I'm not sure about is what the right way to behave here is. I think the idea before was to show an error message. Now the file is silently opened in the raw mode. Is this desired @mc-worker ? Do you know if this is consistent with other viewer behaviors?

tests/src/viewer/mcview__load_zip_magic.c
```diff
@@ -0,0 +1,355 @@
...
+
+extern int we_are_strstrstrbackground;
+int we_are_strstrstrbackground = 0;
+
+void tty_lowlevel_setcolor (int color) { (void) color; }
+void mc_editor_plugin_add (void *p) { (void) p; }
+void mc_editor_plugins_load (void) {}
+int button_get_width (void *b) { (void) b; return 0; }
+
+#pragma GCC diagnostic pop
```
zyv, line 58: I think this should be handled as it is done here: b6012a239d7df950d2acc258cd0b3c9f42a4f8b3 .

tests/src/viewer/mcview__load_zip_magic.c
```diff
@@ -0,0 +1,355 @@
...
+
+/* @CapturedValue */
+static int mcview_show_error__call_count;
+
+/* mock: mcview_compute_areas -- no-op in test (depends on widget geometry) */
+void
+mcview_compute_areas (WView *view)
+{
+    (void) view;
+}
```
zyv:
```suggestion
mcview_compute_areas (MC_UNUSED WView *view)
{
}
```

tests/src/viewer/mcview__load_zip_magic.c
```diff
@@ -0,0 +1,355 @@
...
+    {
+        if (error != NULL)
+            g_error_free (error);
+        return NULL;
+    }
+
+    if (size > 0)
+    {
+        ssize_t written = write (fd, data, size);
+        (void) written;
```
zyv:
```suggestion
        write (fd, data, size);
```

zyv (changes requested):

tests/src/viewer/mcview__load_zip_magic.c
ilia-maslakov: ok

ilia-maslakov:
> > I've cleaned up your test and created a `fix-viewer-zip-crash` branch in this repo. If you agree with my changes, please take them.
> 
> I vaguely remember we had a related problem some time ago:
> 
>     * [mcview does not open for empty archives (xz, zstd, etc.) #4741](https://github.com/MidnightCommander/mc/issues/4741)
> 
> 
> What I'm not sure about is what the right way to behave here is. I think the idea before was to show an error message. Now the file is silently opened in the raw mode. Is this desired @mc-worker ? Do you know if this is consistent with other viewer behaviors?

It’s better to open the file in raw mode and show at least some result (even if it’s empty) than to interrupt the user with a pop-up error message. This makes the behavior more predictable and less annoying.

tests/src/viewer/mcview__load_zip_magic.c
ilia-maslakov: fixed

ossilator:
> This makes the behavior more predictable and less annoying.

most definitely NOT. you can't seriously argue that _silently_ doing something different than expected is predictable.

if you want to avoid the disruptiveness of a message box, put a red banner above the scrollable area. but then some other mechanism to dismiss the message is required - i guess manually switching to raw mode would do.

tuffnatty:
I must say I also hate the error popup. When I hit it, I have to find a file that the viewer *can* open in the neighborhood, hit F3 again, switch to Raw mode, exit the viewer, navigate to the original file and try again. Is there an easier workflow? It seems a much less disrupting way would be displaying a (colored?) error message in the content area, with the possibility to switch to Raw mode right away.

azazar:
I think that the actual cause of that crash is the format-parameter mismatch introduced here: https://github.com/MidnightCommander/mc/commit/d69da19#diff-88088df415d8479112df2e92572298c06a2420e80fc121403b17ce0dbd55e4e6R394 . "Cannot open <OpenDocument file path> in parse mode" issue is not related to that crash.

mc-worker:
The crash was fixed in #5103/#5105.

PR #5066, opened 2026-03-13
Ticket: #5020: Allocate fewer ncurses color pairs

### LINKED ISSUE #5020: Allocate fewer color pairs (reuse the existing one when only the attributes differ)

```shell
4.8.33
```

### Operating system

```shell
.
```

### How to reproduce

Forking from #5018, in particular https://github.com/MidnightCommander/mc/issues/5018#issuecomment-3926679195: 

This is relevant for terminals that only support 8 colors, and their terminfo describes 64 or so color pairs (`tput pairs`), e.g `TERM=linux` or `TERM=xterm`.

Probably only relevant for ncurses.

---

We allocate a new ncurses color pair whenever we encounter a (fg, bg, attrs) combo that we haven't seen before. This is wasteful, and with only 64 colors pairs available we should be more frugal.

We should only allocate a new color pair whenever the (fg, bg) combo is a new one, and move the attributes out of this game. ncurses only insists on the palette-based approach for the (fg, bg) color pair, and allows to set the attributes independently in `attr_set()` and friends.

So there should be a two-step lookup. Internals of mc identify each different (fg, bg, attrs) combo with an index, I wouldn't change this at this point. This would point to a potentially different ncurses color pair id, as well as the attributes.

In 8-color mode, where we map bright colors to the bold attribute, this mapping should be performed first, and then check if we need a new ncurses color pair. Currently these two steps are performed in the opposite order: first `color.c` `tty_try_alloc_color_pair()`) checks if we've seen the combo, then `color-ncurses.c` `tty_color_try_alloc_lib_pair()` does the bright->bold mapping.

For example, seeing a `red;black` would allocate a color pair. Then seeing a `brightred;black` would reuse the same ncurses color pair (since it's mapped to the same `red;black`), and apply the `bold` attribute.

---

I think we should see an improvement in the rendering of the test case of #5018 in 8-color terminals, more text should be displayed in its desired colors.

We'd still run out of the available color pairs eventually and start to use wrong ones, since as explained in https://github.com/MidnightCommander/mc/issues/5018#issuecomment-3926679195 the number of color pairs supported by ncurses should really be 81 instead of 64, but we'd start to err later, meaning that real life syntax highlighting scenarios will more likely work correctly.

### Expected behavior

.

### Actual behavior

.

### Additional context

_No response_

egmontkob:
## Proposed changes

Reuse ncurses color pairs if possible
    
Allocating a new ncurses color pair for every new (fg, bg, attrs) tuple might result in running out of available color pairs too soon.
    
Instead, allocate a new ncurses color pair only if really needed, i.e. if (fg, bg) is a new combination. Apply the attributes separately.

Plus some minor underlying changes.

* Resolves: #5020

## Checklist

### DIFF
--- a/lib/tty/color-ncurses.c
+++ b/lib/tty/color-ncurses.c
@@ -49,57 +49,78 @@
 
 /*** file scope type declarations ****************************************************************/
 
+typedef struct
+{
+    int pair;   // ncurses color pair index
+    int attrs;  // attributes
+} mc_tty_ncurses_color_pair_and_attrs_t;
+
 /*** forward declarations (file scope functions) *************************************************/
 
 /*** file scope variables ************************************************************************/
 
-static GHashTable *mc_tty_color_color_pair_attrs = NULL;
+/*
+ * Our bookkeeping of the ncurses color pair indices, indexed by the "{fg}.{bg}" string.
+ */
+static GHashTable *mc_tty_ncurses_color_pairs = NULL;
+
+/*
+ * Indexed by mc's color index, points to the mc_tty_ncurses_color_pair_and_attrs_t object
+ * representing the ncurses color pair index and the attributes.
+ *
+ * mc's color index represents unique (fg, bg, attrs) tuples. Allocating an ncurses color pair
+ * for each of them might be too wasteful and might cause us to run out of available color pairs
+ * too soon (especially in 8-color terminals), if many combinations only differ in attrs.
+ * So we allocate a new ncurses color pair only if the (fg, bg) tuple is newly seen.
+ * See #5020 for details.
+ */
+static GArray *mc_tty_ncurses_color_pair_and_attrs = NULL;
+
+static int mc_tty_ncurses_next_color_pair = 0;
+
 static int overlay_colors = 0;
 
 /* --------------------------------------------------------------------------------------------- */
 /*** file scope functions ************************************************************************/
 /* --------------------------------------------------------------------------------------------- */
 
-static inline void
-mc_tty_color_attr_destroy_cb (gpointer data)
-{
-    g_free (data);
-}
-
-/* --------------------------------------------------------------------------------------------- */
-
-static void
-mc_tty_color_save_attr (int color_pair, int color_attr)
-{
-    int *attr, *key;
-
-    attr = g_try_new0 (int, 1);
-    if (attr == NULL)
-        return;
-
-    key = g_try_new (int, 1);
-    if (key == NULL)
-    {
-        g_free (attr);
-        return;
-    }
-
-    *key = color_pair;
-    *attr = color_attr;
-
-    g_hash_table_replace (mc_tty_color_color_pair_attrs, (gpointer) key, (gpointer) attr);
-}
-
-/* --------------------------------------------------------------------------------------------- */
-
 static int
-color_get_attr (int color_pair)
+get_ncurses_color_pair (int ifg, int ibg)
 {
-    int *fnd = NULL;
+    char *color_pair_str;
+    int *ncurses_color_pair;
+    int init_pair_ret;
 
-    if (mc_tty_color_color_pair_attrs != NULL)
-        fnd = (int *) g_hash_table_lookup (mc_tty_color_color_pair_attrs, (gpointer) &color_pair);
-    return (fnd != NULL) ? *fnd : 0;
+    color_pair_str = g_strdup_printf ("%d.%d", ifg, ibg);
+
+    ncurses_color_pair =
+        (int *) g_hash_table_lookup (mc_tty_ncurses_color_pairs, (gpointer) color_pair_str);
+
+    if (ncurses_color_pair == NULL)
+    {
+        ncurses_color_pair = g_try_new0 (int, 1);
+        *ncurses_color_pair = mc_tty_ncurses_next_color_pair;
+#if NCURSES_VERSION_PATCH >= 20170401 && defined(NCURSES_EXT_COLORS) && defined(NCURSES_EXT_FUNCS) \
+    && defined(HAVE_NCURSES_WIDECHAR)
+        init_pair_ret = init_extended_pair (*ncurses_color_pair, ifg, ibg);
+#else
+        init_pair_ret = init_pair (*ncurses_color_pair, ifg, ibg);
+#endif
+
+        if (init_pair_ret == ERR)
+        {
+            g_free (ncurses_color_pair);
+            g_free (color_pair_str);
+            return 0;
+        }
+
+        g_hash_table_insert (mc_tty_ncurses_color_pairs, color_pair_str, ncurses_color_pair);
+        mc_tty_ncurses_next_color_pair++;
+    }
+    else
+        g_free (color_pair_str);
+
+    return *ncurses_color_pair;
 }
 
 /* --------------------------------------------------------------------------------------------- */
@@ -109,6 +130,8 @@ color_get_attr (int color_pair)
 void
 tty_color_init_lib (gboolean disable, gboolean force)
 {
+    int default_color_pair_id;
+
     (void) force;
 
     if (has_colors () && !disable)
@@ -122,8 +145,18 @@ tty_color_init_lib (gboolean disable, gboolean force)
         tty_use_truecolors (NULL);
     }
 
-    mc_tty_color_color_pair_attrs = g_hash_table_new_full (
-        g_int_hash, g_int_equal, mc_tty_color_attr_destroy_cb, mc_tty_color_attr_destroy_cb);
+    // our tracking of ncurses's color pairs
+    mc_tty_ncurses_color_pairs = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+    // ncurses color pair 0 always refers to the default colors; add it to our hash
+    mc_tty_ncurses_next_color_pair = 0;
+    default_color_pair_id = get_ncurses_color_pair (-1, -1);
+    g_assert (default_color_pair_id == 0);
+    (void) default_color_pair_id;  // unused if g_assert is eliminated
+
+    // mapping from our index to ncurses's index and attributes
+    mc_tty_ncurses_color_pair_and_attrs =
+        g_array_new (FALSE, FALSE, sizeof (mc_tty_ncurses_color_pair_and_attrs_t));
 }
 
 /* --------------------------------------------------------------------------------------------- */
@@ -131,8 +164,13 @@ tty_color_init_lib (gboolean disable, gboolean force)
 void
 tty_color_deinit_lib (void)
 {
-    g_hash_table_destroy (mc_tty_color_color_pair_attrs);
-    mc_tty_color_color_pair_attrs = NULL;
+    g_hash_table_destroy (mc_tty_ncurses_color_pairs);
+    mc_tty_ncurses_color_pairs = NULL;
+
+    g_array_free (mc_tty_ncurses_color_pair_and_attrs, TRUE);
+    mc_tty_ncurses_color_pair_and_attrs = NULL;
+
+    mc_tty_ncurses_next_color_pair = 0;
 }
 
 /* --------------------------------------------------------------------------------------------- */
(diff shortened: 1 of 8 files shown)


egmontkob:
How to test:

Install `colors.syntax` and  update `Syntax` as per the opening post of #5018.

`unset COLORTERM`

`export TERM=xterm` or `linux`

`mcedit -S default catalog.colors` (that file is again from #5018)

Should look "obviously broken" without this PR and "obviously fixed" with it.

lib/tty/color-ncurses.c
```diff
@@ -122,24 +107,38 @@ tty_color_init_lib (gboolean disable, gboolean force)
...
-        g_int_hash, g_int_equal, mc_tty_color_attr_destroy_cb, mc_tty_color_attr_destroy_cb);
+    // our tracking of ncurses's color pairs
+    mc_tty_ncurses_color_pairs = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+    // ncurses color pair 0 always refers to the default colors; add it to our hash
+    default_color_pair_id = g_strdup ("-1.-1");
+    mc_ncurses_color_pair = g_try_new (int, 1);
+    *mc_ncurses_color_pair = 0;
+    g_hash_table_insert (mc_tty_ncurses_color_pairs, default_color_pair_id, mc_ncurses_color_pair);
+    mc_tty_ncurses_next_color_pair = 1;
```
zyv: This code... (*)

lib/tty/color-ncurses.c
```diff
@@ -178,22 +177,45 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
...
 
+    color_pair_str = g_strdup_printf ("%d.%d", ifg, ibg);
+
+    mc_ncurses_color_pair =
+        (int *) g_hash_table_lookup (mc_tty_ncurses_color_pairs, (gpointer) color_pair_str);
+
+    if (mc_ncurses_color_pair == NULL)
+    {
+        mc_ncurses_color_pair = g_try_new0 (int, 1);
+        *mc_ncurses_color_pair = mc_tty_ncurses_next_color_pair++;
```
zyv: And this code (*) - can't it be somehow extracted in a function like "remember color pair if not there"?

lib/tty/color-ncurses.c
```diff
@@ -264,3 +286,47 @@ tty_use_truecolors (GError **error)
...
+        mvadd_wchnstr (y + row, x, ctext, cols);
+    }
+
+    g_free (ctext);
+#else
+    (void) y;
+    (void) x;
+    (void) rows;
+    (void) cols;
+    (void) color;
```
zyv, line 347: I know you just moved this code, but could you please use `MC_UNUSED` instead of this hacky ugliness for the stuff we touch? Or maybe even at the header, something like:

```c
#ifdef ENABLE_SHADOWS
tty_colorize_area (int y, int x, int rows, int cols, int color)
#else
tty_colorize_area (MC_UNUSED int, MC_UNUSED int, MC_UNUSED int, MC_UNUSED int, MC_UNUSED int)
#endif
```

(if this works)

egmontkob: Currently mc uses `MC_UNUSED` ~10 times and `(void) variable_name;` ~350 times. 

Due to the `#ifdef` branching nature of things, your proposed variant doesn't look cleaner to me than the current code.

If you don't mind, I'd rather keep this untouched – leave it to you to change if you wish to.

(Your other feedback I agree with and will address.)

zyv: Well, shall we (or I) perform a mass search and replace then? The other items accumulated over time before I eventually introduced `MC_UNUSED` to solve the problem properly because not all warnings can be silenced with the `void` hack. Only our approach with Andrew so far was to avoid code-base-wide changes after non-essential changes in policy...

egmontkob: Can't we just leave `(void)` here?

It works here. Whether at other places a similar hack works or not is irrelevant.

By the way, the refactoring I just did (as per your other request) happened to introduce another such `(void)` trick where `MC_UNUSED` would be incorrect: the value _may or may not_ be used. What to do then?

You have a very hard time convincing me that I have to switch to the new way which is used at 10 places, the old way used at 350 places isn't good anymore.

I'd also like to firmly reject on principle the idea that if I move a method to a different file (where it obviously should've belonged to from the beginning) then I'd definitely need to update the body of the method to our new standards. It's a fair request that newly written code should use the new way, you may or may not mass-update the old syntax, but moving a method to a different location shouldn't count as new code and shouldn't be subjected to the new guidelines.

In this very case, I'm also reluctant to go with this change because it introduces more duplication that the current one, it's easier to get it wrong in one of the cases (which I guess our CI doesn't test, and I wouldn't want to manually test the no-shadow case either beacuse that's not what this change is about); it would also need to go consistently in the header vs. impl file which adds another ugly ifdef-branching; overall I just find it uglier, more fragile than the current code, and would need additional testing. I'd much rather spend my time on things that actually do matter – there's a lot to do.

(Note: Moving the method is a prerequisite for the main commit of the PR; see the "ouch" comment and the compiler warning in the first two drafts attached to the bug.)

lib/tty/color-ncurses.c
egmontkob: Done, curious if you like the new version.

egmontkob:
Note:

In the test file, the last row's almost last columns "base lightgray" and "base white" appear incorrectly. I'll investigate why.

It might be related to the recent removal of the "base" keyword; it might easily be that the test is outdated and the code is fine.

egmontkob:
> In the test file, the last row's almost last columns "base lightgray" and "base white" appear incorrectly.

Fixed.

@horkykuba Could you please review the commit "Fix shady trick..."?

`-1` for "the terminal's defaults" got converted to `-0xff0001` or so, causing troubles further down the line.

Also (I haven't tested this, just looked bad in the code) you explicitly exclude the value 0 from the "shady trick of applying a tint of red", thereby letting ncurses change `#000000` to palette index 0. The problem with this is that palette index 0 might be quite far away from a real black. Going with `#010000` is much safer.

lib/tty/color-ncurses.c
```diff
@@ -122,24 +145,36 @@ tty_color_init_lib (gboolean disable, gboolean force)
...
-    mc_tty_color_color_pair_attrs = g_hash_table_new_full (
-        g_int_hash, g_int_equal, mc_tty_color_attr_destroy_cb, mc_tty_color_attr_destroy_cb);
+    // our tracking of ncurses's color pairs
+    mc_tty_ncurses_color_pairs = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+
+    // ncurses color pair 0 always refers to the default colors; add it to our hash
+    mc_tty_ncurses_next_color_pair = 0;
+    default_color_pair_id = get_ncurses_color_pair (-1, -1);
+    g_assert (default_color_pair_id == 0);
+    (void) default_color_pair_id;  // unused if g_assert is eliminated
```
zyv, line 155: Out of curiosity, do you hit the warning when you compile in some configuration, or is this theory?

I'm asking because I've started eliminating these hacks in a separate branch. In the worst case, I'd have to accept that some of those must remain.

lib/tty/color-ncurses.c
```diff
@@ -164,36 +199,40 @@ tty_color_try_alloc_lib_pair (tty_color_lib_pair_t *mc_color_pair)
...
     }
 
-#if NCURSES_VERSION_PATCH >= 20170401 && defined(NCURSES_EXT_COLORS) && defined(NCURSES_EXT_FUNCS) \
-    && defined(HAVE_NCURSES_WIDECHAR)
-    init_extended_pair (mc_color_pair->pair_index, ifg, ibg);
-#else
-    init_pair (mc_color_pair->pair_index, ifg, ibg);
-#endif
-    mc_tty_color_save_attr (mc_color_pair->pair_index, attr);
+    ncurses_color_pair = get_ncurses_color_pair (ifg, ibg);
```
zyv: I'd ask to do `const int ...` here to limit the scope and mutability, and then Andrew will ask exactly the opposite, because he likes to see all variables at the top of the function like in the old, good Pascal days :-(

Maybe you can at least constify it at the top, though?

lib/tty/color-ncurses.c
zyv: > By the way, the refactoring I just did (as per your other request) happened to introduce another such `(void)` trick where `MC_UNUSED` would be incorrect: the value _may or may not_ be used. What to do then?

In C++, they have recently adopted `maybe_unused` 😢, but I think that in C, there is no solution other than accepting the silencer if you cannot change other code (like moving it inside of the disappearing construct).

In my hierarchy of values, I would then accept the silencer as it is less evil than non-generic (duplicating) logic code.

> You have a very hard time convincing me that I have to switch to the new way which is used at 10 places, the old way used at 350 places isn't good anymore.

It's unfortunate that I should have a very hard time doing so, because you seem to be otherwise thinking rationally and have quite an experience with C.

The problem with unused variable silencers is that they... well, silence the warnings. However, they are not dedicated unambiguous syntax constructs to do so, but at best, some hopefully otherwise harmless construct, which is accepted to be abused for the purpose.

As such, these statements then cannot be easily identified by means other than semantic analysis and don’t directly convey the meaning to the reader.

A proper way to silence warnings if needed is to use dedicated attributes (which is what the new macro is using). It resolves into a directive to the compiler that doesn't and cannot affect anything else, and can be easily found in the code. Also, its meaning is directly obvious.

Moreover, there are places where the `void` trick just doesn't work (like labels) and you **must** use attributes if you need to silence a warning there.

There are several classes of cases when the use of silencers is warranted:

1. Function parameters
2. Function returns
3. Internal variables

The general approach should be to avoid the situation if possible; if not, use it.

The first type of usage I find particularly problematic and want to replace it as a priority

The second is used absolutely inconsistently in our code, and in my opinion should be just thrown away. I don't think that the codebase is ready for strict return checking anyway, and refactoring this properly would be a major project... which might not even make sense in the C world, in terms of what you gain for the amount you invest, unless you're working in some kind of regulated environment. In which case, you might as well reconsider your choice of language.

I can see that in the last case, the macro in its current form can only mark the variable at the declaration site, which doesn't make it much better than the silencer.

Maybe we can improve on that by giving it an optional parameter, so that it would expand only if the parameter evaluates to true. I'm not sure that this is doable within the C macro system and with an amount of effort invested that is worth it.

> I'd also like to firmly reject on principle the idea that if I move a method to a different file (where it obviously should've belonged to from the beginning) then I'd definitely need to update the body of the method to our new standards. It's a fair request that newly written code should use the new way, you may or may not mass-update the old syntax, but moving a method to a different location shouldn't count as new code and shouldn't be subjected to the new guidelines.

I don't think that I can be convinced that there is a problem **on principle** with updating the code as it is touched. In my experience, this is actually the only pragmatic and practical way of evolving huge old code bases in the right direction.

The sweeping changes mostly require an inordinate amount of work and just never get done, and on top of that introduce risks, like mass-upgrades to new types (`size_t` and friends) or removal of dead casts. You've recently witnessed the effects of that sort yourself.

The approach of doing something to the code that is touched when it is touched, even if it has nothing to do with the change per se (obviously in a separate commit), makes the upgrades possible in the first place and then limits the risks, as the change is more likely to be correct due to its limited nature and more stringent review.

Nevertheless, what is certainly true is that it's highly problematic if we don't agree on the nature of the change and if you have some issues "on principle" with the approach itself. I expected neither of those as I made my suggestion, otherwise I wouldn't have made it.

Now, of course, I'm not going to block anything on these grounds. Let's leave it the way you want for this PR.

zyv (approved):
I don't think I have anything else to wish for now 👍

lib/tty/color-ncurses.c
zyv: I do like it much better, thank you 🙏

lib/tty/color-ncurses.c
mc-worker: > he likes to see all variables at the top of the function

At the begin of block.

> like in the old, good Pascal days :-(

like in the old, good C days .

mc-worker (approved):

lib/tty/color-ncurses.c
egmontkob: > Out of curiosity, do you hit the warning when you compile in some configuration, or is this theory?

It's experience (probably mostly from VTE/gnome-terminal).

Look at `glib/gtestutils.h`: under an `#ifdef G_DISABLE_ASSERT` these `g_assert()` and friends are eliminated and don't evaluate (don't even look at) their parameter.

I have also shot myself in the foot with code like `g_assert (important_function_call_to_make() == 0);` – works great until you disable asserts.

Now, maybe there's no set of configure option to pass to mc which would set this macro, but still, it's bad practice to rely on it.

lib/tty/color-ncurses.c
egmontkob: > The approach of doing something to the code that is touched when it is touched, even if it has nothing to do with the change per se (obviously in a separate commit)

As long as it's a trivial improvement, why not.

In this case it's a combination of two major things:

- I do not really consider moving an entire method to a different file – but otherwise leaving it unchanged – as "touching it".

- The requested change hardly has any precedence in the source, IMO is absolutely not obvious that it makes the code cleaner (maybe it makes it semantically better, but visually uglier), and if I wanted to be really sure that I didn't make a silly mistake then it would take a bit of additional testing as well.

I hope you don't mind that I just decided to draw the line before this step now. :)

lib/tty/color-ncurses.c
egmontkob: > I'd ask to do const int ... here to limit the scope and mutability

Done – primarily because the line underneath already did the same, so it's now more consistent.

> Maybe you can at least constify it at the top, though?

I'm surely misunderstanding you here... declare as `const` at the top and assign the value later? That's not how `const` works :-D

egmontkob:
Changed `tty_color_deinit_lib()` to reset more variables to their defaults; for consistency and increased robustness.

Will wait for a few days for @horkykuba to possibly chime in, before merging.

PR #5067, opened 2026-03-14
Tree-sitter integration for better syntax highlighting

jtyr:
## Summary

- Adds tree-sitter-based syntax highlighting as an alternative to MC's regex-based system
- 61 languages supported with AST-accurate parsing (no more broken highlighting from unmatched quotes or nested constructs)
- Language injection for HTML (JavaScript + CSS) and Markdown (inline elements + fenced code blocks with per-language highlighting)
- Shared library build (default): each grammar is a separate `.so` module (~5 MB mc binary)
- Static build option: all grammars linked into the binary (`--with-tree-sitter-static`)
- Falls back to legacy `.syntax` highlighting when a tree-sitter grammar is not available

## Motivation

MC's regex-based syntax highlighting is fundamentally limited by its line-oriented pattern matching. Languages with complex nesting, heredocs, multi-line strings, or context-dependent syntax (Bash, Perl, Ruby, Python, and many others) are notoriously difficult to highlight correctly with regexes. Users regularly encounter broken highlighting that propagates through the rest of the file from a single unmatched delimiter.

Tree-sitter solves this by parsing the actual AST of each language. Highlighting is always structurally correct because it operates on the parse tree, not on regex patterns. Incremental re-parsing makes it fast enough for interactive editing where only the changed portion of the tree is rebuilt.

Beyond correctness, tree-sitter enables features that are impossible with regex-based highlighting. Language injection allows one grammar to delegate to another: HTML files get proper JavaScript highlighting inside `<script>` tags and CSS inside `<style>` tags. Markdown fenced code blocks are highlighted according to their language tag. For example a ` ```python ` block gets full Python highlighting, ` ```bash ` gets Bash, and so on for any installed grammar. This kind of multi-language highlighting is a natural fit for tree-sitter's AST approach and opens the door for similar injection support in other templating and polyglot file formats.

## Build options

```
# Shared mode (default) - grammars as .so modules, mc binary stays ~5 MB
./configure --with-tree-sitter

# Static mode - grammars linked into binary
./configure --with-tree-sitter --with-tree-sitter-static

# Build only specific grammars
./configure --with-tree-sitter --with-tree-sitter-grammars=c,python,bash

# Without tree-sitter (default, unchanged behavior)
./configure
```

Grammar sources are automatically downloaded from upstream repositories during configure (pinned to specific commit SHAs for reproducibility).

## Packaging

In shared mode, each grammar produces two files:

- `mc-ts-<name>.so` -- loadable grammar module (installed to `$(libdir)/mc/ts-grammars/`)
- `<name>-highlights.scm` -- highlight query file (installed to `$(datadir)/mc/syntax-ts/queries/`)

Distros can package these as separate optional packages (e.g., `mc-ts-grammar-python`), letting users install only the languages they need. When a grammar module is not installed, MC transparently falls back to legacy regex highlighting for that language.

Users can also build and install custom grammar modules in `~/.local/lib/mc/ts-grammars/` (with query files in `~/.local/share/mc/syntax-ts/queries/`) without root access.

## Note on grammar sources

This feature requires building tree-sitter grammar sources (downloaded C files from upstream projects) into shared libraries. This adds a build-time dependency on `libtree-sitter` and `gmodule-2.0`, and the responsibility of maintaining version pins for 61 grammar repositories. I believe the benefit of reliable, AST-based syntax highlighting, eliminating the class of bugs that regex highlighting cannot solve, justifies this added maintenance scope. The grammar sources are not vendored in the repository; they are downloaded at configure time and pinned to specific commits, making updates straightforward.

## Test plan

- [ ] `./configure --with-tree-sitter && make && make check` (shared mode, all grammars)
- [ ] `./configure --with-tree-sitter --with-tree-sitter-static && make && make check` (static mode)
- [ ] `./configure && make` (without tree-sitter, unchanged behavior)
- [ ] Open files in various languages and verify highlighting matches MC's legacy colors
- [ ] Open an HTML file with `<script>` and `<style>` tags - JS and CSS should be highlighted
- [ ] Open a Markdown file with fenced code blocks - code should be highlighted per language
- [ ] `mc -e file` and execute a macro script - should not hang anymore

See `doc/TREE-SITTER` for full documentation.


src/editor/edit.c
```diff
@@ -1789,10 +1789,19 @@ edit_user_menu (WEdit *edit, const char *menu_file, int selected_entry)
...
 
         // was block file created or modified by menu command?
+        // Use nanosecond-precision mtime when available, because scripts that only
+        // rearrange bytes (e.g. word-wrap replacing spaces with newlines) produce
+        // the same file size and may finish within the same second.
         modified = (!status_before_ok && status_after_ok)
             || (status_before_ok && status_after_ok && status_after.st_size != 0
                 && (status_after.st_size != status_before.st_size
-                    || status_after.st_mtime != status_before.st_mtime));
+#ifdef HAVE_STRUCT_STAT_ST_MTIM
```
mc-worker: Please move this to the separate commit.

src/editor/editcmd.c
```diff
@@ -1939,6 +1939,19 @@ edit_block_process_cmd (WEdit *edit, int macro_number)
 {
     char *fname;
     char *macros_fname = NULL;
+    off_t start_mark, end_mark;
```
mc-worker: Please move this to the separate commit.

src/editor/editwidget.h
```diff
@@ -142,7 +142,20 @@ struct WEdit
 
     unsigned int skip_detach_prompt : 1;  // Do not prompt whether to detach a file anymore
 
-    // syntax highlighting
+    // syntax highlighting (tree-sitter)
```
mc-worker, line 145: Can it be moved to the separate structure?

src/editor/syntax.c
```diff
@@ -149,12 +158,102 @@ typedef struct
     edit_syntax_rule_t rule;
 } syntax_marker_t;
 
+#ifdef HAVE_TREE_SITTER
```
mc-worker: Can it be moved to the separate file?

src/subshell/common.c
```diff
@@ -1901,7 +1901,7 @@ invoke_subshell (const char *command, int how, vfs_path_t **new_dir_vpath)
         // some sort of error.
         if (use_persistent_buffer)
             clear_cwd_pipe ();
-        else
+        else if (mc_global.mc_run_mode == MC_RUN_FULL)
```
mc-worker: Please move this to the separate commit.

configure.ac
```diff
@@ -41,6 +41,7 @@ dnl that ar cannot be found and linking via libtool will fail at a later stage
 AC_CHECK_TOOLS([AR], [ar gar])
 
 AC_PROG_CC
+AC_PROG_CXX
```
mc-worker, line 44: Do we really need C++ compiler?

configure.ac
```diff
@@ -536,11 +537,171 @@ AC_SUBST(CPPFLAGS)
 AC_SUBST(LDFLAGS)
 AC_SUBST(LIBS)
 
+dnl ############################################################################
```
mc-worker, line 540: Please move this to the separate m4-file.

mc-worker (commented):
Quick view. Not related to the tree-sitter itself, but to the your PR content.

Please remove *.po and Makefile.in from your PR,

src/editor/edit.c
jtyr: I will create a separate PR for this.

src/editor/editcmd.c
jtyr: I will create a separate PR for this (together with the change in `src/editor/edit.c`.

src/editor/editwidget.h
jtyr: I have moved it to separate structure.

src/editor/syntax.c
jtyr: I have separated the TS syntax handling.

src/subshell/common.c
jtyr: I will create a separate PR for this.

configure.ac
jtyr: Yes, this is needed for the SQL grammar which has a C++ scanner.

configure.ac
jtyr: I have moved that to a separate m4 file.

jtyr:
The [last commit](https://github.com/MidnightCommander/mc/pull/5067/changes/413006cbe4fca23025cf8c1f238cf4acb938069a) demonstrates how simple it is to add an extra language. I have added QML language that's requested in the PR #5022. All known parsers are listed at [this page](https://github.com/tree-sitter/tree-sitter/wiki/List-of-parsers).

zyv:
Thanks for the PR!

Unfortunately, I don't think it is seriously reviewable as a blob, but it's nice to have a prototype on the basis of which we can have specific discussions and decide how we can move forward, instead of just hand-waving and doing some theoretical exercises.

Before we get into the technical details, I would like to discuss some general design and maintenance issues.

First off, I think it's certainly desirable to replace our internal regex-based highlighting engine if at all possible with something a little bit more maintainable and powerful (more grammar-based at least to the extent that nested constructs are supported). The priorities to my mind should be:

1. Usable both in editor and viewer
2. Written in C, slim, and if externally maintained, then vendorable
3. Available on all supported configurations including embedded systems
4. Supporting most common languages, ideally uncomplicated to add new languages

Non-goals would be:

1. Ultimate highlighting quality

I would absolutely like to avoid to maintain several systems in parallel, or connections to several systems. This PR comes directly after #5065 - not sure if that's a coincidence. Anyways, I think we have the following options:

1. Reject both, insist on improving CoolEdit engine to make it usable for both viewer and editor, at least go that far in the grammar direction as to support nested constructs, make code clean, maintainable, and testable
  - Realistically, I guess this is never going to happen, although probably the most appealing option
2. Accept both, ending with 3 different highlighting systems: CoolEdit and Tree Sitter for editor, a bunch of external ones for the viewer bolted on via SGR1 parser
  - I don't want to have to maintain this
  - Bad user experience leading to maximally inconsistent highlighting between competing parts of the program
3. Accept Tree-Sitter, extending it to the viewer
  - In this case, I would like to remove the CoolEdit engine even with some loss of functionality

I think (1) will happen on its own if our discussions end up nowhere, (2) is not something I can accept, and (3) maybe is what we should aim for and discuss how we get there.

I've had a cursory look at Tree Sitter (I haven't heard about it before). I think it looks pretty much like highlighting done the right way, although it still seems to not be without difficulties in the context of mc's requirements.

My first question is, are there any comparable alternatives at all, that is highlighting libraries that have a slim C core runtime and can generate parsers in C from grammars, or is it pretty much the only contender in this space?

If Tree Sitter is the only alternative, then I would like to clarify the integration aspects.

---

You mention that adding parsers for languages inflates the binary. How bad it really is and what do the numbers mean? I'm surprised about ~5M baseline for `mc`. On macOS, without optimization, with debugging information and unstripped the executable is about ~1M. I think we should consider only something equivalent to `-O3` and without debugging information. The distributions split it anyways, and ones looking for savings do more to the binaries.

So with `-O3` and and stripped, what would be the real difference between static Tree Sitter and normal build? If the difference is "big", why is that, can it be somehow optimized? The parsers should have a lot in common. Can the parts of the code not be re-used?

Now, adding dynamically loaded parser libraries is not something I would accept. We've been moving in the opposite direction for years, and this would be a huge setback. If this is non-negotiable, this would kind of kill Tree Sitter for me.

--- 

The next big issue for me is the grammar fetching magic at configure time. This is absolutely unacceptable. We should be able to bootstrap self-contained tarballs, which have everything in them that is needed for the build.

My understanding is that it's not impossible to vendor Tree Sitter and parsers. Is there any reason why you didn't consider that? If it is possible, then we should think of what vendoring strategy is best.

How often does Tree Sitter runtime change and how big is that in terms of source code? Can we just get away with committing it to a subdirectory?

Same question regarding the parsers, can we just commit them or is it dozens of megabytes of generated source code changing often that will completely blow up our repository?

Are there any other meaningful approaches to that, as in create a separate repository for vendored Tree Sitter stuff and include it as a submodule and/or fetch a ZIP of a hash at the `./autogen.sh` stage?

I understand your ideas about pinning to tags being reasonably secure, but we have other issues to consider. Our main git repository is on GitHub and we have mirrors. I want us to have reasonable control over what we depend upon. Yes, we already lost control of many things. But that's no reason to exacerbate the situation by several orders of magnitude. 

What happens if Gitlab-hosted projects shut down, like it happened to BitBucket? What happens if one of the grammar maintainers stops taking his pills and nukes his repository? Someone takes over and changes the tag, because it was an unsigned one? When relying on so many moving parts, I'm afraid, we'll be dealing with this stuff all the time.

Also, if we come up with a good vendoring strategy (separate repository?), I would like us to have something in the CI generating a report of what the state of vendored grammars are, and ideally semi-automated updates, which we could e.g. do once before release, or immediately after releases, or sometime in the middle of the cycle without it being a major project like tar updates.

Any ideas how this can be done cheaply and reliably?

---

Finally, regarding grammar support, I would actually like to have a comparison table between what we have now and Tree Sitter. We have about ~100 definitions of variable quality. I understand that Tree Sitter brings in ~60, including community stuff.

How much do we lose? How bad is that? Can you make a tabular overview? Is it realistic to lose as little as possible, or are there just no grammars for what we have, which is why you picked built-in, high-quality community grammars and stopped at that?

---

And one final point, requiring a C++ compiler is absolutely a no-go.

jtyr:
Thank you for the thorough and thoughtful review, @zyv. These are exactly the right questions to ask, and I'm glad we're having this discussion with a working prototype rather than in the abstract.

I'll try to address each point. The full documentation is in `doc/TREE-SITTER` in the branch if you'd like more detail on any specific aspect.

## Why tree-sitter?

> My first question is, are there any comparable alternatives at all?

No. Tree-sitter is uniquely positioned for MC's needs:

- **C runtime** (~200KB shared library, ~15 source files) - no foreign language runtimes
- **Incremental parsing** - re-parses only the changed portion of the tree, fast enough for real-time editing
- **Generated C parsers** - each grammar compiles to plain C (one `parser.c` + optional `scanner.c`)
- **Large ecosystem** - 200+ grammars maintained by the community, covering all mainstream languages
- **Battle-tested** - used by Neovim, Helix, Zed, GitHub's code search, and many others

The alternatives don't fit: TextMate grammars are regex-based (same fundamental limitations as CoolEdit), Pygments is Python-only, Lezer is JavaScript-only. Tree-sitter is the only C-native option with grammar-based parsing.

## Option 3: Replace CoolEdit with tree-sitter

I agree this is the right direction. The tree-sitter integration already falls back to legacy highlighting when a grammar is unavailable, so the transition can be gradual:

1. Ship tree-sitter alongside CoolEdit (current state)
2. As grammar coverage improves, mark CoolEdit as deprecated
3. Eventually remove CoolEdit once the remaining gaps are filled

## Viewer support

Tree-sitter can absolutely be extended to the viewer. The infrastructure is the same - `ts_grammar_registry_lookup()` finds the grammar, `ts_query_new()` compiles the highlight query, and `ts_parser_parse_string()` produces the tree. The viewer would parse the file once (or incrementally as the user scrolls), run the query cursor over the visible range, and produce color entries. The viewer's rendering would consume these the same way the editor does. The core tree-sitter code in `syntax_ts.c` is already separated from the editor-specific parts, which makes this extension straightforward.

## Language coverage

Here's a comparison of what MC has today vs what tree-sitter covers:

| | Count |
|---|---|
| MC CoolEdit languages | ~104 |
| Tree-sitter languages in this PR | 63 (62 + markdown_inline) |
| MC-only (would lose highlighting) | ~51 |
| Tree-sitter only (new) | 3 (HCL/Terraform, QML, Scala) |

The ~51 MC-only languages are mostly niche or domain-specific formats:

- **Packaging/distro**: Debian control/changelog, RPM spec, Ebuild, YUM repo
- **Obscure/legacy**: Eiffel, Nemerle, Jal, Yabasic, B language, J language
- **Domain-specific**: POV-Ray, Spice circuits, D-Link switch configs, PIC linker scripts
- **Internal**: MC's own syntax/file highlighting definition formats

The notable gaps are **LaTeX** (no tree-sitter grammar with pre-built parser.c exists - only a scanner), **DOS Batch**, **M4 macros**, and **Puppet**. These could be added over time as tree-sitter grammars become available, or the community could contribute them.

All mainstream programming languages are covered with significantly better highlighting quality than CoolEdit - especially languages with complex syntax like Bash (heredocs, nested quoting), Perl, Ruby, and HTML (embedded JS/CSS).

The full list of available tree-sitter parsers is at https://github.com/tree-sitter/tree-sitter/wiki/List-of-parsers - there are 200+ grammars, many of which could fill the gaps.

## Binary size

Here are the real numbers (Linux x86_64, `-O2`, stripped):

| Build | Size |
|---|---|
| MC without tree-sitter | 1.1 MB |
| MC + 22 common grammars (static) | 18 MB |
| MC + all 63 grammars (static) | 100 MB |

The size comes from the generated `parser.c` files. Some grammars are enormous:

| Grammar | parser.c size |
|---|---|
| Verilog | 46 MB |
| Fortran | 36 MB |
| OCaml | 36 MB |
| SQL | 36 MB |
| C# | 35 MB |
| COBOL | 31 MB |

A curated set of ~22 common languages (C, C++, Python, Bash, Go, Rust, Java, JS, TS, JSON, YAML, HTML, CSS, XML, Lua, Ruby, PHP, Perl, Markdown, Make, Dockerfile) comes to 18 MB stripped. Dropping the largest grammars (Verilog, Fortran, COBOL) helps, but doesn't eliminate the issue - modern languages like C# and Kotlin also have large parsers because the generated code is inherently proportional to grammar complexity.

The `--with-tree-sitter-grammars=LIST` option already allows selecting exactly which grammars to include, so distributions can choose their own subset.

## Dynamic loading

> Adding dynamically loaded parser libraries is not something I would accept.

I understand, and I respect the project's direction. Could you help me understand the reasoning? The grammar module use case seems quite different from general plugin/VFS module loading:

- Grammar `.so` files are pure data (parser tables + a single function), not code that hooks into MC internals
- Missing modules are silently ignored (fallback to legacy highlighting)
- No ABI stability concern - grammars only use the tree-sitter API, not MC's
- The size difference is dramatic: 1.1 MB (shared) vs 18-100 MB (static)
- It enables per-language packaging, which is how both Neovim and distros handle this

That said, if static is the firm decision, then `--with-tree-sitter-grammars=LIST` makes it workable. The distribution maintainer picks a reasonable default set, and users who want more can rebuild.

## Grammar fetching and vendoring

> The grammar fetching magic at configure time is absolutely unacceptable. We should be able to bootstrap self-contained tarballs.

Agreed. The configure-time download was added purely for developer convenience and is not essential. Grammar sources can be provided from any local directory via `TREE_SITTER_GRAMMARS_DIR=/path/to/grammars` at configure time - this is already implemented and documented.

For vendoring, I'd propose a separate repository (`mc-tree-sitter-grammars`) that:

1. Contains all grammar sources (the `parser.c` + `scanner.c` files, ~478 MB total)
2. Has a CI workflow that periodically checks upstream grammar repos for new tags, downloads updates, compiles, and runs the query validation tests
3. Produces versioned release tarballs (e.g., `mc-tree-sitter-grammars-2026.1.tar.gz`)
4. MC's `./autogen.sh` (not `./configure`) fetches a pinned tarball from this repo, or it's included in `make dist` output

This way:
- `make dist` produces a self-contained tarball - no network access needed to build from it
- MC controls exactly which grammar versions are included via the pinned tarball hash
- Grammar updates are vetted by CI before being tagged
- If any upstream grammar repo disappears, the vendored copy in `mc-tree-sitter-grammars` is unaffected
- No git submodule complexity - just a tarball URL and a hash in `autogen.sh`

## Tree-sitter runtime vendoring

> How often does Tree Sitter runtime change and how big is that in terms of source code?

The tree-sitter runtime is small (~15 C source files, ~200KB compiled) and very stable. The C API (`tree_sitter/api.h`) has been backward-compatible for years. The current ABI version is 14-15. We already vendor the internal headers (`parser.h`, `alloc.h`, `array.h`) that grammars need but the system package doesn't install.

Vendoring the entire tree-sitter runtime into MC's source tree is feasible and would eliminate the `libtree-sitter` system dependency. It would add roughly 15 C files to the build. For reference, Neovim uses tree-sitter as a system shared library dependency, but for MC's conservative approach, vendoring may be more appropriate.

## C++ compiler requirement

> Requiring a C++ compiler is absolutely a no-go.

Only one grammar (SQL) has a C++ scanner (`scanner.cc`). All other 62 grammars are pure C. We can drop SQL to eliminate the C++ requirement entirely. If SQL support is desired later, the scanner could potentially be rewritten in C - it's typically a small file (~200 lines).

## Next steps

Based on your feedback, I think the path forward is:

1. Create `mc-tree-sitter-grammars` repository with CI-automated grammar updates and release tarballs
2. Remove configure-time downloading, replace with `autogen.sh` tarball fetch + `TREE_SITTER_GRAMMARS_DIR` for builds from tarballs
3. Drop SQL (C++ scanner) unless the scanner can be rewritten in C
4. Consider vendoring the tree-sitter runtime to eliminate the system library dependency
5. Keep the `--with-tree-sitter-grammars=LIST` for static builds to manage binary size
6. Plan viewer integration as a follow-up

I'm happy to discuss any of these points further or adjust the approach based on your preferences.

ossilator:
this is all just meta; i didn't look at the code.

the idea to vendor the runtime lib is nonsense. afaict, it would be the first external lib that mc vendors.

but i don't understand why the grammars have to be compiled by every user, rather than just being installed as binaries (plug-ins for a loader included in the runtime). this is totally bonkers from a packaging perspective, esp. considering how huge these things are. yet, debian ships only with -src packages (for a select set of grammars). [the faq](https://github.com/sogaiu/ts-questions/blob/master/questions/should-parser-source-be-committed/README.md) totally fails address this fundamental question.
i can think of two reasons:
- plug-in systems are platform dependent, which results in a lot of ugly code. as TS wants no dependencies, it can't just use e.g. glib's module stuff. but this would be addressable by a separate library that offers this "service". isn't there one?
- the grammars might have compile-time options that cannot be implemented as runtime options without major performance impact.

the hand-waving about the yet-unsupported grammars is a tad unconvincing. porting mc's (generally very simple) syntax defs sounds like a relatively minor (though probably boring) task, and a commitment to doing so (and therefore enabling the elimination of the internal engine) would be quite a boost in my eyes.

similarly, a commitment to actually port the viewer would also be a major boost.

there is a flip side to eliminating the internal engine: given how huge the TS grammars are, this step would imply the removal of syntax hl for resource constrained systems. well, actually, there is a middle way: syntaxes that are actually relevant for such systems (we're talking home routers and such) are very simple (config files of various types) and could actually be included. so non-issue, i guess?

i don't understand yuri's opposition to c++, esp. as an optional build dependency for 3rd-party components. the only situation where this would have a _practical_ impact is if a new-ish c++ version is required, as some niche/legacy systems fail to deliver in that regard. and of course, this is relevant only if the distribution question above does not get a good answer.

jiri, your answer has a whiff of chatgpt. i don't mind if you use it, but it has a bad taste when it's recognizable as such.

we like a nice clean git history. you are likely to end up in merge hell if you don't start cleaning up early on during the major refactoring of this PR.

jtyr:
Thanks for the feedback, @ossilator.

On the shared libraries point - yes, that's exactly what the prototype already implements. Each grammar compiles to a standalone `.so` exporting a single `tree_sitter_<name>()` function. MC loads them on demand via `g_module_open()` (GLib's GModule, which MC already depends on). The loader is about 100 lines of code. Missing grammars are silently skipped with fallback to CoolEdit. So distros could package each grammar as a separate optional package - just an `.so` file and an `.scm` query file per language. Users install only what they need.

Regarding why grammars need to be compiled rather than distributed as binaries - they don't, in principle. The `.so` files are the binary form. The question is who compiles them. Neovim delegates this to `nvim-treesitter` which downloads and compiles grammars at runtime. We could theoretically reuse that, but it's tightly coupled to Neovim and depends on Node.js tooling. I think MC is better off owning its grammar build pipeline - a dedicated repo with CI that produces tested release tarballs of `.so` files. That way `make dist` (or a distro build) can include prebuilt binaries without end users needing to compile anything.

On vendoring the tree-sitter runtime - I agree, that was not a good suggestion. Keeping it as a system dependency (`libtree-sitter`) is the right approach, consistent with how MC handles other libraries.

On the missing grammars - I can commit to expanding coverage from the existing tree-sitter grammar ecosystem (200+ grammars available). Most of the ~51 MC-only syntaxes are simple keyword/string/comment definitions that translate straightforwardly to tree-sitter queries, provided a grammar exists. For the handful of truly niche formats where no grammar exists, keeping CoolEdit as a fallback seems reasonable. CoolEdit is also lighter on CPU/memory than tree-sitter parsing, which could matter on embedded systems. Having `--with-tree-sitter` as an opt-in feature means resource-constrained builds can stick with CoolEdit entirely. I'd leave this decision to the maintainers.

On the viewer - I can commit to extending tree-sitter highlighting to the viewer. The `syntax_ts.c` separation was done with this in mind. The infrastructure (grammar lookup, query compilation, highlight cache) is reusable; it just needs a different integration point in the viewer's rendering path.

On C++ - only one grammar (SQL) needs it. Happy to drop it or keep it, depending on what the project prefers.

On AI usage - yes, I used AI assistance for both the implementation and drafting the previous response. The design decisions are mine but the tooling helped enormously with the volume of work. I find the structured format contributes to better readability even if it has that recognizable feel.

On git history - I can squash and rebase the commits if that's how the project prefers PRs to be structured. Just let me know when the design discussion settles and I'll clean it up before final review.

ossilator:
yeah, i know that this is what you implemented, and i appreciate it (i also don't understand why yuri doesn't like the idea). my question was basically: why isn't this upstream? they could even provide multiple loaders (in separate libraries) if they don't want to commit to a particular toolkit, provided the resulting plugins are binary compatible. if they fail to do that, they work against their own adoption, given that each user needs to a) obtain, b) build, and c) install an own copy. i hate that as a user (both developer and end-user), and it goes against (some) distros' policy to unbundle dependencies. mind researching this further and potentially raising it with upstream?

jtyr:
Good question @ossilator. I did some more digging.

Tree-sitter upstream provides no loader or standard distribution mechanism. Their [documentation](https://tree-sitter.github.io/tree-sitter/using-parsers/1-getting-started.html) only covers compile-time integration. Each editor rolls its own: Neovim compiles grammars at runtime via [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter), Helix compiles at build time into the binary, and Emacs has a third-party project ([tree-sitter-langs](https://github.com/emacs-tree-sitter/tree-sitter-langs)) that publishes pre-built `.so` bundles.

The Emacs bundle is interesting - it has 120 grammars as `.so` files, built via CI for Linux/macOS/Windows (x86_64 + aarch64), ~13 MB per platform. The symbols are identical to ours (`tree_sitter_python`, etc.), so they're technically loadable. But there's a catch: query files (the `.scm` highlight rules) are tightly coupled to specific grammar versions because node names might change between releases. If the grammar version in the Emacs bundle doesn't match what our query file expects, highlighting silently breaks. So reusing someone else's pre-built binaries only works if you also pin the query files to the same grammar versions.

The [tree-sitter CLI](https://tree-sitter.github.io/tree-sitter/cli/build.html) can also build grammars (`tree-sitter build .` in a grammar repo produces a `.so`), but that still requires obtaining the source and having a C toolchain.

I think the real gap isn't the loader (50 lines of `g_module_open` as you say) but the lack of a standard convention for grammar location and version pinning. Each editor needs its own query files matched to specific grammar versions, which is why everyone ends up building their own pipeline. A shared grammar registry with versioned query bundles would be the proper solution, but that's a tree-sitter ecosystem problem beyond what we can solve for MC.

For MC specifically, I think our best path is the separate `mc-tree-sitter-grammars` repo with CI that builds both the `.so` files and ships the matching `.scm` query files together as a versioned bundle. That gives us version consistency and takes the build burden off end users.

I could raise the general distribution question with tree-sitter upstream if that's useful, though I suspect the answer will be that they consider it an editor concern rather than a core library concern.

ossilator:
the queries are application-specific, but syntax hl is part of the ts core, so installing the matching .scm files along with the .so files seems rather obvious at first sight.

that emacs project's mission statement is basically exactly what we need, though in practice they intentionally deviate from the upstream queries. so go figure. :thinking: 

you are quite correct that this is a ts ecosystem problem.
but it's somewhat crucial to address it to increase ts' acceptance by projects to which ts' added value doesn't justify the price that needs to be paid given the status quo. my role in mc is only advisory, but i strongly feel that mc falls into that category.
and it's clearly a problem that doesn't solve itself (i'll note here that ts is 13 years old and has already "survived" one lead developer change). so if you want to push things forward in mc, it would be advisable to take a leading role on the ts side as well, and maybe even become a maintainer of that sub-project.

from the practical side, i'd probably fork the emacs project into the ts namespace and "de-emacsify" it. the emacs project in turn would then need to pin itself to specific versions of the central project, so they could ship matching vendored .scm files in an emacs-specific directory.

plan b for mc would be to simply use the emacs project verbatim, but this doesn't feel quite right.

having our own build integration is at best plan c as far as i'm concerned.

data-man:
Thank you, it looks very promising for the future. I'd be nice to add an editor option to disable tree-sitter.

I tried editing sqlite3.c (~10 MB), and it took a very long time to open it.

And `tree-sitter-c` couldn't even parse it,  so the parsing was done the same way as before.

jtyr:
Update since the last review.

Majority of the feedback has been addressed - TS code moved to separate files (`syntax_ts.c`/`syntax_ts.h`), TS fields in a separate struct, configure macro in its own m4 file. The two unrelated bug fixes were submitted separately as #5070 and #5071.

Grammar sources come from a separate [tree-sitter-grammars](https://github.com/jtyr/tree-sitter-grammars) repository that builds and publishes versioned release tarballs with pre-built `.so`/`.dylib`/`.dll` modules and query files for five platforms (Linux x86_64, Linux aarch64, macOS aarch64, macOS x86_64, Windows x86_64). The download script (`scripts/ts-grammars-download.sh`) fetches either source tarballs (for static builds) or pre-built shared libraries, pinned to a specific release version for reproducible builds. The grammars repo has its own CI - weekly upstream update checks, automated PR creation with build/validate across all platforms, and auto-merge.

There are two build modes to suit different packaging needs. Shared mode (default) keeps the `mc` binary small at 5.1 MB, Each grammar is a separate `.so` module loaded on demand, letting distros package grammars individually (e.g. `mc-ts-grammar-python`). Users can also install custom grammars in `~/.local/lib/mc/ts-grammars/` without root. Static mode links selected grammars directly into the binary for environments where shared libraries aren't practical (7.7 MB with 3 grammars, 10 MB with 10). All four build configurations have been tested: without TS (5.0 MB), shared (5.1 MB), static with subset, static with all 63 grammars (109 MB, for build validation only).

The status bar now shows `S:[TS]`/`S:[Legacy]`/`S:[None]` reflecting the actual mode per file. Manual syntax selection from the Options menu works with TS too.

Added a config-driven wrapper grammar system for template languages. A single config file (`misc/syntax-ts/wrappers`) defines which grammars can wrap which host languages. No hardcoded grammar names in C code. This makes Helm charts work nicely: a `.yaml` file with `{{ }}` template syntax automatically falls back to Go template highlighting with YAML injected into the text portions. Compound extensions like `README.md.gotmpl` also work. Injections are recursive up to 3 levels, so fenced code blocks inside Markdown inside Go templates all get proper highlighting.

Parse errors show as red, and if the parser completely fails (like with macro-heavy C amalgamation files), it gracefully falls back to partial highlighting instead of painting everything red.

Created sample files and comparison reports for 63 languages. Eight got query fixes to better match legacy colors (Ada, C++, Fortran, GLSL, IDL, Meson, OCaml, TCL). Found two broken upstream grammars. Puppet has pervasive parse errors and Swift has an incompatible ABI version.

The Go template grammar includes 240 builtin functions (Go template + Sprig + Helm) with proper color distinction between builtins, user functions, fields, and variables.

Fixed the scanner list in `configure.ac` and split the typescript entry in the grammars repo into proper `typescript` and `tsx` entries. There's a `mc-syntax-dump` tool in `tests/syntax/` with its own `Makefile` for comparing legacy vs TS output side by side. Everything is documented in `doc/TREE-SITTER`.

ossilator:
the PR still contains lots of stuff that is (or should be) in the external repo, so i'll wait until you clean that up (and rewrite the history).

the rest concerns the external repo. firstly, i'm impressed - you went full in. but:
- did you talk to anyone in the ts community (and can point to relevant discussions)? it would be a pity if this turned into an [xkcd 927](https://xkcd.com/927/)-like situation, which distributors would certainly hate as much as i would.
- two things about the repo contents itself:
  - why didn't you include the grammar sources? that would be required for proper guarding against upstream deletion, and some distros would want to run the generator step themselves out of principle. the sources are small compared to the generated files, so that wouldn't be a biggie.
  - why did you duplicate the ts headers in every grammar? i'd create a top-level tree-sitter/ directory, with one subdirectory for each supported abi version.

jtyr:
Thanks for looking into my recent commits, @ossilator.

Regarding files in the PR, the grammar sources (`parser.c`, `scanner.c`) and upstream query files are `.gitignore`d and not part of the PR. They need to be downloaded before building using `scripts/ts-grammars-download.sh`. The files that are in the PR are only the build infrastructure: `Makefile.am` for static mode, `ts-grammar-registry.h`, compatibility headers in `tree_sitter/` (needed because system libtree-sitter only installs `api.h`, not the internal headers that grammars `#include`), the download script, and MC-specific query overrides that tune colors to match MC's scheme. Could you point out which specific files you think should move out?

Regarding TS community discussions, I haven't discussed this with the TS community directly. There are already several query standards (neovim, emacs, helix) and each editor maintains their own query collection. Trying to unify them would likely end up exactly as depicted in xkcd 927. The external repo stores upstream queries as-is and MC applies its own overrides on top. This keeps things simple without introducing yet another standard.

Regarding grammar sources in the external repo, they are already there. The [tree-sitter-grammars](https://github.com/jtyr/tree-sitter-grammars) repo stores `parser.c`, `scanner.c`, and query files under `grammars/<lang>/src/` and `grammars/<lang>/queries/`. The release tarballs contain pre-built `.so` files for convenience but all sources are in the repo. Distros can build from source if they prefer.

Regarding duplicated `tree_sitter` headers, they are copied alongside each grammar's source because that's how upstream grammars ship them. This maintains 100% compatibility per grammar. The headers can change between versions and maintaining separate per-ABI-version directories would be fragile. In any case, this is an internal detail of the external repo and doesn't affect the MC PR.

The following are a few screenshots showing the TS highlighting in action.

### Helm chart template helper (Go TMPL)
{185AF676-969D-4A69-94A7-8CB1DAE7375E}

### Helm chart YAML template (Go TMPL with YAML injection)
{DD1F1841-24C4-413E-9B59-F3E380CBA368}

### Go TMPL with Markdown injection
{1A8C41F1-E8A9-48D5-9330-FD1C23A1CCE9}

### Markdown file with YAML and Bash injection
{DBED2FDE-01FE-4895-B2F5-87610752CC9A}

### YAML file
{82A4DFB5-DC2F-47BF-8045-91693FF8A089}

### Terraform file
{BF5C60FC-B0AD-4166-9C01-BF5E290C6A6F}

### Bash
{D7E826E3-6358-4B32-8252-796F50593E4A}

### Python
{AA1FBAAF-95D3-4EC8-9D85-E3F245C6B109}

### C
{3B495913-13AD-4C09-9B20-97F21595131C}

ossilator:
>  compatibility headers in `tree_sitter/` (needed because system libtree-sitter only installs `api.h`, not the internal headers that grammars `#include`),

why is this required for building mc?
if the answer is "for the static build", then why doesn't the external repo build static libraries?

> MC-specific query overrides that tune colors to match MC's scheme.

what is the objective of this? just to make the highlighting identical to cooledit's? is that a worthwhile goal?

> Regarding TS community discussions, I haven't discussed this with the TS community directly. There are already several query standards (neovim, emacs, helix) and each editor maintains their own query collection. Trying to unify them would likely end up exactly as depicted in xkcd 927. The external repo stores upstream queries as-is and MC applies its own overrides on top. This keeps things simple without introducing yet another standard.

a dependency that is specific to mc is basically just a part of mc, so the whole thing is kinda a cheat if you don't get community buy-in.

the objective is to get a single systemwide set of plugins, preferably along with the upstream queries. if projects want to install overrides for the queries, that's just fine, as these are small compared to the plugins.

> Regarding grammar sources in the external repo, they are already there. The [tree-sitter-grammars](https://github.com/jtyr/tree-sitter-grammars) repo stores `parser.c`, `scanner.c`, and query files under `grammars/<lang>/src/` and `grammars/<lang>/queries/`.

those c files are build artifacts in text form, not sources. the input to the ts cli is the source.

jtyr:
Thank you for your your feedback, @ossilator.

Regarding the compatibility headers, these are only needed for the static build path where grammar C files are compiled directly into the MC binary. For shared builds (the default), grammars are pre-built `.so` files and the headers aren't involved. The question is which build modes you want to support going forward. If shared-only is acceptable, no headers are needed in MC's tree at all. If static builds should remain an option, we could ship pre-built `.a` files from the external repo instead of compiling grammar sources inside MC's build.

Regarding the query overrides, it's not about replicating CoolEdit's output. Upstream queries are designed for GUI editors with rich color themes. In a 16-color terminal, many upstream captures are either missing or too granular to be useful. The overrides address concrete gaps - for example, Ada's upstream query doesn't capture predefined types, GLSL lacks builtin function highlighting, Makefile queries miss automatic variables and shell assignments. These are small `.scm` text files that ship with MC, not in the external repo. MC needs its own overrides to make the highlighting work in its context, just like Neovim and Helix maintain their own query sets.

Regarding the community buy-in, I understand the concern, but this is fundamentally a tree-sitter ecosystem problem. No shared grammar distribution exists today - Neovim, Helix, and Emacs each maintain their own collections. Creating a unified one would be a significant undertaking. The external repo is structured to be editor-agnostic (standard `.so` exports, upstream queries shipped as-is) so it could serve other projects, but driving ecosystem-wide adoption is beyond the scope of what I can take on as an individual contributor. If this is important to the MC project, it might be worth raising with the tree-sitter community from MC's position as an established project with broader reach.

Regarding the grammar sources, good point. I can add the `grammar.js` files to the external repo alongside the generated C files. The tree-sitter CLI could then regenerate `parser.c` and the internal headers from `grammar.js`, so distros that prefer to build from true sources would have that option. The generated `parser.c` files would remain for convenience since the tree-sitter CLI is a Rust tool that not everyone has. Note that `scanner.c` files are hand-written by grammar authors and can't be regenerated, so those still need to be fetched from upstream regardless.

ossilator:
> If static builds should remain an option, we could ship pre-built `.a` files from the external repo instead of compiling grammar sources inside MC's build.

yes, that was my point.

distros would put the .so and .a builds into separate packages, which you should do as well.

> Regarding the query overrides, it's not about replicating CoolEdit's output. Upstream queries are designed for GUI editors with rich color themes. In a 16-color terminal, many upstream captures are either missing or too granular to be useful.

i would expect exactly the opposite. something important must be missing from the argument.

> MC needs its own overrides to make the highlighting work in its context, just like Neovim and Helix maintain their own query sets.

off-hand, i wouldn't expect neovim's requirements to deviate fundamentally from mc's.
to me, the reasonable thing seems to be consolidating things in the various upstreams. why shouldn't this work, except for devs being obstinate?

> Regarding the community buy-in, I understand the concern, but this is fundamentally a tree-sitter ecosystem problem.

yes, and i told you that to get buy-in _from me_, you need to make an effort to solve their ecosystem problem, because nobody else is doing it. i can't speak for the actual mc maintainers, but my impression was that at least yury would agree with my stance.

> driving ecosystem-wide adoption is beyond the scope of what I can take on as an individual contributor.

all it takes is a single guy with stamina, and you seem to be that. make a problem statement, advertise your project as the solution (to both ts and the significant users), and offer to transfer your repo into the ts namespace and maintain it. if they cannot be convinced, you can try to convince the big linux distros instead - i'm fairly sure that they would like the idea.

> If this is important to the MC project, it might be worth raising with the tree-sitter community from MC's position as an established project with broader reach.

assuming the mc maintainers (all two of them) approve the direction, you can act with the whole authority of the mc project.

jtyr:
Update since last discussion:

Static builds now use pre-built `.a` files from the [external grammars repo](https://github.com/jtyr/tree-sitter-grammars). Grammar sources are no longer compiled inside MC's build. The compatibility headers (`tree_sitter/`) have been removed entirely. The external repo produces separate shared and static tarballs per platform.

Regarding query files and upstream consolidation, the grammar `.so`/`.a` libraries are universal. Any project can use them as-is. Query files are not. Each editor maintains its own set because upstream queries are unreliable in practice:

- Some grammars ship incomplete queries with missing captures for common node types
- Some use editor-specific captures (`@spell`, `@conceal`, `@indent`) that have special meaning in one editor but are meaningless in others
- Capture naming conventions vary across grammars. There is no enforced standard
- Naming also depends on the language itself. There is a small set of captures that can be common across editors (e.g. `@code`, `@italic`, `@bold`), but many are inherently language-specific and have no universal equivalent
- A single invalid node name in a query causes `ts_query_new()` to reject the entire file, silently breaking all highlighting for that language

This is why Neovim, Helix, and Emacs each maintain their own query sets. The mapping from AST nodes to visual presentation is inherently editor-specific. Consolidating into a single universal query file would require all editors and all grammar authors to agree on capture conventions, which has not happened despite tree-sitter existing for years.

The external grammars repo covers building and distributing the grammar libraries. Each consuming project creates its own query files tailored to its needs, the same way Neovim and Helix do today.

I have to be upfront that I'm not in a position to drive ecosystem-wide standardisation across the tree-sitter community. That would require time and a level of engagement I'm not able to commit to. What I can offer is what's already there: the grammars repo provides a working, reusable solution that any project can build on. I have been using MC for almost 30 years and built this because I believe it deserves a modern syntax highlighting system on par with what other editors offer today.

m4.include/mc-with-tree-sitter.m4
```diff
@@ -47,73 +47,47 @@ AC_DEFUN([mc_WITH_TREE_SITTER], [
...
-            ts_need_cxx=no
+            dnl Static mode: discover available .a files and build grammar list
+            ts_grammar_dir="$srcdir/src/editor/ts-grammars"
+
+            if test x"$with_tree_sitter_grammars" = xall; then
+                dnl Auto-discover all available .a files
+                tree_sitter_grammars=""
+                for a in "$ts_grammar_dir"/*/*.a; do
+                    test -f "$a" || continue
+                    g=`basename "\`dirname \"$a\"\`"`
```
ossilator: what a quoting nightmare. i suggest you use a temporary for the dir name.

INSTALL
ossilator: huh, i don't think you meant to commit that diff (which certainly came from autoreconf -f).

ossilator:
> Static builds now use pre-built `.a` files [...]

ok, i guess the mc side is now fundamentally sane.

> Consolidating into a single universal query file would require all editors and all grammar authors to agree on capture conventions, which has not happened despite tree-sitter existing for years.

but how many people have even _tried_? it's entirely plausible that upstream just f'd up by neglecting the ecosystem, and everyone approached the issue just like you.

> I have to be upfront that I'm not in a position to drive ecosystem-wide standardisation across the tree-sitter community.
> That would require time and a level of engagement I'm not able to commit to. 

that seems like an unnecessarily pessimistic approach.

as far as i'm concerned, you absolutely _have_ to advertise your repo and propose adoption, because otherwise it's all a bit phony as explained previously. that step doesn't require much effort from your side.

once upstream and one or two downstream projects "bite", they might also get interested in standardizing the query files. that's not something you have to be involved in if you are successful in having your repo adopted/transferred upstream. it's also a lot less pressing than consolidating the grammars, so it's not _that_ bad if it doesn't happen.

@zyv and @mc-worker, please state you opinions.

jtyr:
After careful analysis of the tree-sitter ecosystem and the feedback from the review, I've redesigned the grammar distribution model.

The tree-sitter ecosystem has no standardisation for query files. This is by design. Tree-sitter provides the parser and the AST, the editor provides the interpretation. Query files bridge the two by mapping AST node patterns to editor-specific capture names and predicates.

While node names are defined by the grammar and are universal, capture names carry editor-specific semantics. Neovim uses `@spell` to trigger spell checking and `@conceal` to hide text. Helix uses `@combined` to merge injection ranges. Zed uses `@font-style` to change font rendering. MC uses captures like `@delimiter.special` to map to terminal colors. These mean nothing to each other's rendering engines.

Predicates are similarly editor-specific. Neovim adds `#lua-match?` and `#vim-match?` for Lua/Vim regex matching. Helix adds `#set! injection.combined` for injection control. Editors that don't implement a predicate will silently ignore it or skip the pattern entirely, producing wrong highlighting.

Maintaining a single universal query file across editors would mean carrying dead weight from every editor's custom captures and predicates, with each editor ignoring what it doesn't understand and risking subtle breakage when another editor adds or changes something. This is why every tree-sitter editor - Neovim, Helix, Zed, Emacs - maintains its own set of query files. MC does the same, in a separate community repo.

This also means query files and grammar libraries must be validated together. A grammar update can change node names, silently breaking any query file that references the old names. With the new model, CI validates every query file against its grammar on every PR and every release. If a grammar update changes node names, the PR check fails until a contributor fixes the query file. There is no way for broken queries to reach a release.

MC now follows the same model as every other tree-sitter editor:

- **MC ships only the TS integration code** - no grammars, no query files, no grammar configuration.
- **A separate community repo** ([mc-ts-grammars](https://github.com/jtyr/mc-ts-grammars)) provides everything else: MC-curated query files, per-grammar configuration, grammar sources (as backup), and pre-built release bundles.
- **An installer tool** (`mc-ts-grammar`) written in Bash handles downloading, building, installing, updating, listing available grammars, and removing grammars. Works on all platforms.

This is how Neovim (`nvim-treesitter`), Helix, Zed, and Emacs all handle grammar distribution. It's the established pattern because the ecosystem simply doesn't support a one-size-fits-all approach.

The MC codebase contains:

- `syntax_ts.c` / `syntax_ts.h` - the TS highlighting engine
- `ts-grammar-loader.h` - runtime .so loading via `g_module_open()`
- `mc-with-tree-sitter.m4` - the `--with-tree-sitter` configure flag
- `mc-syntax-dump` - a testing tool for comparing TS vs Legacy output

That's it. No grammar data, no query files, no config files. The static build mode has been dropped. It added significant complexity for no real benefit since every platform MC runs on supports shared libraries via GLib's `g_module_open()`.

A new persistent setting "Tree-sitter highlighting" is available in *Options > General*. It defaults to enabled when compiled with `--with-tree-sitter`, but if no grammars are installed, the legacy engine handles everything transparently. Users who never install grammars see no difference from a non-TS build.

When TS is enabled:

- Status bar shows `S:[TS]`, `S:[Legacy]`, or `S:[None]`
- `Ctrl+S` cycles through all three modes
- `Ctrl+T` toggles between TS and Legacy for quick comparison

When TS is disabled (or not compiled in):

- Status bar looks exactly like the original MC
- `Ctrl+S` cycles Legacy/None as before
- `--no-tree-sitter` CLI flag still works as a runtime override

The grammars repo [mc-ts-grammars](https://github.com/jtyr/mc-ts-grammars) provides:

- **68 languages** with MC-curated query files and color configuration, each validated against its grammar in CI.
- **358 grammar sources** tracked from upstream as backup (community can contribute queries for any of them).
- **Per-grammar self-contained directories**: `config.ini` (INI format with metadata + colors), `highlights.scm`, optional `injections.scm`.
- **Pre-built release bundles** for 5 platforms (x86_64/aarch64 Linux, x86_64/aarch64 macOS, x86_64 Windows).
- **CI pipeline**: PR validation (query files compiled against built grammars), weekly upstream grammar sync, weekly releases.
- **Comparison reports** documenting TS vs Legacy highlighting quality for each language.

The per-grammar directory structure makes distro packaging straightforward. Each grammar is independent and always ships with a validated query file. Distros never have to worry about grammar/query version mismatches:

```text
Package: mc-ts-grammar-python
Files:
  /usr/lib/mc/ts-grammars/python.so
  /usr/share/mc/syntax-ts/python/config.ini
  /usr/share/mc/syntax-ts/python/highlights.scm
```

A meta-package can depend on all grammar packages, or users can install individual grammars via the installer:

```bash
# See what's available in the latest release
mc-ts-grammar available

# Install specific grammars
mc-ts-grammar install python bash yaml markdown

# Install everything
mc-ts-grammar install --all
```

MC maintainers need to maintain:

- The TS integration code in `syntax_ts.c` (~1500 lines) - this is stable, well-tested, and rarely needs changes.
- The `mc-syntax-dump` tool for testing.
- The `--with-tree-sitter` configure flag.

Grammar updates, new language support, query fixes, and community contributions all happen in the external repo with no impact on MC's release cycle.

ossilator:
> Maintaining a single universal query file across editors would mean carrying dead weight from every editor's custom captures and predicates,

this is correct, but correctly implemented the runtime cost of this would be negligible.

> with each editor ignoring what it doesn't understand and risking subtle breakage when another editor adds or changes something.

this wouldn't be a problem if everything was covered by autotests.

anyway, i don't really care, as the queries are somewhat tiny.

i do however care about the grammars. the status quo is that having five editors that use ts installed implies half a gigabyte of duplicated binaries. that's insanity.

i therefore consider it _mandatory_ that an upstream-approved grammar aggregator repository exists. downstreams can pin their queries to particular versions, though obviously they all need to pin to the latest release within a short time for the de-duplicating effect to set in.

if "the ecosystem" isn't interested, then i'm not interested in that ecosystem. it's really as easy as that.

jtyr:
First, I genuinely appreciate the feedback throughout this review. The pushback on the original approach shaped the current design into something much cleaner, and the move to a separate grammars repo with prebuilt tarballs came directly out of this discussion.

On the aggregator point, the reason the ecosystem looks the way it does is that query files and grammar binaries are tightly coupled by version. A grammar update can rename AST nodes, and `ts_query_new()` rejects the entire query file on a single unknown node, silently disabling all highlighting for that language. This forces each editor to pin grammar versions to match its query files. A single shared binary bundle would require every consumer to lockstep on grammar versions, which is exactly what editors cannot do in practice. Neovim, Helix, Zed, and Emacs each ship their own binaries for this reason, not out of laziness or lack of interest in cooperation.

The orthogonal approach, where each grammar author publishes `.so` and `.a` artifacts from their own repo, would solve deduplication, but it introduces an orphan risk. If a grammar repo disappears or stops publishing, every downstream editor loses highlighting for that language. A curated mirror is still needed.

At the system level, Linux distro packaging (`libts-grammar-python.so` shared across `mc-ts`, `helix`, `nvim-treesitter` packages) is the established way to deduplicate without forcing version alignment on editors. That is where this concern is properly solved.

I am happy to keep contributing on the MC side, but driving standardisation across the tree-sitter community is outside what I can commit to, and it is a separate conversation from this PR.

Looking at the broader picture, the change is fully gated behind `--with-tree-sitter`, isolated from the rest of the codebase, and optional at runtime via a persistent setting. Users who never compile it in or never install grammars see identical behaviour to today's MC. Users who do install grammars get a highlighting engine that solves problems the legacy regex engine cannot, and that many have been asking for over the years.

Could I ask the maintainer group (@zyv, @mc-worker, and others) to discuss this internally and reach an official decision on whether MC is interested in the TS integration as a whole? If the answer is yes, I can address remaining review comments under a clear direction. If it is no, I would rather know so I can stop consuming reviewer time on a PR that will not be merged.

ossilator:
> Linux distro packaging [...] is the established way to deduplicate without forcing version alignment on editors.

it's _one_ way to solve it, but this just forces the version alignment issue downstream. that makes things worse overall, because packagers often have even less time, are often less qualified, and resources are spread over several distros.

i think a plausible approach would be merging the query repositories even if the queries themselves are not merged. such centralization would tremendously simplify the logistics of keeping things in sync. the person wanting to do a grammar upgrade for "their" project would also have to pay the (rather small) price of updating all "foreign" queries, or at least to get everyone else to amend the PR. shared review and universal CI would ensure that nothing goes wrong.

> the change is fully gated behind --with-tree-sitter

this argument has way less pull than you may think.

jtyr:
Thanks @ossilator, this is useful context. Merging query repositories across editors is a strictly larger coordination problem than a shared binary aggregator, since every grammar upgrade would need synchronised PRs across every editor's queries with shared governance that does not exist today. That work is well outside the scope of a single MC PR and it is not something I can take on.

On the `--with-tree-sitter` point, if an opt-in flag that reuses the existing highlighting pipeline and ships zero grammar data by default is considered too much, then the real concern is not about logistics or dedup. I would genuinely like to know what it is, so I can either address it or step back.

Could @zyv, @mc-worker, and other maintainers weigh in on whether MC wants this integration in principle? I am happy to keep iterating if there is a path forward, and equally happy to stop if there isn't.

zyv:
> Could @zyv, @mc-worker, and other maintainers weigh in on whether MC wants this integration in principle? I am happy to keep iterating if there is a path forward, and equally happy to stop if there isn't.

Sorry, I had to (and will have to, at least until the end of May) concentrate on my exams, and Andrew is having internet access issues. I'm not sure he has interest and/or opinions on the design rather than the implementation details. I started trying to get a picture for myself of the design and ecosystem, and writing a long response to clarify the points which I think are important the following weekend. I wasn't able to finish it though and had to put it off. Maybe I'd find a chunk of time earlier, but I'm not sure. In the meantime, a lot of discussion happened here, which I couldn't follow yet.

To answer your question quoted above compactly: yes, of course, the integration is wanted **in principle**. However, I have many concerns from design and maintainability points of view. In as far as implementation is concerned, these are details in my opinion, and you can certainly fix them to mutual satisfaction. It's the understanding and clearing up of these concerns that needs my time, which at this moment I wasn't able to make, and it is made worse by the marketing-biased AI wall of text stuff.

The most important points for me (this has already been partly addressed implicitly):

1. Replace existing highlighting engine and unify highlighting in editor / viewer
2. Don't lose too much / break too many important use cases by doing so
    * This includes language coverage, which would somewhat degrade in terms of quantity and ease of adding new stuff, but dramatically improve in quality for the important stuff
    * But also systems coverage, which now I'm getting more concerned about (like viability on embedded in terms of space and speed)
3. Have somebody else / community properly maintain the TS-based grammar horror in the future, since upstream apparently is not doing that
4. Have a reasonable packaging / building / distribution model (apparently here a lot of good progress has been made since the PR was first submitted)

I'm not sure how non-negotiable (1) is yet for me. I'd like to say it isn't negotiable, but maybe it's simply impossible given the limitations of TS, and if this is the only ground to reject it, it would harm mc as a project in an unjustified way. I can't tell. I simply need time to understand and make up my mind.

I would certainly feel a lot better about (1) if there is some way to dramatically reduce our own surface, if not completely remove the old highlighter. Maybe our definitions can be somehow "easily" generated from TS-grammars and our code rewritten in a very clean / compact way, and covered with tests. Isn't it the promise of all those Claude-makers that stuff like that will now happen in a snap of the fingers? :)

In this case, the justification for dropping (1) is that we have an extremely simple and "bad" internal system, but with low maintenance requirements, light, fast and sloppily extendable for user "grammars", and available on embedded, or as a fallback, or in resource-constrained environments of some sort, or for anyone not wanting TS, and a proper editor-grade system for everybody else as a default. 

Regarding other points, I think you've provided a lot of information already and even a working prototype, so before requesting something else, it should be studied and tested first.

I hope this helps in some way.

ossilator:
> every grammar upgrade would need synchronised PRs across every editor's queries

no, it wouldn't, as i already explained.

> with shared governance that does not exist today. 

indeed, and that's kinda the crux.

>  if an opt-in flag that reuses the existing highlighting pipeline and ships zero grammar data by default is considered too much, then the real concern is not about logistics or dedup.

no, that's an _additional_ real concern.
if you drop an opt-in solution that no-one can or wants to use (by which i also mean distro packaging) because the logistics suck, then that's basically just dead code in mc. code that causes an additional maintenance burden, because changes to the living code now need to consider the impact on the "undead" code. "keep it in your own fork" would be a reasonable response in that case.

jtyr:
Thank you for your response, @zyv. No pressure on the timeline, take the time you need for exams and the longer response.

On language coverage, TS currently ships 68 grammars versus 104 legacy syntax files. Filling the gap for the simpler cases is easy, the MC-specific config files in particular would take very little effort. For now the two engines would coexist and legacy handles anything TS does not cover. Once TS coverage reaches parity, legacy could be dropped.

The `.syntax` generation idea is interesting but I am not sure how feasible it is in practice. Keywords are baked into the grammar source, not the query files, so our queries mostly add capture groups on top. Extracting keyword lists would mean parsing grammar.js, which is variable in shape across projects. Might be worth a prototype for a few grammars to see how far it goes.

On embedded and resource-constrained environments, your concern is legitimate. I can profile memory and CPU with and without TS on a real file to give concrete numbers. The opt-in flag already keeps the TS code path out of non-TS builds, and legacy staying as the fallback for those systems feels like the right outcome either way.

Looking forward to the longer response whenever you have time.

PR #5068, opened 2026-03-15
panel: make Up/Down cycle through matches during quick search

karpovilia:
## Proposed changes

Add two new panel actions `SearchNext` and `SearchPrev` that cycle through  files matching the current quick search pattern, with wrap-around. These actions are only active while quick search is running.

By default, `SearchNext` is bound to `Ctrl+S` (same key that starts/continues  the search), and `SearchPrev` has no default binding. Both can be customized  in the `[Panel]` section of the keymap file.

This follows the existing mc convention of using symbolic action names bound through keymap files, rather than hardcoding specific keys.

## Checklist

### DIFF
--- a/src/filemanager/panel.c
+++ b/src/filemanager/panel.c
@@ -3530,11 +3530,81 @@ panel_execute_cmd (WPanel *panel, long command)
 {
     int res = MSG_HANDLED;
 
-    if (command != CK_Search)
+    if (command != CK_Search && command != CK_SearchNext && command != CK_SearchPrev)
         stop_search (panel);
 
     switch (command)
     {
+    case CK_SearchNext:
+    case CK_SearchPrev:
+        if (panel->quick_search.active)
+        {
+            int direction = (command == CK_SearchNext) ? 1 : -1;
+            int start = panel->current + direction;
+            int i, found = -1;
+            gboolean wrapped = FALSE;
+            char *reg_exp, *esc_str;
+            mc_search_t *search;
+
+            if (panel->quick_search.buffer->len == 0)
+                break;
+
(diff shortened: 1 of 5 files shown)


mc-worker:
BTW, in the active quick search mode, pressing of `ctrl-s` or `alt-s` moves you to the next match.

Please remove NEWS file from your PRs. This file is created from the wiki page in the release process.

src/filemanager/panel.c
```diff
@@ -3530,6 +3530,79 @@ panel_execute_cmd (WPanel *panel, long command)
 {
     int res = MSG_HANDLED;
 
+    /* When quick search is active, Up/Down cycle through matches instead of exiting search */
+    if (panel->quick_search.active && (command == CK_Up || command == CK_Down))
```
egmontkob: I believe a better thing to do would be to add two new symbolic names e.g. `CK_Prev/Next_Match` or so, and bind them in `mc.*.keymap`. Maybe some users want to map it to different keys.

egmontkob (commented):
> When quick search is NOT active, Space marks the current file
> and Backspace navigates to the parent directory.

You make multiple unrelated changes. They should preferably go to separate commits.

I truly dislike this particular change, I expect Space to act like a Space (see also freshly filed #5072). We already have two different shortcuts (Insert and Ctrl+T) for toggling the file, we don't need a third one.

Okay, I misunderstood your change, I thought you'd only do this if the command line is empty. You always do it, so I literally cannot type a command line containing a space, nor fix a mistake I made while typing. This is outright unacceptable.

Even if only at empty command line, I'd be against this feature. I want to be able to start a command with a space. And I want to be able to remove a command line by long-pressing Backspace where of course I won't be able to release it at the very exact moment it's erased all the letters.

Issue #3254, opened 2014-08-22
Shift+Fn keys are off by two in ncurses


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3254 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `egmont` (@egmontkob) |
> | **Mentions** | `mrmazda@….net` |

<p>
In the slang build, Shift+F1 - Shift+F10 work as expected.  (Also, in the editor Shift+F1 can be used instead of F11 to bring up the User menu, and Shift+F2 instead of F12 for Save as.)<br />
</p>
<p>
In the ncurses build they are off by two, e.g. Shift+F6 wants to delete the file under the cursor.<br />
</p>
<p>
I see this behavior on Ubuntu Trusty, TERM=xterm-256color, with xterm, gnome-terminal and st (all manually upgraded to their newest version, they emit Shift+F1=\e[1;2P, Shift+F5=\e[15;2~ etc.), konsole (Shift+F1=\eO2P, Shift+F5=\e[15;2~).<br />
</p>
<p>
The bug does not occur on pterm (putty) which generates Shift+F1=\e[23~, Shift+F5=\e[28~ etc.<br />
</p>

egmontkob:
I'm working on this, really want to fix this for 4.9.

mc uses an off-by-10 notation, and even uses `KEY_F` with off-by-10 semantics, e.g. `KEY_F(13)` for Shift+F3, even though `KEY_F` is ncurses's macro specifically meant to be used as off-by-12. No wonder it goes wrong somewhere.

egmontkob:
As for the Linux console, I've filed an update at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=698885.

I've found @zyv's `4632_shift_fx` branch which also points to this direction, so I'll probably pick that into my work.

zyv:
My kernel patches were met with resounding silence after several attempts to move forward:

https://lore.kernel.org/linux-serial/BD85D5D1-4D09-41AC-BFD8-C53E986C629F@shurup.com/

Thus, I've gotten demotivated and given up on this. Please feel free to pick up the kernel part - I really don't understand why I was being ignored... And, of course, salvage anything you can from my branch.

ossilator:
[...]

PR #5073, opened 2026-03-18
Ticket #4512: load keys from terminfo and termcap

### LINKED ISSUE #4512: mc should use kPRV5 from terminfo for Ctrl+PageUp
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4512 |
> | **Reporter** | `CyberShadow` (thecybershadow@….com) |

<p>
I noticed that the Ctrl+<tt>PageUp</tt> combination (to navigate to the parent directory) didn't work in wezterm (with <tt>config.term = "wezterm"</tt>). A bit of experimentation showed that it worked when wezterm identified itself as xterm, or if mc is started with --xterm. Upon investigating, I found that the reason why it works with TERM=xterm is only because mc itself hard-codes the sequence that xterm uses (the xterm_key_defines array), when it detects that it is running in that terminal emulator (check in init_key) or with --xterm.<br />
</p>
<p>
I am not sure if this is standardized anywhere, but I noticed that many terminal emulators declare the sequence for Ctrl+<tt>PageUp</tt> as the kPRV5 capability in their terminfo. Grepping for kPRV5 in my /usr/share/terminfo gives me over 60 matches. wezterm's terminfo doesn't have it yet, but it would not be a problem to add it.<br />
</p>
<p>
I think it would be nice if mc checked terminfo for capability strings such as kPRV5, in order to support such key combinations in other terminals.<br />
</p>
<p>
I can work on a patch with some guidance. I noticed the Key_xxx definitions in keyxdef.c, but it's not clear to me how to extend that to only teach mc about some new terminfo keys.<br />
</p>

egmontkob:
## Proposed changes

Load kPRV5 and friends from terminfo and termcap

* Resolves: #4512

## Checklist

- [ ] I have signed-off my contribution with `git commit --amend -s`

---

This is a modification of @mc-worker's `4512_terminfo` branch in the main repo:
- updated against current master
- addressed my five earlier comments at https://github.com/MidnightCommander/mc/commit/e490f856d08199dc958982b838f392baedbd87d0 (I have to click on the "Collapse file" arrow in front of the filename, at the top-left corner of the bottom-right panel, and click again to Expand, in order to see those comments – github bug??)
- most importantly: the change didn't actually do anything; it looked up the sequences, stored in the array, and never looked at them again. Now it passes them to `define_sequence()` (and doesn't store them in the array, it's not needed).

The change currently only affects slang, it's a no-op with ncurses, but I'm about to change that in #3254 / #4632.

### DIFF
--- a/lib/tty/key.c
+++ b/lib/tty/key.c
@@ -253,6 +253,14 @@ typedef struct
     int action;
 } key_define_t;
 
+typedef struct
+{
+    int code;
+    const char *terminfo_name;
+    const char *termcap_name;
+    int action;
+} term_key_define_t;
+
 /* File descriptor monitoring add/remove routines */
 typedef struct
 {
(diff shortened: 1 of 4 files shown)


egmontkob:
_Of course_ I forget to check with slang... _of course_ it doesn't compile :) hang on...

egmontkob:
@mc-worker 

I'm really not sure what approach you had in mind for Slang. (It's already broken in your branch.)

I do like the overall change, we'll definitely need to have something like this.

Would you prefer to update this branch and fix for Slang, in the near future; or would you prefer me completing it (not necessarily exactly the way you imagined it)?

mc-worker:
> would you prefer me completing it (not necessarily exactly the way you imagined it)?

Yes, please. I believe you make it better than me.

PR #5080, opened 2026-03-25
vfs/shell: add MC_SSH_COMMAND and MC_SSH environment variable support

### LINKED ISSUE #2256: shell link via ssh fails to restore after connection lost
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/2256 |
> | **Reporter** | `crissi` (crissi99@….de) |
> | **Mentions** | `zaytsev` (@zyv), `gotar@….pl` |

<p>
I use a remote dir using "Shell link" (via ssh). If the connection was resetted by peer (e.g. adsl disconnect and reconnect with new ip address) changing the remote director fails instead of reconnecting. mc knows my password already, so it could reconnect and shoudl do that. Selecting "Rescan" fails also :(<br />
Reconnect via "Shell link" also fails. Only way is to close mc and start again from new.<br />
</p>

melnikovsky:
Add support for `MC_SSH_COMMAND` and `MC_SSH` environment variables (analogous to `GIT_SSH` and `GIT_SSH_COMMAND`). Enables custom wrapper usage, particularly [autossh](https://www.harding.motd.ca/autossh/)

* Example:
  `MC_SSH_COMMAND='autossh -M0 $@' mc`

* Resolves: #2256

### DIFF
--- a/src/vfs/shell/shell.c
+++ b/src/vfs/shell/shell.c
@@ -515,11 +515,38 @@ static void
 shell_open_archive_pipeopen (struct vfs_s_super *super)
 {
     char gbuf[10];
-    const char *argv[10];  // All of 10 is used now
+    const char *argv[14]; // increased for optional sh -c prefix
     const char *xsh = (super->path_element->port == SHELL_FLAG_RSH ? "rsh" : "ssh");
+    const char *mc_ssh_cmd;
+    gboolean use_shell_cmd;
     int i = 0;
 
-    argv[i++] = xsh;
+    mc_ssh_cmd = g_getenv ("MC_SSH_COMMAND");
+    use_shell_cmd =
+        (mc_ssh_cmd != NULL && *mc_ssh_cmd != '\0' && super->path_element->port != SHELL_FLAG_RSH);
+
+    /* MC_SSH: plain executable override, lower priority than MC_SSH_COMMAND.
+     * Ignored for rsh connections, mirroring the GIT_SSH convention. */
+    if (!use_shell_cmd && super->path_element->port != SHELL_FLAG_RSH)
+    {
+        const char *mc_ssh = g_getenv ("MC_SSH");
+
+        if (mc_ssh != NULL && *mc_ssh != '\0')
+            xsh = mc_ssh;
+    }
+
+    if (use_shell_cmd)
+    {
+        /* Invoke as: sh -c "$MC_SSH_COMMAND" ssh [MC-generated-args...]
+         * MC's flags (-C, -p, -l, host, remote-cmd) arrive as $@ inside the
+         * shell command, mirroring the GIT_SSH_COMMAND convention. */
+        argv[i++] = "sh";
+        argv[i++] = "-c";
+        argv[i++] = mc_ssh_cmd;
+        argv[i++] = "ssh"; // $0 inside the shell script
+    }
+    else
+        argv[i++] = xsh;
     if (super->path_element->port == SHELL_FLAG_COMPRESSED)
         argv[i++] = "-C";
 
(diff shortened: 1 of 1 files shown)


zyv:
I don't quite see how it properly fixes #2256:

My understanding is that the pipe to `ssh` command closes and we don't reopen it properly. Therefore, I thought that all that is needed to get it fixed is to try to re-open the pipe if it's closed. If `autossh` closes the connection (for whatever reason), then the same will happen with it as well.

Why try to solve this problem by introducing support for an external `ssh` command? Is it not possible to just attempt to re-open the pipe? Are there any other advantages that an external command brings?

melnikovsky:
Admittedly, this PR does not _fix_ #2256 but kind of _resolves_ it. A proper fix would require going deeper in the `mc` codebase, which I tried to avoid.

Instead of implementing reopen (how many attempts?) and proper cleanup afterward - which is arguably even more important, since currently one has to restart `mc` to reconnect - I suggest offloading this responsibility to an external tool. `autossh` already does a decent job of re-establishing connection (up to `AUTOSSH_MAXSTART` times) and keeping the pipe alive. Works for me...

Speaking of "other advantages", this approach could potentially allow unifying `ssh/rsh` code paths. That is assuming `rsh` is still relevant nowadays (I think the last time I used it was ~20 years ago)...

zyv:
Alright, thanks for the explanation. I understand now why you submitted this PR, but hopefully you can also see the problem I have with it as a maintainer.

I can take it in to introduce a new (so far undocumented) way to swap out the default ssh command. Without documentation, it will be undiscoverable. With documentation, less so, but nevertheless, it will have to be discovered and configured by every user who wishes to have it. On top of that, you'd have to have an external tool (autossh) installed and working. All in all, terrible user experience for some new code pathways and extra documentation.

At the same time, one could just put the following in the `~.ssh/config` file to the same effect:

```
Host *
  ServerAliveInterval 60
```

This is how I solve the problem for me at the moment.

Also, I think a minimal halfway proper implementation is rather simple. All you have to do is to check if the pipe is alive in the VFS command thing, and if not, close VFS with an error. That's it. The user will be able to reconnect manually afterwards without having to restart mc. If you want to make it nicer, then you can try to re-open the pipe at this point once, and if it doesn't work, then see above.

I think this will be perfect in terms of user experience, will work out of the box for everyone without any external tools, and should not be that hard to implement. I wish you could look into that as an alternative to your current solution.

melnikovsky:
> At the same time, one could just put the following in the `~.ssh/config` file to the same effect:
>
> ```
> Host *
>   ServerAliveInterval 60
> ```
> 
> This is how I solve the problem for me at the moment. 
> ... All you have to do is to check if the pipe is alive in the VFS command thing, and if not, close VFS with an error.

Probably I am missing something important. My `ssh` setup *does* have finite `ServerAliveInterval` and the connection *does* die after a while. But I can *not* reconnect to the same host without restarting `mc`. How should I?

I would not bother fixing anything if I were able to simply reconnect!

zyv:
> Probably I am missing something important. My `ssh` setup _does_ have finite `ServerAliveInterval` and the connection _does_ die after a while.

Well, it doesn't for me. It might die for you though. Why not? Consider genuine connection loss after `ServerAliveInterval` is exceeded, because exactly at this moment you happened to have a hiccup with your internet connection.

> But I can _not_ reconnect to the same host without restarting `mc`. How should I?

You can't. That's a bug. It seems that you have misinterpreted my statement. What I was saying is that fixing this bug so that you *are* able to reconnect (or even try to reconnect once automatically) is not too hard, instead of adding a workaround of questionable usability.

melnikovsky:
> Without documentation, it will be undiscoverable... terrible user experience for some new code pathways and extra documentation.

As an `mc` user myself (since nineties) I never relied on documentation. I just browsed through available menus. Admittedly, the canonical way to configure `ssh` command would be to add another menu item rather than set an environment variable.

> On top of that, you'd have to have an external tool (autossh) installed and working.

It looks like *two* separate questions are being discussed at once:

1. The bug: `mc` should not fall apart when the pipe closes.
2. The feature: reconnect automatically. I am certain `mc` should not try to reconnect. Instead it should delegate this task to an external tool. Be it `autossh` or a shell wrapper.

> What I was saying is that fixing this bug ... is not too hard

OK, I got your point now. Maybe I will look deeper in the code and fix the bug one day... when I have time. Or somebody else will.

PR #5083, opened 2026-03-31
Ticket #5082: add file cloning support for FreeBSD, Solaris and macOS

### LINKED ISSUE #4991: File move on ZFS TrueNAS

yes, moving files with mc is slow it is executing copy-delete instead of rename

mc correctly detects that both source and destination are in the same ZFS pool and executes mv instead of copy-delete

_No response_

### Additional context

_No response_

### LINKED ISSUE #5082: File clone VFS operation support on FreeBSD

On FreeBSD with ZFS, Midnight Commander does not support file cloning.

A problem report that is probably related is #4991, if the author had the original FreeBSD-based TrueNAS in mind, and not the Linux-based TrueNAS Scale.

It would be nice, if on FreeBSD, file cloning were supported on applicable ZFS datasets.

An alternative is using `cp` and `mv` for zero-copy file transfer operations.

### Additional context

_No response_

tuffnatty:
## Proposed changes

Use [`copy_file_range(2)`](https://man.freebsd.org/cgi/man.cgi?copy_file_range) on FreeBSD for file cloning.

It should resolve #5082 and probably #4991. `copy_file_range(2)` also supports remote NFS operations, but it's out of scope for this change.

* Resolves: #5082
* Resolves: #4991

While we're at it, also add support for [`clonefile(2)`](https://www.unix.com/man_page/mojave/2/clonefile/) on macOS and [`reflink(3C)`](https://docs.oracle.com/cd/E88353_01/html/E37843/reflink-3c.html) on Oracle Solaris.

The patch also adds support for cloning in Append/Reget modes. This is working only in Linux and FreeBSD and only if the filesystem allows it (e.g. the start offsets must be block-aligned).

As file cloning is not always a desired thing, a new configuration option is introduced to disable it.

## Checklist

~~I have not added any tests for the feature, as it would involve actions fairly complicated from the ops point of view (like, create a test ZFS filesystem), and the test infrastructure does not offer any support for superuser/VM operations. But~~ it's working locally, copy/move within one ZFS pool became instant, and the existing tests pass.


zyv:
Wow, is it possible to test it in some way? As you can see, we now have CI on FreeBSD. Maybe some way to check that `copy_file_range` indeed gets called?

tuffnatty:
I've added a very dumb test that seems to work on Linux as well as on FreeBSD.

zyv:
> I've added a very dumb test that seems to work on Linux as well as on FreeBSD.

To me that looks fantastic, but can't you check that `ENOTSUP` is returned on other platforms? We have also CI that runs on Solaris and macOS. This way, all possibilities will be covered. Thank you!

tuffnatty:
Hi. It took a bit of time... The new pull request adds macOS and Oracle Solaris support as well. On these platforms, the syscalls for file cloning respond for destination file creation, that is, they take in a path instead of a file descriptor, so the logic around the copy operation had to be adapted.
So the test tests if both kinds of mocked syscalls are called with proper arguments, and that on unsupported platforms, ENOTSUP is returned.
Also, the patch introduces a new user configuration option to enable or disable COW file cloning, as it is not always a desired kind of operation. This part of the pull request depends on #5087 to be merged first.

lib/vfs/vfs.c
```diff
@@ -53,6 +53,12 @@
 #ifdef HAVE_SYS_IOCTL_H
 #include <sys/ioctl.h>
 #endif
+#elif defined(__FreeBSD__) && defined(HAVE_COPY_FILE_RANGE)
+#include <unistd.h>  // COPY_FILE_RANGE_CLONE
+#elif defined(__APPLE__) && defined(HAVE_SYS_CLONEFILE_H)
+#include <sys/clonefile.h>  // CLONE_NOOWNERCOPY
+#elif defined(__sun) && defined(HAVE_REFLINK)
+#include <unistd.h>  // reflink()
```
zyv: Is there any reason not to do it this way since you are now checking at the configure stage anyways? I think that this would be a bit safer in terms of possible drift.

```suggestion
#elif defined(HAVE_COPY_FILE_RANGE)
#include <unistd.h>  // COPY_FILE_RANGE_CLONE
#elif defined(HAVE_SYS_CLONEFILE_H)
#include <sys/clonefile.h>  // CLONE_NOOWNERCOPY
#elif defined(HAVE_REFLINK)
#include <unistd.h>  // reflink()
```

lib/global.c
```diff
@@ -109,6 +109,7 @@ mc_global_t mc_global =
     {
         .cd_symlinks = TRUE,
         .preallocate_space = FALSE,
+        .file_cloning = FALSE
```
zyv: Do you think that there is any downside to enabling it by default, especially if we provide a GUI option to turn it on and off?

I wonder how the other file managers handle this.

lib/global.h
```diff
@@ -220,6 +220,8 @@ typedef struct
         // Preallocate space before file copying
         gboolean preallocate_space;
 
+        // Use COW file cloning on supported filesystems;
```
zyv:
```suggestion
        // Use COW file cloning on supported filesystems
```

tests/lib/vfs/vfs_clone_file.c
```diff
@@ -0,0 +1,245 @@
...
+#ifdef __linux__
+#include <linux/fs.h>   // FICLONE
+#include <sys/ioctl.h>  // ioctl()
+#elif defined(__FreeBSD__) && defined(HAVE_COPY_FILE_RANGE)
+#include <unistd.h>  // copy_file_range()
+#elif defined(__sun) && defined(HAVE_REFLINK)
+#include <unistd.h>  // reflink()
+#elif defined(__APPLE__) && defined(HAVE_SYS_CLONEFILE_H)
+#include <sys/clonefile.h>  // clonefile()
+#endif
```
zyv: I guess same here...

zyv (commented):
Wow, this really looks fantastic. I'm much impressed. I wonder what the others think. I have but just a few minor comments.

lib/global.c
tuffnatty: Well, I believe that setting it off by default would lead to less astonishment, but it's just my opinion.

As for the other file managers - Linux, macOS, and Solaris `cp`, as well as Far2l, guard it with an option and do plain copy by default, while FreeBSD `cp`, Gnome Files (and other `g_file_copy`-based file managers) and KDE Dolphin leave no choice and just clone on platforms where they support cloning.

lib/vfs/vfs.c
tuffnatty: Probably you're right from Autoconf POV. However, there's an already existing `#ifdef __linux__` above. Also, the identifiers are a little bit too generic - for example, Linux has `copy_file_range()`, but we want to use `FICLONE` on Linux, so the additional check brings a little more context although not strictly needed at the moment.
While we're at it, it would be convenient to add something like:
```c
#if defined(FICLONE) || defined(HAVE_COPY_FILE_RANGE)
#define HAVE_FILE_CLONING_BY_FD
#elif defined(HAVE_SYS_CLONEFILE_H) || defined(HAVE_REFLINK)
#define HAVE_FILE_CLONING_BY_PATH
#endif
```
And use these in several other places instead of the uglier disjunctions. But I don't have an idea where to put it (configure.ac? global.h?) and how to name the #defines (do these violate the autoconf namespace?)

tuffnatty:
I've added support for file cloning in append/reget modes. It's working on FreeBSD and Linux when the file offsets are block-aligned. I've also removed the extra #ifs by moving the corresponding checks to autoconf.

tuffnatty:
The linter fails because I have touched a file with pre-existing invalid formatting.

tests/lib/vfs/vfs_clone_file.c
tuffnatty: Done

lib/global.c
zyv: @tuffnatty any comment on that?

zyv:
> I've added support for file cloning in append/reget modes. It's working on FreeBSD and Linux when the file offsets are block-aligned. I've also removed the extra #ifs by moving the corresponding checks to autoconf.

This looks good to me! I'm just left wondering whether this is a good idea to enable it by default.

Are you planning to do anything else in this PR, or are you considering it ready for review now?

tuffnatty:
I had to find and click the Submit review button as my previous review comments were in pending state.

I don't think I have anything to add or remove at this moment.

lib/global.c
zyv: Alright, would you mind adding a change to enable it by default then? We can always opt to disable it by default if there are massive complaints.

I'm just afraid of adding options like that, because I think it will lead to zero usage.

lib/vfs/vfs.c
zyv: You could do it in autoconf, right? Something like this:

```m4
if test "x$ac_cv_header_sys_clonefile_h" = xyes || \
   test "x$ac_cv_func_reflink" = xyes; then
    AC_DEFINE([HAVE_FILE_CLONING_BY_PATH], [1], [Define if system can clone files by path])
fi
```

lib/global.c
tuffnatty: Done.

zyv (approved):
I've fixed the formatting error by adding a few hanging commas (in the same commit). Inasmuch as I'm concerned, this looks ready for merge. Thank you for your high-quality contribution! 👍

tuffnatty:
Thanks @zyv!

src/filemanager/boxes.c
```diff
@@ -569,6 +569,7 @@ configure_box (void)
                     QUICK_CHECKBOX (_ ("Mkdi&r autoname"), &auto_fill_mkdir_name, NULL),
                     QUICK_CHECKBOX (_ ("&Preallocate space"), &mc_global.vfs.preallocate_space,
                                     NULL),
+                    QUICK_CHECKBOX (_ ("Use COW file cloning"), &mc_global.vfs.file_cloning, NULL),
```
mc-worker, line 572: Please document the new option in the mc help.

src/filemanager/file.c
```diff
@@ -2624,13 +2657,28 @@ copy_file_file (file_op_context_t *ctx, const char *src_path, const char *dst_pa
     appending = ctx->do_append;
     ctx->do_append = FALSE;
 
-    // Try clone the file first.
-    if (vfs_clone_file (dest_desc, src_desc) == 0)
+#ifdef HAVE_FILE_CLONING_BY_RANGE
+    // Try clone the file first, but not if the file is in O_APPEND mode
+    if (mc_global.vfs.file_cloning && !(open_flags & O_APPEND))
```
mc-worker, line 2662:
```
if (mc_global.vfs.file_cloning && (open_flags & O_APPEND) == 0)
```

src/filemanager/file.c
```diff
@@ -2624,13 +2657,28 @@ copy_file_file (file_op_context_t *ctx, const char *src_path, const char *dst_pa
...
-        return_status = FILE_CONT;
-        goto ret;
+        if ((appending ? mc_lseek (dest_desc, 0, SEEK_END) >= 0 : TRUE)
+            && vfs_clone_file (dest_desc, src_desc) == 0)
+        {
+            dst_status = DEST_FULL;
+            return_status = FILE_CONT;
+            goto ret;
+        }
+        else if (appending && !(open_flags & O_APPEND))
```
mc-worker, line 2671:
```
if (appending && (open_flags & O_APPEND) == 0)
```
There is no need for `else` after `goto`.

configure.ac
```diff
@@ -380,12 +380,56 @@ case $host_os in
...
+            have_ficlonerange=yes
+            AC_DEFINE(HAVE_FICLONERANGE, 1, [Define if FICLONERANGE is supported on Linux])
+            AC_MSG_RESULT(yes)
+        ],
+        [
+            AC_MSG_RESULT(no)
+        ])
 esac
 
+dnl Check for copy_file_range(2) support (FreeBSD 13+)
```
mc-worker, line 405: These three tests are for different OSes. I think it can be join to the one `case`.

zyv: Oh yes, that makes sense.

tuffnatty: Done.

src/filemanager/file.c
tuffnatty: Fixed.

src/filemanager/file.c
tuffnatty: Fixed.

tuffnatty:
I've updated the pull request with man page updates; unfortunately, the localized man pages were partially outdated, so I had to fix that, and now the PR depends on #5093, #5095, #5096, #5097, #5098, #5099. My Hungarian is definitely not at the level where I could fix the Hungarian man page as well, so I had to keep it as-is.

zyv:
Could you please rebase and resolve the conflicts? Thank you!

tuffnatty:
I've rebased the PR against master and updated the description above to match the actual content.

mc-worker (approved):

zyv:
Thank you very much for your high-quality contributions!

P.S. I didn't realize that we are neighbors :) We can go do some flying one day if you want.

tuffnatty:
@zyv You're welcome! Thanks for your reviewing efforts!

> I didn't realize that we are neighbors :) We can go do some flying one day if you want.

Nice surprise ;)

PR #5089, opened 2026-04-06
Fix Integer overflow in binary CPIO handler

### LINKED ISSUE #5084: Security Issue Report

```shell
Hello mc maintainers.
How can I safely report a security issue in mc?
I don't want to reveal sensitive info in public GitHub issue.
Regards.
```

### Operating system

```shell
Linux
```

- [ ] I confirm the issue is still reproducible with the latest version of Midnight Commander

### How to reproduce

.

### Expected behavior

Some info on website or SECURITY.md on GitHub.

### Actual behavior

No SECURITY.md

### Additional context

_No response_

sectroyer:
Add a bounds check on st_size immediately after it is computed, before it is used in any arithmetic or allocation.  A reasonable upper bound for a symlink target is MC_MAXPATHLEN (4096).

* Resolves: #5084

### DIFF
--- a/src/vfs/cpio/cpio.c
+++ b/src/vfs/cpio/cpio.c
@@ -518,7 +518,9 @@ cpio_create_entry (struct vfs_class *me, struct vfs_s_super *super, struct stat 
                 // FIXME: do we must read from arch->fd in case of inode != NULL only or in any
                 // case?
 
-                inode->linkname = g_malloc (st->st_size + 1);
+                if (st->st_size < 0 || (gsize) st->st_size >= G_MAXSIZE)
+                    return STATUS_FAIL;
+                inode->linkname = g_malloc ((gsize) st->st_size + 1);
 
                 if (mc_read (arch->fd, inode->linkname, st->st_size) < st->st_size)
                 {
@@ -597,7 +599,12 @@ cpio_read_bin_head (struct vfs_class *me, struct vfs_s_super *super)
     st.st_rdev = u.buf.c_rdev;
 #endif
     st.st_size = ((off_t) u.buf.c_filesizes[0] << 16) | u.buf.c_filesizes[1];
-
+    if (st.st_size < 0 || st.st_size > MC_MAXPATHLEN)
+    {
+        message (D_ERROR, MSG_ERROR, _ ("Corrupted cpio header encountered in\n%s"), super->name);
+        g_free (name);
+        return STATUS_FAIL;
+    }
     vfs_zero_stat_times (&st);
 
     st.st_atime = st.st_mtime = st.st_ctime =
@@ -751,7 +758,12 @@ cpio_read_crc_head (struct vfs_class *me, struct vfs_s_super *super)
     u.st.st_rdev = makedev (hd.c_rdev, hd.c_rdevmin);
 #endif
     u.st.st_size = hd.c_filesize;
-
+    if (u.st.st_size < 0 || u.st.st_size > MC_MAXPATHLEN)
+    {
+        message (D_ERROR, MSG_ERROR, _ ("Corrupted cpio header encountered in\n%s"), super->name);
+        g_free (name);
+        return STATUS_FAIL;
+    }
     vfs_zero_stat_times (&u.st);
     u.st.st_atime = u.st.st_mtime = u.st.st_ctime = hd.c_mtime;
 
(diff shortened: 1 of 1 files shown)


zyv (approved):
Thank you for the fix! 👍

src/vfs/cpio/cpio.c
```diff
@@ -518,7 +518,9 @@ cpio_create_entry (struct vfs_class *me, struct vfs_s_super *super, struct stat
                 // FIXME: do we must read from arch->fd in case of inode != NULL only or in any
                 // case?
 
-                inode->linkname = g_malloc (st->st_size + 1);
+                if (st->st_size < 0 || (gsize) st->st_size >= G_MAXSIZE)
```
mc-worker, line 521: Why do you reduce `off_t` to `gsize`? I think, `sizeof (off_t) >= sizeof (gsize)`, so the correct way is
```
st->st_size >= (off_t) G_MAXSIZE
```

src/vfs/cpio/cpio.c
```diff
@@ -597,7 +599,12 @@ cpio_read_bin_head (struct vfs_class *me, struct vfs_s_super *super)
     st.st_rdev = u.buf.c_rdev;
 #endif
     st.st_size = ((off_t) u.buf.c_filesizes[0] << 16) | u.buf.c_filesizes[1];
-
+    if (st.st_size < 0 || st.st_size > MC_MAXPATHLEN)
```
mc-worker, line 602: Why do you compare file size with path length? It is correct for link not for regular file,

src/vfs/cpio/cpio.c
```diff
@@ -751,7 +758,12 @@ cpio_read_crc_head (struct vfs_class *me, struct vfs_s_super *super)
     u.st.st_rdev = makedev (hd.c_rdev, hd.c_rdevmin);
 #endif
     u.st.st_size = hd.c_filesize;
-
+    if (u.st.st_size < 0 || u.st.st_size > MC_MAXPATHLEN)
+    {
```
mc-worker, line 762: Likewise.

sectroyer: You are completely right. I have update the PR w newer fix.

src/vfs/cpio/cpio.c
```diff
@@ -518,7 +518,10 @@ cpio_create_entry (struct vfs_class *me, struct vfs_s_super *super, struct stat
                 // FIXME: do we must read from arch->fd in case of inode != NULL only or in any
                 // case?
 
-                inode->linkname = g_malloc (st->st_size + 1);
+                if (st->st_size < 0 || st->st_size >= (off_t) G_MAXSIZE
+                        || st->st_size > MC_MAXPATHLEN)
```
mc-worker: MC_MAXPATHLEN is definitely less than G_MAXSIZE. Thus 2nd test isn't needed.

zyv:
@sectroyer, are you still planning to do the last fix? I'd like to merge this as soon as master gets unblocked. Thanks!

sectroyer:
@zyv i would definitely would like this fix to be pushed. Can it be applied directly or are there any issues and you would prefer me to first update pr to git head ?

src/vfs/cpio/cpio.c
zyv: @sectroyer I thought this comment was still open. Could you please comment and resolve all remarks which you think are fixed and re-request an approval?

sectroyer:
Here is a list of changes in push request and replies to comments:
Thread 1 (outdated — (gsize) st->st_size >= G_MAXSIZE):

 Fixed in the updated commit — the cast direction is now correct: st->st_size >= (off_t) G_MAXSIZE. 
 
 
Thread 2 (outdated — st.st_size > MC_MAXPATHLEN in cpio_read_bin_head):

 Fixed. The MC_MAXPATHLEN upper bound has been removed from cpio_read_bin_head and cpio_read_crc_head. Those functions handle all entry types, so only the < 0 guard (which catches the
 signed overflow root cause) belongs there. The MC_MAXPATHLEN check is now placed exclusively in cpio_create_entry, inside the S_ISLNK branch where st_size is actually used as a path
 length.
 
 Thread 3 (Likewise / cpio_read_crc_head — already replied):

 Same fix applied — > MC_MAXPATHLEN removed from cpio_read_crc_head.
 
 Thread 4 (new — G_MAXSIZE redundant given MC_MAXPATHLEN):

 You are correct — MC_MAXPATHLEN (4096) is far smaller than G_MAXSIZE, making the >= (off_t) G_MAXSIZE check completely redundant. Removed in the latest commit. The guard is now
simply
 st->st_size < 0 || st->st_size > MC_MAXPATHLEN. 
 
 @zyv I think this resolves all issues ?

PR #5111, opened 2026-05-02
ftp: add MFMT support to preserve file modification time

peripherium:
This adds support for the FTP MFMT command to set the remote modification time after upload.
The command is always sent after upload; if the server does not support MFMT, the response is ignored and no error is raised. This does not cause any malfunction or error messages in mc.

* Resolves: #4456

## Checklist
- [x] I have referenced the issue(s) resolved by this PR (if any)
- [x] I have signed-off my contribution with `git commit --amend -s`
- [x] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #5109, opened 2026-04-28
Add more Debian extensions to filehighlight.ini

sergey-dryabzhinsky:
## Proposed changes

Add mode debian file extensions
- ddeb
- udeb

* Resolves: #5108 

## Checklist

👉 Our coding style can be found here: https://midnight-commander.org/coding-style/ 👈

- [x] I have referenced the issue(s) resolved by this PR (if any)
- [x] I have signed-off my contribution with `git commit --amend -s`
- [ ] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #5113, opened 2026-05-05
Ticket #5113: fix build due to C23 string function prototype changes

zyv:
I just want to unblock the CI, so that I can merge PRs. Therefore, I'm not rewriting the code to make it better, but just trying to make sure that it remains correct and functions like it was before.

The part about `extfs_find_entry_int` I find particularly disgusting. Apparently, this function was modifying strings passed to it through pointer trickery without the callers even knowing. I don't know of a simple way to fix this without rewriting other than just making a copy of the incoming string. The whole logic of this function doesn't make sense to me.

Blocking:

* #5114
* #5111
* #5110
* #5109
* #5089

### DIFF
--- a/lib/serialize.c
+++ b/lib/serialize.c
@@ -80,10 +80,10 @@ static const char *
 go_to_end_of_serialized_string (const char *non_serialized_data,
                                 const char *already_serialized_part, size_t *offset)
 {
-    size_t calculated_offset;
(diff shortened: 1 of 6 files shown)


zyv:
@mc-worker, could you please have a look and fix it directly if you don't like something?

mc-worker (approved):

Issue #1984, opened 2010-01-29
something like syntax.d for highlighting


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/1984 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `stronny at celestia dot ru` |
> | **Mentions** | `zaytsev` (@zyv) |

<p>
From <a class="ext-link" href="http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=520148">Debian bugzilla</a>:<br />
</p>

```
Seems like there is no easy way to add some syntax files to mcedit, as
the only file it parses on startup is
/usr/share/mc/cedit/syntax/Syntax (or ~/.mc/cedit/Syntax, which is not
relevant here). So if some package wants to provide additional
highlighting definitions, the only way to do it is to patch that file,
which is not very nice I think.

So I suggest there should be a directory (say,
/usr/share/mc/cedit/syntax.d) where should reside all *.syntax files,
so if any package wants to add another, it just places it there and
mcedit will find it next time it starts.

It should not be hard to do I think.
```

PR #5133, opened 2026-07-08
include ya?ml.j2 files in yaml syntax highlighting

richlv:
## Proposed changes

Includes YAML Jinja2 files for YAML syntax highlighting.

## Checklist

- [ ] I have signed-off my contribution with `git commit --amend -s`

### DIFF
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -318,7 +318,7 @@ include cuda.syntax
 file ..\*\\.cl$ OpenCL\sProgram
 include opencl.syntax
 
-file ..\*\\.(?i:ya?ml)$ YAML\sFile
+file ..\*\\.(?i:ya?ml|ya?ml\.j2)$ YAML\sFile
 include yaml.syntax
 
 file .\*\\.osl$ OSL\sProgram


zyv:
I think that I understand your intent - probably you're working on Jinja-templated Ansible manifests and want to have at least some sort of highlighting, but I don't think that we can accept this solution upstream.

The proper one would be composite syntaxes for multiple languages, which our engine doesn't support.

Your kludge of just ignoring Jinja constructs and highlighting YAML anyways *might* (or might not) be better than nothing, but this brings another question: where do we stop?

Do we add `j2` extension to every file that can be templated? This is obviously nonsense. And for now we have no precedent for doing this.

I'm afraid that at the moment, the best place for such changes if you really want them is per-user customization.

richlv:
Thanks, noted. Regarding "where do we stop" one approach could be pragmatic - whatever is popular enough :)
But I can understand the desire to keep it max clean.
Is there any way to override or add one line only, and read the rest from the system-wide file?

zyv:
> But I can understand the desire to keep it max clean. 

Oh, it's more like cleanish, or at least not much dirtier than it currently is... :-/

> Is there any way to override or add one line only, and read the rest from the system-wide file?

I *think* you might be able to include the global syntax from the local one, but I haven't checked. This could be a solution for #1984.

PR #5137, opened 2026-07-28
Added mtime and file size in "Find file" results

Tobias-DG3YEV:
Hi,

I started in the mid 80's with the Norton Commander under MS-DOS 3. Then in 1996 I started using Linux with Midnight Commander. I also use Free Commander under Windows. I just can't away from this two panel tools. One thing always missed in mc was this: When using the "Find file" function, I got files with same name results in different directories but was missing meta data of the files found at this place. I wanted to know if it is the same (same date and size) or if it is a newer/different version of the file. Now I added this feature for myself.

Your pull request form I tried filling out:

- [no] I have referenced the issue(s) resolved by this PR (if any)
No PR found for this.
- [yes ] I have signed-off my contribution with `git commit --amend -s`
- [ yes] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [no] I have added tests that prove my fix is effective or that my feature works
How to implement testing for this? I tested it visually with several terminal resolutions (cols, rows) and IMHO it always looks good.
- [no ] I have added the necessary documentation (if appropriate)
Not sure if documentation is even needed for this additional information the upgrade provides.

I hope you and the vast majority of mc find this patch useful. 

Best regards
Tobias

### DIFF
--- a/doc/man/mc.1.in
+++ b/doc/man/mc.1.in
@@ -1421,6 +1421,9 @@ You can start the search by pressing the OK button.
 During the search you can stop from the Stop button and continue from
 the Start button.
 .PP
+The list of found files shows the modification time of each file at the
+right edge, unless the window is too narrow to display it.
+.PP
 You can browse the filelist with the up and down arrow keys. The Chdir
 button will change to the directory of the currently selected
 file. The Again button will ask for the parameters for a new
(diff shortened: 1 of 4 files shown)


mc-worker:
I don't like an idea to modify a base widget WListbox for an every special case. Instead, create a child widget with overridden handlers of some messages (MSG_DRAW at first). Some refactoring of WListbox could be required.

PR #5140, opened 2026-08-16
Control file cloning option directly from the Copy/Move dialog

### LINKED ISSUE #4510: The problem with copying to the same partition
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4510 |
> | **Reporter** | `bindto` (bindto@….ru) |

<p>
partition :  3TB, xfs<br />
</p>
<p>
Linux 6.2.0-36-generic #37~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Oct  9 15:34:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux<br />
</p>
<p>
current version: 4.8.27, tested with 4.8.30, the problem persists <br />
</p>
<p>
Copying files to the same partition works very fast, without calculating the size. A huge 1.7 TB directory was copied in 1 second, the du command run from the root of the partition showed a size larger than the partition size in the df command. The copied files are not mapping or link, the modified target file is different from the source file.<br />
</p>
<p>
This problem only occurs when copying files on the same 3 TB partition, copying from another partition to a 3 TB partition works correctly. <br />
</p>
<p>
The cp command on the same partition works correctly. The file system has no errors.<br />
 <br />
</p>

tuffnatty:
## Proposed changes

Add a checkbox to control the file cloning option to the Copy/Move dialog.

I believe it improves both discoverability and predictability of the option.
The documentation already mentions the option in the Copy/Move dialog section.

* Probably fixes: #4510

## Checklist


mc-worker:
I think it's unnecessary. I think most users don't care about internals of copy process.

tuffnatty:
> I think it's unnecessary. I think most users don't care about internals of copy process.

I believe the opposite (although I don't have any statistically significant data to prove it as well). That is, I believe that mc, as well as other "orthodox 2-panel file managers" (OFM), has lost its value for most non-power users when GUIs has taken their place, and since that moment mc is mostly a power user's tool. And a power user's tool should (IMHO) have its settings well exposed.

Another thing is that I (and probably many others) don't consider zero-copy cloning the same operation as plain copy. AND I would like to be able to do both, without having to look up the setting in menus.

For the reference, far2l, the only other console OFM providing this functionality, has this setting in the Copy dialog.

tuffnatty:
Found and linked the corresponding issue #4510.

Issue #1756, opened 2009-10-29
Automatical update of the file list


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/1756 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `alex_zol` (alex_zol@….ru) |
> | **Mentions** | `powerman-asdf@….ru` |

<p>
Now when you change the list/attributes of files/directories are not from mc, mc panel are not automatically updated and you must to press ctrl + r.<br />
</p>
<p>
This is inconvenient because sometimes you forget do that and you come in strong surprise.<br />
</p>
<p>
Especially surprised people accustomed to the file manager where this feature is implemented (eg nautilus, far).<br />
</p>
<p>
I think that it is not very difficult to implement (notification of fs or through the update timer).<br />
</p>

PR #5141, opened 2026-08-17
Hot-reload the active skin when its file changes on disk

axelfontaine:
This is a prerequisite for mc following live theme-switches on Omarchy (something I've wanted for a while).

Made with Claude.

Notes from Claude:

## Proposed changes

Currently, picking up a change to the current skin's `.ini` file requires restarting mc, or re-selecting the same skin from Options > Appearance. This is inconvenient for skins that are regenerated by an external process — e.g. a script that derives an mc skin from the desktop theme and wants it to apply live, without the user closing and reopening mc.

On Linux, this watches the active skin file with `inotify` and re-applies it automatically on change, reusing the same reload path (`mc_skin_deinit` + `mc_skin_init` + panel/filehighlight refresh + `repaint_screen`) already used by the interactive skin picker in the Appearance dialog (`src/filemanager/boxes.c`'s `skin_apply`), now exposed as `mc_skin_reload()`. The watch fd is integrated into the existing `select()` loop via `add_select_channel()`/`delete_select_channel()`, the same generic mechanism already used for background job and subshell pipes — no new event-loop machinery.

The watch is armed once at startup and intentionally never re-armed on each event: the expected write pattern is an in-place truncate+write, which never invalidates the original watch descriptor. Re-arming from within the event callback was tried first and found to self-trigger forever, since `inotify_rm_watch()` itself raises `IN_IGNORED` on the same fd, which the next `select()` sees immediately — worth calling out in case it comes up in review.

This is Linux-only (`inotify` has no equivalent on the other platforms mc supports), guarded with `#ifdef __linux__`, matching existing precedent elsewhere in the tree (`cons.handler.c`, `key.c`, `subshell/proxyfunc.c`).

Manually verified: launched mc, edited the active skin file's colors on disk, and confirmed the running instance recolors in place with no restart, no keypress, and no measurable CPU overhead at idle or after reload.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — this changes runtime behavior at the process/event-loop level (verified manually as described above); happy to add coverage if there's a preferred pattern for this kind of thing in the existing test suite
- [ ] I have added the necessary documentation — didn't see an obvious place to document this in the man pages; pointers welcome

### DIFF
--- a/src/filemanager/boxes.c
+++ b/src/filemanager/boxes.c
@@ -485,6 +485,16 @@ task_cb (WButton *button, int action)
 /*** public functions ****************************************************************************/
 /* --------------------------------------------------------------------------------------------- */
 
+/* Re-read the currently configured skin from disk and redraw. Used to hot-reload
+   the skin file when it changes on disk (see the skin_watch_* functions in main.c). */
+void
+mc_skin_reload (void)
+{
+    skin_apply (NULL);
+}
+
+/* --------------------------------------------------------------------------------------------- */
+
 void
 about_box (void)
 {
--- a/src/filemanager/boxes.h
+++ b/src/filemanager/boxes.h
@@ -18,6 +18,7 @@
 
 /*** declarations of public functions ************************************************************/
 
+void mc_skin_reload (void);
 void about_box (void);
 void configure_box (void);
 void appearance_box (void);
--- a/src/main.c
+++ b/src/main.c
@@ -31,6 +31,7 @@
 
 #include <config.h>
 
+#include <errno.h>
 #include <locale.h>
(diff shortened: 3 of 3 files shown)


MBroholmA:
+1 to this. Omarchy needs a themed CLI commander for file management

mc-worker:
Skin reloading is a special case of notify support in mc. The main notify action should be an automatic reloading of a directory when file is added/reloaded/modified in it: #1756.

axelfontaine:
@mc-worker The general case still hasn't landed after 17 years. Start with this one use-case now and generalize later? Let's not make perfect the enemy of the good here 😃

Issue #4781, opened 2025-10-08
Slow startup on macOS with bash and intr=^G


egmontkob:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Midnight Commander version and build configuration

```shell
GNU Midnight Commander 4.8.33-353-gc57edfe67
Built with GLib 2.80.4 (using GLib 2.80.4)
Built with NCurses 5.7.20081102 (using ncurses 5.7.20081102)
Built with libssh2 1.11.0
With builtin editor
With subshell support as default
With support for background operations
With mouse support on xterm
Virtual File Systems:
 cpiofs, tarfs, sfs, extfs, ftpfs, sftpfs, shell
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64; uintmax_t: 64;
```

### Operating system

```shell
Darwin Kernel Version 21.6.0 [...] root:xnu-8020.140.49~2/RELEASE_ARM64_T8101 arm64
```

### Is this issue reproducible using the latest version of Midnight Commander?

- [x] I confirm the issue is still reproducible with the latest version of Midnight Commander

### How to reproduce

- macOS
- bash 3.2 as shell
- `stty intr ^G`

Start up mc, observe delay.

### Expected behavior

No delay.

### Actual behavior

About 10 seconds of delay.

### Additional context

Issue is present both without and with the patches from #4625.

Cannot reproduce on Linux. Cannot reproduce with the default `stty intr ^C`. Cannot reproduce with zsh if the patches from 4625 are applied.

On macOS, killing mc (which I had to do a lot while debugging 4625 / 4780) leaves the terminal in a dirty state, including the interrupt character being ^G.

The command `reset` (on both platforms) leaves the interrupt character unchanged. On macOS it prints "Interrupt is control-G (^G)" but this might be somewhat misleading: it did not set this value newly, it just observed and warned about this unusual nondefault value.

Two questions:
- Why does mc leave the terminal with this setting? Is it an mc bug or weird circumstance that we should fix?
- Why does mc have a startup delay with this setting?

egmontkob:
mc does sometimes send the hardcoded byte 0x03 (Ctrl-C) with the intent that it will interrupt whatever's happening inside. 4625's first patch is relevant here.

We could make it dynamic: query the stty settings and send that byte. Or we could say that anything other than ^C is unsupported; still, we could maybe check that on startup and warn the user.

[...]

PR #5148, opened 2026-08-30
Make startup subshell async to speed it up

### LINKED ISSUE #4294: Start a subshell in background
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4294 |
> | **Reporter** | `slajerek` (slajerek@….com) |
> | **Keywords** | `subshell` |

<p>
This is related to ticket #3580, not sure how to re-open tickets so I created a new one.<br />
</p>
<p>
I have a similar problem, but it is not related to the solution described there and was not able to solve that problem. I see on other forums that people have similar problem so I suggest to anyway think about possible fix for this once for all and init subshell in background as described in ticket #3580 (copying here for a reference):<br />
</p>
<p>
<em>It might worth to start the subshell on a background thread to be able to show the main UI faster. Then a rotating dash would be displayed in the command prompt and in Ctrl-O until the subshell is ready.<br />
</em><br />
In my case on macOS 10.15.7 when transitioned to ZSH the MC starts in around 6 seconds. This is totally not acceptable. The ZSH on its own starts in less than one second. The <tt> mc --nosubshell </tt> starts MC instantly, but there's no subshell which is also not acceptable. Thus, there's something fishy with instantiation of subshell in the MC, may be related to the configuration - but actually, that is not the problem of the fact that subshell init takes long, but that it is not done in the background at startup.<br />
</p>

GivAlz:
## Proposed changes

* Resolves: #4294

## Checklist

`init_subshell()` currently forks the user's shell and then blocks synchronously until it finishes sourcing its rc file. On my mac this is very slow and it contributes to #4625 and #4781.

This PR skips that wait: `init_subshell()` still forks the shell at start up but does so asyncronously.
It registers the CWD-report pipe with mc's existing event-loop machinery used for
background-job progress and the steady-state prompt updates in `load_prompt()`, making this a minor code change.

The loading is protected: if the user tries to access the shell before the async loading is done they'll wait for it finish loading.

On My macOS I measured the change (I use oh-my-zsh and other init routines): `init_subshell()`'s blocking window drops from ~850ms to ~10ms.

- [X] I have referenced the issue(s) resolved by this PR (if any)
- [X] I have signed-off my contribution with `git commit --amend -s`
- [X] Lint and unit tests pass locally with my changes (`make indent && make check`)
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added the necessary documentation --> I added some comments, I hope this is enough.

## Notes

I am a long time mc user, but it has been a while since I contributed to an open source project like this: I made this change because the wait time is pushing me towards not using mc. Please let me know if there is anything I missed or I could do better. Thank you!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -727,6 +727,7 @@ tests/lib/vfs/mc.charsets
 tests/lib/widget/Makefile
 tests/src/Makefile
 tests/src/filemanager/Makefile
+tests/src/subshell/Makefile
 tests/src/editor/Makefile
 tests/src/editor/edit_complete_word_cmd_test_data.txt
 tests/src/vfs/Makefile
--- a/src/subshell/common.c
+++ b/src/subshell/common.c
@@ -180,7 +180,9 @@ typedef enum
  * the current command buffer and the location of the cursor. */
 #define SHELL_BUFFER_KEYBINDING "_"
 
-/*** forward declarations (file scope functions) *************************************************/
+/*** file scope functions ************************************************************************/
+
+static int subshell_init_ready (int fd, void *info);
 
 /*** file scope variables ************************************************************************/
 
@@ -1702,8 +1704,6 @@ do_subshell_chdir (const vfs_path_t *vpath, gboolean force, gboolean update_prom
 void
 init_subshell (void)
 {
-    vfs_path_t *vfs_subshell_cwd;
-
     // This must be remembered across calls to init_subshell()
     static char pty_name[BUF_SMALL];
 
@@ -1823,7 +1823,19 @@ init_subshell (void)
         g_free (precmd);
     }
 
-    // Wait until the subshell has started up and processed the command
+    // Finish the handshake asynchronously from the main loop once the shell is ready
+    subshell_initialized = FALSE;
+    add_select_channel (subshell_pipe[READ], subshell_init_ready, NULL);
+}
+
+/* --------------------------------------------------------------------------------------------- */
+// Second half of init_subshell(): consume the shell's first CWD report, probe the persistent
+// command buffer and get a fresh prompt printed. Runs from the main loop as soon as the shell
+// is ready, or right away from invoke_subshell() if the user got there first.
+
+static void
+subshell_finish_init (void)
+{
     subshell_state = RUNNING_COMMAND;
     tty_enable_interrupt_key ();
     if (!feed_subshell (QUIETLY, TRUE))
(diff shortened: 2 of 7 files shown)


zyv:
Oh man, after all the tons of work that went into fixing subshell initialization races, this sounds maximally dangerous, and it comes with no tests. I wish @egmontkob could look into this, because he did the earlier fixing, but I guess he wouldn't have time and interest. And I don't have time at the moment, unfortunately.

GivAlz:
> Oh man, after all the tons of work that went into fixing subshell initialization races, this sounds maximally dangerous, and it comes with no tests. I wish @egmontkob could look into this, because he did the earlier fixing, but I guess he wouldn't have time and interest. And I don't have time at the moment, unfortunately.

Yeah, I was kind of wondering if my "little change" would be dangerous for everything else.

I can think of some tests & try to fix the failing builds but if this step is so relevant for the whole project I understand this might wait...indefinitely...anyhow thank you!

Edit: builds are failing because I forgot to commit the test folder (sorry), I'll double check them and push them in the next days.

GivAlz:
I force pushed because I also did a rebase.

I reviewed the tests I had written and now I pushed them. I tried to copy the style from the other tests.

ossilator:
so much "fun" ...
i may have a look at this later, but right now i'm still a tad "indisposed".

please squash the commits, and write a _proper_ commit message. elaborate on how the event dependencies change, which potential problems you have considered, etc. Questions raised in the long thread's on egmont's PRs may also be relevant targets. some "answers" may be best provided by test cases.

GivAlz:
Hi!

Sorry for taking long but the failing build lead me to finding a few real issues in the async window and the earlier subshell threads:

1. Build: tests/src/subshell was compiled unconditionally, which broke --without-subshell. It is now only built under ENABLE_SUBSHELL, like src/subshell.
2. Panel navigation before the handshake: Changing directory in the panels sent a cd to a shell that was still sourcing its rc files. subshell_chdir() now ignores calls until the handshake is done.
3. Shell dying before the handshake: Exec failure or an exit in a startup file left use_subshell on. The blocking code used to turn it off; sigchld_handler() now does the same when the shell dies before initialisation.
4. Restart after exit in the subshell: subshell_initialized stayed set from the previous shell, so the restarted one skipped the wait. init_subshell() now resets it.

Tests now cover also the guards for handshake and navigation.

I now squashed everything into a single commit and tried to write a detailed commit message, describing the even flow before and after all changes I introduced.

The performance is the same as before: on my machine I go from about 1s loading time to 30ms.

To be clear: working on this I realised how this "little change" interacts with a lot of different functionalities; I hope there are no side effects I am not seeing, but I might be wrong...

PR #5149, opened 2026-09-02
Ticket #5112: highlight OpenSSH ssh_config / sshd_config files

### LINKED ISSUE #5112: syntax: default highlighting for OpenSSH client/server config (ssh_config, sshd_config)

When opening `/etc/ssh/sshd_config` (or `/etc/ssh/ssh_config`, and their `*.d/*.conf` drop-ins) in `mcedit`, no syntax highlighting is applied and Options -> Syntax Highlighting shows `unknown`. These are among the most commonly edited system files on any Linux host, and editing them with highlighting would be nice.

The reason is that `misc/syntax/Syntax.in` only maps files whose name ends in `.cfg`, `.cnf`, or `.conf` to `conf.syntax`:

```
file ..\*\\.(?i:cfg|cnf|conf)$ Config\sfile
include conf.syntax
```

Since `sshd_config` and `ssh_config` use an underscore and have no extension, they fall through to the final `file .\* unknown` rule and get `unknown.syntax`.

Map OpenSSH's canonical config filenames to a syntax definition by default. Two reasonable options:

1. **Minimal (preferred for a first step)** — reuse the existing generic `conf.syntax`. In `misc/syntax/Syntax.in`, before the `unknown` catch-all, add:

   ```
   file (.*/)?ssh(d)?_config(\\..*)?$ OpenSSH\sconfig\sfile
   include conf.syntax
   ```

   This covers `ssh_config`, `sshd_config`, and drop-ins under `ssh_config.d/` / `sshd_config.d/` (e.g. `50-cloud-init.conf` users typically also have a parent `ssh_config` to edit). It immediately fixes the "unknown" status and gives `#`-comments, quoted strings, numbers, and `=` highlighting from `conf.syntax`.

2. **Nicer (optional follow-up)** — ship a dedicated `misc/syntax/sshd.syntax` (or `openssh.syntax`) that additionally highlights the well-known OpenSSH keywords as defined in [`sshd_config(5)`](https://man.openbsd.org/sshd_config) and [`ssh_config(5)`](https://man.openbsd.org/ssh_config) — e.g. `Host`, `Match`, `Port`, `ListenAddress`, `PermitRootLogin`, `PasswordAuthentication`, `PubkeyAuthentication`, `AuthorizedKeysFile`, `AllowUsers`, `Ciphers`, `MACs`, `KexAlgorithms`, `HostKey`, `Subsystem`, `Include`, etc., and the `yes`/`no`/`prohibit-password` value tokens. Then point the `Syntax.in` rule above at it instead of `conf.syntax`.

Either way, the user-visible improvement is: opening `sshd_config` in mcedit no longer shows "unknown" and gets at least baseline coloring out of the box.

- Temporarily tenaming/symlinking system files to `sshd.conf` so the existing `*.conf` rule kicks is
- Relying on the first-line regex in `Syntax.in`. `sshd_config` typically starts with `# $OpenBSD: sshd_config,v ...` but it's not unique enough to safely key off, so a filename rule is the right place.

### Additional context

Happy to send a PR for option 1 (one-liner in `Syntax.in`) and if desired, a follow-up PR adding a dedicated `sshd.syntax`.

valleyco:
## Proposed changes

Map extensionless OpenSSH client and server config filenames (`ssh_config`, `sshd_config`, and dotted backups such as `sshd_config.bak`) to the existing `conf.syntax` highlighter so mcedit no longer reports them as `unknown`.

Drop-ins under `*.d/*.conf` were already covered by the generic `*.conf` rule.

* Resolves: #5112

## Checklist

### DIFF
--- a/misc/syntax/Syntax.in
+++ b/misc/syntax/Syntax.in
@@ -357,6 +357,9 @@ include toml.syntax
 file .\*\\.(mch|ref|imp)$ B\sFile
 include b.syntax
 
+file (.\*/)?ssh(d)?_config(\\..\*)?$ OpenSSH\sconfig\sfile
+include conf.syntax
+
 file ..\*\\.(?i:cfg|cnf|conf)$ Config\sfile
 include conf.syntax
 


valleyco:
Closing to re-open with a cleaned-up description.

Issue #3987, opened 2019-05-17
Midnight Commander loses content of the current directory


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3987 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `memy` (michael@….net) |

When I am browsing a mounted SMB directory Midnight Commander sometimes loses directory content.
When I hit Ctrl-r the content of the directory reappears.

It started today after upgrading of the kernel and samba.
All other means of browsing file system work perfectly fine: shell, Thunar, File Open dialogs, etc.
I have shares mounted on different servers:

* Synology NAS
* Raspberry Pi `4.19.42-1-ARCH #1 SMP PREEMPT Wed May 15 00:39:47 UTC 2019 armv7l GNU/Linux`
* Dell mini `5.1.2-arch1-1-ARCH #1 SMP PREEMPT Wed May 15 00:09:47 UTC 2019 x86_64 GNU/Linux`

MC demonstrates the same behavior on all of the shares.<br />

When I browse to a directory on a mounted SMB share MC occasionally shows empty directory. Re-read immediately restores the panel. The same happens when I am returning back to the parent directory.

The shares are mounted at boot time using systemd like this but the problem persists with any other way of mounting the SMB share:

```
[Unit]
Description=Mount Share at boot
Requires=network-online.target
After=network-online.target
Wants=network-online.target

[Mount]
What=//host/share
Where=/mnt/SMB/share
Options=uid=<my-uid>,credentials=/home/username/secret.txt,iocharset=utf8,rw,x-systemd.automount
Type=cifs
TimeoutSec=30

[Install]
WantedBy=multi-user.target
```

I run Arch Linux.
Kernel version: `5.1.2-arch1-1-ARCH #1 SMP PREEMPT Wed May 15 00:09:47 UTC 2019 x86_64 GNU/Linux`
samba version: 4.10.3-1 upgraded from 4.10.2-1

LC_MESSAGES=C mc -V

```
GNU Midnight Commander 4.8.22
Built with GLib 2.58.3
Using the S-Lang library with terminfo database
With builtin Editor
With subshell support as default
With support for background operations
With mouse support on xterm and Linux console
With support for X11 events
With internationalization support
With multiple codepages support
Virtual File Systems: cpiofs, tarfs, sfs, extfs, ftpfs, sftpfs, fish, smbfs
[...]

Issue #4635, opened 2025-01-25
Fuse mounted directory under kopia does not show contents in mc


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4635 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `rahrah` (richard1hoyle@….com) |

<p>
OS:      Slackware64-current<br />
LIBC:    glibc-2.40<br />
Kernel:  6.12.10<br />
ARCH:   x86_64<br />
</p>

```
mc -V
GNU Midnight Commander 4.8.33
Built with GLib 2.82.4
Built with S-Lang 2.3.3 with terminfo database
Built with libssh2 1.11.1 
With builtin editor and aspell support
With subshell support as default
With support for background operations
With mouse support on xterm and Linux console
With support for X11 events
With internationalisation support
With multiple codepages support
With ext2fs attributes support
Virtual File Systems:
 cpiofs, tarfs, sfs, extfs, ftpfs, sftpfs, shell
Data types:
 char: 8; int: 32; long: 64; void *: 64; size_t: 64; off_t: 64; uintmax_t: 64;

```
<h2 id="Description">Description</h2>
<p>
Using mc with kopia (fuse) mounted snapshots.  Cli commands work fine, but mc sees an empty directory.<br />
</p>
<p>
I found mc 4.8.25 did not show this problem, but mc 4.8.26 did.  I used a git bisection to find the issue.<br />
</p>
<h2 id="Kopia">Kopia</h2>
<p>
Kopia is a backup program written in Go.  It uses go-fuse for its fuse kernel module interface.  Backups are accessed in directory hierarchies.  Inserted into the hierarchy path string are machine names and date and time stamps.<br />
</p>
<p>
Kopia can be found here:<br />
</p>
<p>
<a class="ext-link" href="https://kopia.io">https://kopia.io</a><br />
</p>
<p>
and<br />
</p>
<p>
<a class="ext-link" href="https://github.com/kopia/kopia/">https://github.com/kopia/kopia/</a><br />
</p>
<p>
The<br />
</p>
<p>
mount -l<br />
</p>
<p>
gives this after a kopia mount operation on my Linux test machine:<br />
</p>

```
kopia on /mnt-test/kopia/mount type fuse.kopia (rw,nosuid,nodev,noatime,user_id=0,group_id=0,max_read=131072)
```
[...]

Issue #4590, opened 2024-09-21
Extended attributes - Cannot get attributes of source directory - Invalid argument (22)


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4590 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `andyrozman` (@andyrozman) |

<p>
Version: 4.8.32-12-g9e06a50eb<br />
OS: Ubuntu 24.04.1 LTS<br />
</p>
<p>
Sometimes when I am copying files I get red screen "Cannot get attributes of source directory - Invalid argument (22)".<br />
</p>
<p>
In most cases I can use "<tt>Skip/Skip</tt> All" function and files are still moved or copied... <br />
</p>
<p>
This is failing when I am trying to copy files on veracrypt encrypted volume (I have full permission on drive, even started with umask=000, which gives full permissions). Copying of folders in this case fails... If I enter the directory and copy files, I still get upper message, but I can still copy the files.<br />
</p>
<p>
I started receiving this errors since I did my own build of mc few weeks ago (before that I used version supplied with Ubuntu 22.04 LTS). After I upgraded Ubuntu to 24.04 last weekend, this error started preventing copying folders.<br />
</p>
<p>
It would be nice if this could be fixed. I rely on mc for most of operations.<br />
</p>
<p>
I tried to copy files with different software and copying there works without problems...<br />
</p>

> [!NOTE]
> Original attachments:
>
> * [Screenshot 2024-09-24 at 07.32.26.png](https://github.com/MidnightCommander/trac-archive/blob/master/attachments/ticket/4590/Screenshot%202024-09-24%20at%2007.32.26.png) ([raw](https://github.com/MidnightCommander/trac-archive/raw/master/attachments/ticket/4590/Screenshot%202024-09-24%20at%2007.32.26.png)) by `zaytsev` (@zyv) on <span title="2024-09-24T05:39:53+00:00">`Sep 24, 2024 at 5:39 UTC`</span>
[...]

Issue #4289, opened 2021-09-08
sshfs mounts cannot be browsed with mc


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/4289 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `basz` (mc@….eu) |
> | **Keywords** | `sshfs`, `empty`, `browse` |

<p>
remote directory mounted with <tt>sshfs</tt> (FUSE) can be normally used from cmd-line, however it cannot be browsed with <tt>mc</tt> 2.8.26. the same mount point, on the same host, works fine with <tt>mc</tt> 2.8.22.<br />
</p>
<p>
example steps to reproduce:<br />
</p>
<ol><li><tt>/etc/fstab</tt> mount looks like this:

```
user@host:/mnt/remote/stuff  /mnt/stuff  fuse.sshfs 
 noauto,_netdev,user,transform_symlinks,identityfile=/path/to/key,default_permissions,uid=user,gid=user,allow_other,cache=no   0   0
```
</li></ol><ol start="2"><li><tt>docker run -it --rm -v /mnt/stuff:/mnt:ro debian:11</tt>
</li></ol><ol start="3"><li><tt>apt update &amp;&amp; apt install mc</tt>
</li></ol><ol start="4"><li><tt>mc /mnt</tt>
</li></ol><p>
directory  will appear to be empty, even though <tt>ls</tt> on the same dir shows real content.<br />
</p>
<p>
changing debian image version in step 2 from 11 to 10 allows to browse mountpoint's content as usual.<br />
</p>
<p>
note that <tt>mc</tt> 2.8.22 is the default in <tt>debian:10</tt> (buster - aka: oldstable) and <tt>mc</tt> 2.8.26 is the default in <tt>debian:11</tt> (bullseye - aka: stable).<br />
</p>

PR #5156, opened 2026-09-16
fix(local_opendir): clear stale errno before the EINTR check

KuzinAndrey:
## Proposed changes

The Ticket #3987 workaround (27de03754) reopens the directory when the first readdir() returns NULL and errno == EINTR, to recover from an interrupted readdir() on CIFS shares with Linux >= 5.1.

The check reads a stale errno: readdir() does not set errno when the directory stream is exhausted (man 3 readdir: "errno is not changed"), and errno is a per-thread value that any previously interrupted syscall leaves behind. mc's event loop routinely survives EINTR from select() and read() when the subshell's SIGCHLD or the terminal's SIGWINCH arrives, and success paths such as open, fstat and opendir do not clear it either.

With an empty directory and a stale errno == EINTR the condition is true on every iteration -- no syscall in the loop body writes errno on the success path -- so local_opendir() spins forever: opendir, fstat, getdents, closedir, repeat. Observed on a FUSE mount: four FUSE round trips per iteration (~80k/s), ~27% stime in mc and ~34% in the FUSE daemon, until the directory becomes non-empty or the process exits. On a local filesystem the same loop saturates a core with plain syscalls.

Set errno = 0 before the readdir() call so that only an EINTR reported by that readdir() itself triggers the reopen.

## Checklist


src/vfs/local/local.c
```diff
@@ -106,6 +106,7 @@ local_opendir (const vfs_path_t *vpath)
         if (dir == NULL)
             return NULL;
 
+        errno = 0;
         if (readdir (dir) == NULL && errno == EINTR)
```
ossilator, line 110: fwiw, the NULL && !EINTR case should be handled in a separate path, as the rewinddir() obviously makes no sense then.

ossilator:
looks reasonable.

is this code covered by autotests by any chance? then an extension would be in order.

i find the commit message somewhat excessive - it doesn't really matter how errno can become non-null outside the strictly controlled local scope; it's an obvious programming error.

KuzinAndrey:
I am currently developing a FUSE filesystem for personal use, and I ran into an issue where Midnight Commander would hang when accessing an empty directory. An AI agent identified the problematic spot through static analysis and by examining the state of the hung processes. I want to apply a quick fix without altering the existing codebase unnecessarily. I’m not sure how to write unit tests for this, as it involves a complex combination of system calls and the launching of external processes.

The commit message provides a detailed explanation of the situation to ensure there is no ambiguity.

A quick Google search on using `readdir` reveals a common recommendation: reset `errno` to zero before making the call. The existing code failed to do this.

ossilator:
> A quick Google search on using readdir reveals a common recommendation: reset errno to zero before making the call. The existing code failed to do this.

yes, exactly. that's all that is to say about this. the long analysis how it actually happens is entirely irrelevant.

src/vfs/local/local.c
```diff
@@ -106,6 +106,7 @@ local_opendir (const vfs_path_t *vpath)
         if (dir == NULL)
             return NULL;
 
+        errno = 0;
```
zyv, line 109:
```suggestion
        // Reset errno before readdir to avoid reading a stale EINTR (#5156)
        errno = 0;
```

zyv (approved):
Could you please adjust the commit message to start like all other ones in our project:

https://github.com/MidnightCommander/mc/commits/master/

"Ticket #5156: vfs - clear stale errno before checking for EINTR"

I also think that the rest of the commit message can be removed - it carries no additional information. Instead, I'd add a comment before reset to reference this ticket - if needed one can always check it for background information.

Otherwise LGTM, thank you. Unfortunately, we don't have VFS test for local VFS / `local_opendir`. I would also very much like to see one, but I wouldn't make it a prerequisite for merging this fix.

KuzinAndrey:
I don't have time right now to make immediate changes; the @ossilator is also flagging issues in the loop involving the `rewinddir` call, so the logic for this entire section might need rethinking. You can amend the commit if these fixes are important, or I can try to rework the algorithm entirely later on.

zyv:
> I can try to rework the algorithm entirely later on.

I'm happy to wait unless Andrew wants to take care of this immediately. Thank you.

src/vfs/local/local.c
KuzinAndrey: I completely reworked this block; I think that’s exactly what was meant.

ossilator: yes, thanks, but the "fwiw" was meant to highlight that this is a pre-existing problem, and optional to fix. the clean way to address this would be adding a separate commit on top (it should be trivial to split the commit with `git gui` or another interactive tool).

i'm assuming that the maintainers won't make a fuss about it either way.

zyv:
I have got $100 worth of promotional cloud credits from Anthropic for free and decided to use them to review this PR. The initial review used about 3M tokens over 1 hour of wall-clock time, but the results are quite impressive.

The agents wrote a minimal FUSE-based file system to reliably reproduce the problem and check that it's gone, as well as analyze the code in terms of other possible issues, approaches to fixing those, and corresponding tradeoffs.

I find it entertaining that they also didn't like the comment and proposed a change along the lines that I've suggested, and also echoed @ossilator's standing preference for splitting commits, which I actually concur with on principle, but am more susceptible to accept less perfect splits if nobody's willing to put the work into it (including myself). Otherwise, they have

* found a similar issue in the editor and signal handlers
* suggested that it's actually trivial to write a regression / unit test
* pointed to a possible fix at a different level to decrease the penalty on FUSE-based FS
* noted that skipping `rewinddir` as suggested and implemented breaks on XNU / BSD
* surfaced a number of other pre-existing issues around the loop

I will push the fixes to this PR and update it with what is open, but what I'd rather not do, or not in this PR.

src/vfs/local/local.c
```diff
@@ -138,7 +138,13 @@ local_readdir (void *data)
     struct dirent *d;
     unsigned char type;
 
-    d = readdir (*(DIR **) data);
+    // Retry if interrupted, resetting errno to avoid reading a stale EINTR (#5156)
+    do
+    {
+        errno = 0;
+        d = readdir (*(DIR **) data);
```
ossilator: wouldn't this also be affected by the need to rewinddir() under freebsd?

ossilator:
i don't know whether this is genuine or just an artifact of gh's rendering, but the attribution of the AI commits seems fishy - they seem to have both the author set to claude _and_ a co-authored-by claude trailer, which makes no sense.

tests/src/vfs/local/local_opendir.c
```diff
@@ -0,0 +1,199 @@
...
+static struct dirent *
+mock_readdir (DIR *dirp)
+{
+    if (mock_readdir__eof)
+        return NULL;  // like readdir(), leaving errno untouched
+
+    if (mock_readdir__fail_after < 0)
+        return readdir (dirp);
+
+    // FreeBSD and macOS keep the advanced offset if reading fails after partial progress
```
ossilator: well, this is an _interesting_ data point ...

the tested code should also get a comment, so they next one to see the code doesn't suggest the same cleanup as i did ...

zyv:
I've pushed the updates. Here are the remaining open issues from Claude review and my evaluation:

| # | Where | Finding | Evaluation |
|---|---|---|---|
| 1 | `local_opendir()` loop | No retry cap: a `readdir()` that fails with `EINTR` every time (e.g. a FUSE daemon answering every `READDIR` with `EINTR`) makes mc spin forever | Pre-existing, sounds more like a local DoS from an evil demon; I'd leave it as is |
| 2 | `local_opendir()` probe | A probe `readdir()` failing with anything but `EINTR` (`EIO`, `EACCES`, `ESTALE`, …) is swallowed and the directory is shown as empty | Pre-existing, risky behavioral change that could flood users with error messages (#4590) |
| 3 | `local_opendir()` probe | Probe + `rewinddir()` empties listings on FUSE filesystems that cannot seek directories (#4289, #4635) and fetches the first `READDIR` batch twice | Pre-existing, trade-off accepted in #4635, fixing needs a redesign, e.g. keeping the probed entry instead of rewinding |

Findings along the same lines outside the scope:

| # | Where | Finding | Evaluation |
|---|---|---|---|
| a | `read_one_line()` in `src/editor/syntax.c` | Same stale-`errno` bug class: `continue` on `EINTR` without `clearerr()` spins at EOF after one interrupted read | Real-world EINTR source uncertain (FUSE daemon answering FUSE_INTERRUPT on sync reads?); fix: `clearerr (f);` (and `errno = 0;`) before `continue` |
| b | `SIGCHLD`/`SIGWINCH` handlers | The handlers don't save and restore `errno`, so a signal between `readdir()` and the check can replace a real `EINTR` | A storm test hit it 3 times in 20M iterations: unlikely; fix: save and restore errno in each handler |

@mc-worker, maybe you can consider fixing the editor problem since you care about the editor.

My promotional $100 are gone, and my conclusion is that this thing can do a lot of good if used appropriately, but it can also work as a mega-slopgun and maintainer DoS amplifier, which we are currently observing elsewhere.

I wonder if open source projects can get free access to this on a permanent basis. For me, it would make it possible to process PRs (at all) using my limited time. Or maybe I should apply as a Claude driver if such jobs exist.

I'm interested in your opinions.

zyv:
> i don't know whether this is genuine or just an artifact of gh's rendering, but the attribution of the AI commits seems fishy - they seem to have both the author set to claude _and_ a co-authored-by claude trailer, which makes no sense.

This is apparently how Claude is set up to do it. It created the commits, so it sets itself as the author and I'm the committer, and it also sets the co-authored tag. This is also what I saw in other projects. I'm fine with that, but I could, of course, reset the author to myself. Is there any reason to bother, or is it fine as it is?

tests/src/vfs/local/local_opendir.c
zyv: You are right, I've extended the comment sentence before the loop. Thanks.

src/vfs/local/local.c
zyv: Yes, that's correct; it only really fixes the problem on Linux, but on BSD the interrupted batch is lost (and the rest of the listing is recovered).

I've added a comment to that effect; fixing it properly would correspond to the item (3) in my table (getting rid of `rewinddir` completely and redesigning the whole thing).

Issue #3635, opened 2016-04-21
Develop tests for FISH scripts


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/3635 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `zaytsev` (@zyv) |

<p>
Currently, FISH scripts are a can of worms: there are no tests and any change risks to subtly break part of the functionality. It would be great if we could cover them with some kind of tests, such that the scripts would become more maintainable.<br />
</p>
<p>
See #3611 for more details.<br />
</p>

Issue #77, opened 2008-12-26
savannah: fish copies files to tmp before transfer -> fails with low disk space


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/77 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `tropikhajma` |
> | **Mentions** | `pahan@….info` (@Hubbitus), `bunk@….de`, `god12@….hut.fi` |

<p>
Original: <a class="ext-link" href="http://savannah.gnu.org/bugs/?21692">http://savannah.gnu.org/bugs/?21692</a><br />
</p>
<table class="wiki">
<tr><td>Submitted by:</td><td>hajma &lt;tropikhajma&gt;</td><td>Submitted on:</td><td>Sat 01 Dec 2007 02:34:28 PM UTC
</td></tr><tr><td>Category:</td><td>VFS</td><td>Severity:</td><td>3 - Normal
</td></tr><tr><td>Status:</td><td>None</td><td>Privacy:</td><td>Public
</td></tr><tr><td>Assigned to:</td><td>None</td><td>Open/Closed:</td><td>Open
</td></tr><tr><td>Release:</td><td>4.6.1</td><td>Operating System:</td><td>GNU/Linux
</td></tr></table>
<p>
Original submission:<br />
</p>

```
To reproduce:
1. have a machine with a 4G file on it and just 1G free space
2. mc
3. Left -> Shell connection, enter remote machine name and connect
4. In the right pane, locate the big file and hit F5 to copy it to 
the remote machine

result:
while mc pretends it is transferring the file to the remote machine,
it is actually copying it to ~/tmp/mc-$USER.
If I had enough space it would start transfer AFTER the file is 
completely copied (after it CLAIMS to user 100% was transferred).
But since I do not have this much space, I am SOL and it fails.

mc --version
GNU Midnight Commander 4.6.1
Virtual File System: tarfs, extfs, cpiofs, ftpfs, fish, undelfs
With builtin Editor
Using system-installed S-Lang library with terminfo database
With subshell support as default
With support for background operations
With mouse support on xterm and Linux console
With internationalization support
With multiple codepages support

Mandriva Linux 2008.0 here

the patches applied by distribution are available here:
http://svn.mandriva.com/cgi-bin/viewvc.cgi/packages/updates/2008.0/mc/current/SOURCES/
```

[...]

Issue #68, opened 2008-12-26
savannah: Aborting a FISH file transfer is grossly inefficient


mc-butler:
> [!IMPORTANT]
> This issue was migrated from Trac:
>
> | **Origin** | https://midnight-commander.org/ticket/68 |
> | :------------- | :------------------------------------------------------------------------------------------------ |
> | **Reporter** | `ptsekov` |
> | **Mentions** | `god12@….hut.fi` |

<p>
Original: <a class="ext-link" href="http://savannah.gnu.org/bugs/?19721">http://savannah.gnu.org/bugs/?19721</a><br />
</p>
<table class="wiki">
<tr><td>Submitted by:</td><td>Pavel Tsekov &lt;ptsekov&gt;</td><td>Submitted on:</td><td>Fri 27 Apr 2007 08:40:45 AM UTC
</td></tr><tr><td>Category:</td><td>VFS</td><td>Severity:</td><td>3 - Normal
</td></tr><tr><td>Status:</td><td>None</td><td>Privacy:</td><td>Public
</td></tr><tr><td>Assigned to:</td><td>None</td><td>Open/Closed:</td><td>Open
</td></tr><tr><td>Release:</td><td>All versions</td><td>Operating System:</td><td>All
</td></tr></table>
<p>
Original submission:<br />
</p>

```
I was looking at the fish code recently and noticed that aborting
a running file transfer still causes MC to read the whole file sent 
by the remote end. I realized that the way FISH is currently
implemented, i.e. commands and data sent over the same channel, this
 is the only way to clear the data channel so that command replies 
would get to the FISH layer without re-establishing a new link. 
While this could be acceptable for small transfers I doubt it that 
it makes sens for multi-megabyte files.

Ideas on how to fix it are welcome. One way I can see is to
open a separate FISH connection for the data transfer i.e. like FTP. 
```
<p>
Comment 1 by Oswald Buddenhagen &lt;ossi&gt; at Fri 27 Apr 2007 12:20:17 PM UTC:<br />
</p>

```
no, i think we can do like ssh does, i.e., tunnel multiple virtual 
connections through one physical connection. this adds some 
overhead, though (especially cpu-wise, as we have to call dd for 
every chunk).

btw, you might want to look at kde's fishserv.pl, it has some 
optimizations. never looked at it myself, though.
```
<p>
Comment 2 by Pavel Tsekov &lt;ptsekov&gt; at Fri 27 Apr 2007 01:38:47 PM UTC:<br />
[...]

PR #5157, opened 2026-09-17
vfs/shell: handle remote errors on send

### LINKED ISSUE #3128: dataloss: moving a file and skipping upon problem deletes original file

tuffnatty:
## Proposed changes

I've hit #3128 and tried to fix it.

The old workflow during moving a file to a write-protected remote location looked like:
1. `copy_file_file()` calls `mc_open()` on the destination vpath.
2. `mc_open()` calls `shell_open()`
3. `shell_open()` generates a directory entry in the local VFS cache and returns a handle to it.
4. `copy_file_file()` transfers the file content into the local temporary copy and sets `dst_status` to `DEST_FULL`.
5. `copy_file_file()` calls `mc_close()`.
6. `mc_close()` calls `shell_file_store()`.
7. `shell_file_store()` calls remote script, `src/vfs/shell/helpers/send`.
8. `send` answers with `### 001`.
9. `shell_file_store()` receives `### 001` and starts to send file data.
10. `send` fails to truncate the destination file with `> "${FILENAME}"`, ignores the error, and starts the loop to consume exactly the specified number of bytes, attempting to write them to the write-protected destination path and ignoring any errors.
11. After consuming the specified number of bytes, `send` answers with `### 200`.
12. `mc_close()` returns successfully.
13. `copy_file_file()` tries to chmod and chown the file, fails, the user chooses to Ignore the errors.
14. `copy_file_file()` returns `FILE_CONT`, and `move_file_file()` removes the source file, leaving no copy neither in source nor in destination.

The simplest fix in the scenario with a write-protected destination is provided in the first patch, which adds an attempt to truncate the file or append 0 bytes to the file in `shell_open()` before creating the fake directory entry in the local VFS cache, and returns an error on failure. It also has a positive side effect of failing early, before creating the local temporary copy etc.

But the first patch alone does not solve the issue when the remote error occurs later during the copy, e.g. in an out of space situation. So the task was to make the `send` (and `append`) script detect and report the errors while also still working on the lowest common denominator of a shell environment. To do it, I had to implement some tests and some primitive test infrastructure for these scripts, which, as I've found out later, was already discussed in #3635. The test scripts are designed to run wherever the GitHub CI is running, they assume a fairly POSIX-compliant sh and are known to work in `dash`, `bash`, and FreeBSD Almquist shell. It's the subject of the second patch.

The third patch contains the updated `send` and `append` scripts. They are written in an archaic dialect, so that [Heirloom Bourne shell](https://heirloom.sourceforge.net/sh.html) can run them. They handle the write errors, consume the rest of the data, and respond with `### 500`.

But even when the scripts return `### 500` and `mc_close()` returns an error, `dst_status` is still equal to `DEST_FULL`, which is obviously a lie. The fourth patch catches this condition and changes `dst_status` to `DEST_SHORT_QUERY`, asking the user to choose to keep or delete the partial file, and preventing the deletion of the source file in `move_file_file()`.

* Resolves: #3128.
* Probably also makes #68, #77 more bearable in some situations.
* Probably is also a fix for #3961, but it needs to be confirmed.

~~* Needs: #5155.~~ Well it's not a hard dependency, so I have rebased this PR on the master branch to ease reviewing.

## Checklist


src/vfs/shell/helpers/append
```diff
@@ -1,15 +1,62 @@
+case "${SHELL_HAVE_POSIX_DD-0}${SHELL_HAVE_HEAD-0}${SHELL_HAVE_PERL-0}${SHELL_HAVE_DD-0}" in
```
ossilator: this commit does too many things at once and should be split.

src/vfs/shell/shell.c
```diff
@@ -1686,6 +1686,28 @@ shell_fill_names (struct vfs_class *me, fill_names_f func)
...
+        char *quoted_name;
+        char *command;
+        struct vfs_s_super *super;
+        struct vfs_class *me;
+        int r;
+
+        me = VFS_CLASS (vfs_path_get_last_path_vfs (vpath));
+        name = vfs_s_get_path (vpath, &super, 0);
+        quoted_name = str_shell_escape (name);
+        command = g_strdup_printf ("if %s /%s; then echo '### 200'; else echo '### 500'; fi\n",
```
ossilator: this construct smells. it probably has a good chance of failing with some shells. i'd feel a lot more comfortable with `if : >> file;` instead, though it's still weird.

and why the leading slash? it's really missing from the file name?

src/vfs/shell/Makefile.am
```diff
@@ -38,7 +38,7 @@ shelldef.h: $(SHELL_HELPERS:%=$(srcdir)/helpers/%)
 		_upper="$$(printf '%s' "$$_src" | $(SED) -e 'y/abcdefghijklmnopqrstuvwxyz/ABCDEFGHIJKLMNOPQRSTUVWXYZ/' -e 's/FEXISTS/EXISTS/')"; \
 		printf '#define VFS_SHELL_%s_DEF_CONTENT ( \\\n' "$$_upper"; \
 		$(SED) \
-			-e 's/[\\"]/\\&/g' \
+			-e 's/[?\\"]/\\&/g' \
```
ossilator: looks like this should be squashed into one of your other commits.

ossilator (changes requested):

src/vfs/shell/helpers/append
tuffnatty: Agreed

src/vfs/shell/shell.c
tuffnatty: `>filename` is a traditional way to truncate files. But shellcheck started to dislike it lately, so probably `:` should be added for clarity.
Yes, the path is initially relative to vfs root, so `send` and `append` scripts add the leading slash.

src/vfs/shell/Makefile.am
tuffnatty: The first three commits visible here belong to #5155.

src/vfs/shell/helpers/append
tuffnatty: I've split out the new methods stuff and submitted it as #5158.

tuffnatty:
I've applied @ossilator 's suggestions and rebased the PR on master to ease reviewing.

src/vfs/shell/shell.c
```diff
@@ -1686,6 +1686,28 @@ shell_fill_names (struct vfs_class *me, fill_names_f func)
 static void *
 shell_open (const vfs_path_t *vpath, int flags, mode_t mode)
 {
+    if (flags & O_WRONLY)
```
mc-worker: Please use the explicit test: `if ((flags & O_WRONLY) != 0)`

src/vfs/shell/shell.c
```diff
@@ -1686,6 +1686,28 @@ shell_fill_names (struct vfs_class *me, fill_names_f func)
...
+        char *command;
+        struct vfs_s_super *super;
+        struct vfs_class *me;
+        int r;
+
+        me = VFS_CLASS (vfs_path_get_last_path_vfs (vpath));
+        name = vfs_s_get_path (vpath, &super, 0);
+        quoted_name = str_shell_escape (name);
+        command = g_strdup_printf ("if : %s /%s; then echo '### 200'; else echo '### 500'; fi\n",
+                                   (flags & O_APPEND) ? ">>" : ">", quoted_name);
```
mc-worker: Please use the explicit test: `(flags & O_APPEND) != 0`

src/vfs/shell/shell.c
tuffnatty: fixed

src/vfs/shell/shell.c
tuffnatty: fixed

PR #5160, opened 2026-09-19
vfs/sftpfs: Fix keyboard-interactive auth to FreeBSD OpenSSH server with UsePAM

### LINKED ISSUE #4585: ftp /sftp does not work
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/4585 |
> | **Reporter** | `fabian3357` (satay-zauberei0r@….com) |
> | **Keywords** | `ftp` |

<p>
i have tried several times to connect to midnight commader via ftp /sftp but it always fails with the message that the connection is not possible.<br />
</p>
<p>
i then tried lftp there the connection works directly but not with midnight commander what is the problem i entered the login link exactly according to the help under F1.<br />
</p>

tuffnatty:
## Proposed changes

I tried to use the SFTP vfs. I couldn't connect to any of my SFTP servers. First of all, one has to add an entry to ~/.ssh/config to be able to use publickey auth, and it's not documented anywhere (added #5161). After publickey auth fails, MC asks for a password. On a recent FreeBSD with default sshd settings, `PasswordAuthentication` is disabled, `KbdInteractiveAuthentication` is enabled, and `UsePAM` is enabled. So the password authentication fails, and MC starts performing keyboard-interactive auth. The first thing it does is potentially dereference a NULL pointer by passing it to `strncmp` for an unknown purpose. Then, it checks if one of the prompts passed by keyboard-interactive auth is "Password: " and answers to that. Unfortunately, in the described configuration the prompt string looks like "Password for user@hostname:", so auth fails.

This patch is a quick fix to make the default configuration work. However, the whole procedure is quite fragile, as the password prompt may be customized by user in PAM configuration.

I think the symptoms are very much alike the problem report #4585. Although it's missing details, I believe it would be hard to collect them without `gdb` at hand, as there is nothing special in this configuration and it should "just work".

* Could potentially resolve: #4585

## Checklist


ossilator:
yes, this looks incredibly fragile.

if mc is using a pty to talk to ssh, then password prompts could be detected by querying  the tty state - echo off means password.

the second though is: given that this is using libssh2, isn't there a more structured way to obtain the prompt and know the type of it?

the last though would be: why try so hard to interpret things, rather than just present them verbatim, like pam and sasl ask for (though these do tell the user what type of prompt it is)?

the commit message is again a bit on the short side. always try to answer the question "why?", in all variations that seem relevant.

tuffnatty:
Probably the right way to implement this would be not asking the user for a password in advance and presenting them the literal prompts from the remote side. It could enable various 2FA schemes as a side effect. Overall, it would require more research on the matter and potentially more disrupting changes, while this patch is a quick bug fix which could probably make it into the next release.

tuffnatty:
Closing in favour of #5162.

PR #5161, opened 2026-09-19
vfs/sftpfs: support default keys without tweaking .ssh/config

### LINKED ISSUE #3238: Add RSA key based authentication for SFTP
> [!IMPORTANT]
>
> | **Origin** | https://midnight-commander.org/ticket/3238 |
> | **Reporter** | `kslychan` (krzysztof.slychan@….pl) |
> | **Mentions** | `richlv@….net`, `dnh@….org`, `mooffie@….com` (@mooffie) |
> | **Keywords** | `ssh`, `sftp`, `rsa`, `key` |

<p>
mc asks me for user's password on a SSH server whenever I connect to it via SFTP, even though the client's public RSA key is authorized for that user account on the server and I can connect to it via openssh-client without giving password.<br />
</p>
<p>
Test case:<br />
</p>
<ol><li>on machine "foo", install mc, openssh-client
</li><li>on machine "foo", use ssh-keygen to generate a pair of RSA keys
</li><li>on machine "bar", install openssh-server
</li><li>on machine "bar", add the contents of machine A's ~/.ssh/id_rsa.pub to the ~/.ssh/authorized_keys file
</li><li>verify that you can connect from foo to bar without being asked for password: ssh bar (or: ssh differentusername@bar)
</li><li>if 5., then run mc, connect via SFTP, enter hostname
</li><li>you're asked for password
</li></ol><p>
Consider using key-based SSH authentication.<br />
</p>

tuffnatty:
## Proposed changes

1. The first patch adds support for multiple `IdentityFile` config lines, filtering out missing files and trying the rest in order.
2.  It also checks the exact result code from libssh before retrying with a passphrase, thus avoiding requesting password from user for non-acceptable keys.
3.  It changes the passphrase prompt from "Enter passphrase for USER" to "Enter passphrase for KEY_FILENAME".
4. As a side effect it reduces memory leaks in the code (the old `pubkey` and `privkey` were never freed).
5. The second patch fixes SSH config parsing to include toplevel file content, not just the Host sections.
6. The third patch adds an attempt to use `ssh -G` to get SSH to pre-parse its configuration for us, adding support for `Include` parsing and the default values for `IdentityFile`. Direct config file parsing is kept as a fallback.

I believe that this changeset improves UX around public key SFTP auth, making it 'just work' in the common case.

* Resolves: #3238

## Checklist


src/vfs/sftpfs/config_parser.c
```diff
@@ -331,6 +333,8 @@ sftpfs_fill_config_entity_from_config (FILE *ssh_config_handler,
 
 done:
     mc_search_free (host_regexp);
+    if (config_entity->identity_file != NULL)
```
mc-worker: This check is redundant, `g_slist_reverse()` handles `NULL` correctly.

src/vfs/sftpfs/connection.c
```diff
@@ -723,34 +723,46 @@ sftpfs_open_connection_ssh_key (struct vfs_s_super *super, GError **mcerror)
...
-    if (sftpfs_super->privkey == NULL)
+    if (sftpfs_super->privkeys == NULL)
         return FALSE;
 
-    if (libssh2_userauth_publickey_fromfile (sftpfs_super->session, super->path_element->user,
-                                             sftpfs_super->pubkey, sftpfs_super->privkey,
-                                             super->path_element->password)
-        == 0)
-        return TRUE;
+    for (node = sftpfs_super->privkeys; node != NULL && ret_value == FALSE; node = node->next)
```
mc-worker: Please do not explicit compare gboolean variables with `TRUE` and `FALSE`.
Please use the the `g_slist_next()` API instead of direct access to `GSList` member.

`for (node = sftpfs_super->privkeys; node != NULL && !ret_value; node = g_slist_next (node))`

src/vfs/sftpfs/connection.c
```diff
@@ -723,34 +723,46 @@ sftpfs_open_connection_ssh_key (struct vfs_s_super *super, GError **mcerror)
...
+    {
+        const char *privkey = node->data;
+        char *pubkey;
+        int res;
 
-    p = g_strdup_printf (_ ("sftp: Enter passphrase for %s "), super->path_element->user);
-    passwd = vfs_get_password (p);
-    g_free (p);
+        if (!g_file_test (privkey, G_FILE_TEST_EXISTS))
```
mc-worker: We have an `exist_file()` function.

tuffnatty:
Updated with the suggested changes.

ossilator:
re 2nd commit, is there actually a known case where the fallback would be necessary? it would be cleaner to just do away with it.

tuffnatty:
@ossilator The fallback works when `ssh` is not found or is pre-`-G` version. It does not cost much, as we still have to parse the output of `ssh -G`, so we need to keep most of the code.

ossilator:
well, obviously. the questions are:
- what's the point of it all when ssh is not found? if it needs only libssh, then it can work in principle, but is such a setup actually plausible?
- is "pre- -G ssh" actually a thing? when was this introduced?
- is the fallback actually clean, i.e. is the error message handled in an acceptable way? i think it just goes to stderr?

tuffnatty:
@ossilator of course there are many possible configurations where MC can run, including the ones where `ssh -G` is not available. I've added a short history of the option to the patch, as well as stderr redirection, thanks.

tuffnatty:
I've also inserted a commit to enable parsing toplevel configuration content, which is in accordance with how ssh does it and fixes compatibility with pre-9.2 implementation of `ssh -G`.

src/vfs/sftpfs/config_parser.c
```diff
@@ -364,17 +367,28 @@ sftpfs_get_config_entity (const vfs_path_element_t *vpath_element, GError **mcer
     config_entity->pubkey_auth = TRUE;
     config_entity->port = SFTP_DEFAULT_PORT;
 
+    // First, ask ssh to parse configs for us, filling in the defaults, includes etc.
+    // Only if it fails, parse user's config ourselves.
+    // ssh -G was intoduced in OpenSSH 6.7, but before version 9.2 it omitted
+    // the Host line at top.
```
ossilator: "at _the_ top"

but i kind of miss the point of the comment. it doesn't seem the relate to any consequence.

ossilator (commented):
ok, fair enough.

i still find the commit messages lacking. ;-)
for the ones that you do extend, please adjust according to https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

src/vfs/sftpfs/config_parser.c
tuffnatty: Added more detail.

tuffnatty:
> i still find the commit messages lacking. ;-) for the ones that you do extend, please adjust according to https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

@ossilator  In the commit messages, I tried to fit the essence of the change in a single line, and I find them basically fine, when you take the PR description into account. I also have not understood what do you mean by 'extending'. Could you probably come up with an improvement?

zyv:
I think this should be merged when ready notwithstanding, but I just want to note that this is yet another reason to switch to `libssh`. If you want to work on sftp, my messages in #3654 might be useful, whether we decide to patch up our `libssh2`-based client or move to `libssh`, if you haven't read them already.

ossilator:
> I tried to fit the essence of the change in a single line,

well, don't. read the link i gave you.

> when you take the PR description into account.

that's exactly the mistake. commit messages should be self-contained. github may be gone tomorrow, yet we want the history to persist. in git terms, the PR description is the cover letter, and it's irrelevant beyond the contribution process itself.

tuffnatty:
Made commit messages shorter. Added a commit to handle missing `.ssh/known_hosts` gracefully.

